PR #21, opened 2025-07-22
Update README.md

stefanfrench:
@alexmeckes - Was reading through the readme and noticed we use the term `evaluate` - I think we should drop this?


alexmeckes (commented):
lgtm

alexmeckes (approved):
lgtm

PR #23, opened 2025-07-23
feat: add api_base to aws

### LINKED ISSUE #22: Add Completion `api_base` support for provider `aws`
Thank you for putting this package together.

I was testing Amazon Bedrock with `any-llm` and realised the [FIPS endpoints](https://docs.aws.amazon.com/general/latest/gr/bedrock.html#br-rt) are unreachable. For example the Bedrock FIPS endpoing for us-east-2 is `bedrock-runtime-fips.us-east-2.amazonaws.com` the FIPS is before the region and so cannot be specific through the `AWS_REGION` environment variable.

Ideally, I could provide the Bedrock Runtime endpoint through `api_base`:

```python
response = completion(
    model='<model-id>',
    messages=[{"role": "user", "content": "Hello!"}],
    api_base="https:// bedrock-runtime-fips.us-east-2.amazonaws.com ", 
)
```

However, `api_base` is not used within the AWS provider implementation. I think this would be a useful feature, since it also enables users to use regular endpoints without having to specify the region through an environment variable.

I have it working locally making the change to `providers.aws._make_api_call()` by adding the following lines:
```python
        if self.config.api_base:
            client = boto3.client("bedrock-runtime", endpoint_url=self.config.api_base)
        else:
            client = boto3.client("bedrock-runtime", region_name=self.region_name)  # type: ignore[no-untyped-call]
```

This prioritizes `api_base` over the environment variable, but could also be switched if useful the other way around. I'm happy to open a PR if you're excepting external contributors.

PeskyPotato:
Closes #22

### DIFF
--- a/src/any_llm/providers/aws/aws.py
+++ b/src/any_llm/providers/aws/aws.py
@@ -51,10 +51,8 @@ class AwsProvider(Provider):
     ) -> ChatCompletion | Stream[ChatCompletionChunk]:
         """Create a chat completion using AWS Bedrock with instructor support."""
         self._check_aws_credentials()
-        if self.config.api_base:
-            client = boto3.client("bedrock-runtime", endpoint_url=self.config.api_base)
-        else:
-            client = boto3.client("bedrock-runtime", region_name=self.region_name)  # type: ignore[no-untyped-call]
+
+        client = boto3.client("bedrock-runtime", endpoint_url=self.config.api_base, region_name=self.region_name)
 
         if "response_format" in kwargs:
             instructor_client = instructor.from_bedrock(client)
--- a/src/any_llm/providers/aws/aws.py
+++ b/src/any_llm/providers/aws/aws.py
@@ -52,7 +52,7 @@ class AwsProvider(Provider):
         """Create a chat completion using AWS Bedrock with instructor support."""
         self._check_aws_credentials()
 
-        client = boto3.client("bedrock-runtime", endpoint_url=self.config.api_base, region_name=self.region_name)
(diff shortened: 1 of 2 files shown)


src/any_llm/providers/aws/aws.py
```diff
@@ -51,7 +51,10 @@ def _make_api_call(
     ) -> ChatCompletion | Stream[ChatCompletionChunk]:
         """Create a chat completion using AWS Bedrock with instructor support."""
         self._check_aws_credentials()
-        client = boto3.client("bedrock-runtime", region_name=self.region_name)  # type: ignore[no-untyped-call]
+        if self.config.api_base:
+            client = boto3.client("bedrock-runtime", endpoint_url=self.config.api_base)
+        else:
+            client = boto3.client("bedrock-runtime", region_name=self.region_name)  # type: ignore[no-untyped-call]
```
daavoo: I believe (haven't tested this) that we don't need an if/else because botocore will properly resolve the client (and won't break in the default case where `endpoint_url=None`) if you just pass both args:

```suggestion
        client = boto3.client(
            "bedrock-runtime", 
            region_name=self.region_name, 
            endpoint_url=self.config.api_base)
```

PeskyPotato: Good catch, I just tested this locally with a region as an environment where I do not have the model access enabled and passed in the api_base to a region where I do. It picked up the api_base. When I removed the api_base it then used the region instead and returned an unauthorised error as expected.

Will remove the if-else and just pass in both variables.

PeskyPotato (commented):
Change done

PeskyPotato:
Added two tests, with and without api_base. If all looks good, I can merge the commits to be cleaner.

Issue #25, opened 2025-07-23
Add support for llama api


njbrake:
https://www.llama.com/products/llama-api/

PR #29, opened 2025-07-24
Llama support

AshAnand34:
Added Llama API support to any-llm in response to #25.

Please let me know if there are any changes that need to be made to the code.

### DIFF
--- a/.github/workflows/tests-integration.yaml
+++ b/.github/workflows/tests-integration.yaml
@@ -42,5 +42,7 @@ jobs:
 
       - name: Run Integration tests (parallel with xdist)
         env:
+          GEMINI_API_KEY: ${{ secrets.GEMINI_API_KEY }}
           MISTRAL_API_KEY: ${{ secrets.MISTRAL_API_KEY }}
+          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
         run: pytest -v tests/integration -n auto
--- a/src/any_llm/providers/azure/azure.py
+++ b/src/any_llm/providers/azure/azure.py
@@ -6,7 +6,7 @@ from typing import Any
 
 from openai.types.chat.chat_completion import ChatCompletion
 from any_llm.provider import Provider, ApiConfig
-from any_llm.exceptions import MissingApiKeyError, UnsupportedParameterError
+from any_llm.exceptions import UnsupportedParameterError
 from openai._streaming import Stream
 from openai.types.chat.chat_completion_chunk import ChatCompletionChunk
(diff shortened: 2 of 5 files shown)


tests/unit/providers/test_llama_provider.py
```diff
@@ -0,0 +1,27 @@
...
+from any_llm.exceptions import UnsupportedParameterError
+
+def test_verify_kwargs():
+    provider = LlamaProvider(config={"api_key": "test_key"})
+
+    # Test unsupported stream parameter
+    with pytest.raises(UnsupportedParameterError):
+        provider.verify_kwargs({"stream": True})
+
+def test_make_api_call():
```
daavoo, line 12: This mock doesn't make sense, you are mocking the high-level method that you are directly calling later.

If you were to mock something, it should be the underlying client (i.e. `LlamaClient` in this case), similar to how it is done in existing tests:

https://github.com/mozilla-ai/any-llm/blob/main/tests/unit/providers/test_anthropic_provider.py

pyproject.toml
```diff
@@ -66,6 +66,11 @@ azure = [
   "httpx",
 ]
 
+llama = [
+  "llama-api-client",
+  "instructor",
```
daavoo, line 71: This dependency is not used

src/any_llm/providers/llama/llama.py
```diff
@@ -0,0 +1,65 @@
...
+    """Llama Provider using the new response conversion utilities."""
+
+    PROVIDER_NAME = "Llama"
+    ENV_API_KEY_NAME = "LLAMA_API_KEY"
+    PROVIDER_DOCUMENTATION_URL = "https://www.llama.com/docs/integration-guides/"
+
+    def verify_kwargs(self, kwargs: dict[str, Any]) -> None:
+        """Verify the kwargs for the Llama provider."""
+        if kwargs.get("stream", False):
+            raise UnsupportedParameterError("stream", self.PROVIDER_NAME)
```
daavoo, line 31: This parameter is actually supported:

https://github.com/meta-llama/llama-api-python/blob/a26e06f8a30c89c46c27feaead08e240c9e11d52/README.md?plain=1#L85-L102

daavoo (changes requested):
@AshAnand34 thanks for the contribution, but I see a lot of indications of this P.R. being mostly (or fully) auto-generated. Please correct me if I am wrong.

@njbrake unless I am missing something, this API requires joining a waitlist (https://llama.developer.meta.com/join-waitlist/), are we interested in adding support for an API that is not publicly available yet (do you have access?)

PR #56, opened 2025-07-31
feat(docs): Add `Streaming` section and `Stream Supported` column.

daavoo:


scripts/hooks.py
```diff
@@ -20,6 +20,21 @@
...
+    "cohere",
+    "deepseek",
+    "google",
+    "inception",
+    "mistral",
+    "moonshot",
+    "nebius",
+    "openai",
+    "xai",
+]
```
njbrake, line 36: This should be done programatically, should not be hardcoded.

PR #58, opened 2025-08-04
tests(xai): Use `grok-3-mini-latest` instead of `grok-3-latest`.

daavoo:
It's about 10x cheaper and works the same for our testing purposes:

https://docs.x.ai/docs/models#model-pricing


daavoo:
Integration tests: https://github.com/mozilla-ai/any-llm/actions/runs/16724858251

PR #84, opened 2025-08-06
feat: add streaming support in ollama, fix anthropic bug for system m…

njbrake:
…essages

### DIFF
--- a/src/any_llm/providers/ollama/utils.py
+++ b/src/any_llm/providers/ollama/utils.py
@@ -161,7 +161,7 @@ def _create_response_dict_from_ollama_response(
         for tool_call in response_message.tool_calls:
             tool_calls.append(
                 {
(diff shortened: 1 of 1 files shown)


tests/integration/test_streaming.py
```diff
@@ -21,7 +21,10 @@ def test_streaming_completion(
         for result in completion(
             f"{provider.value}/{model_id}",
             **extra_kwargs,
-            messages=[{"role": "user", "content": "Say 'Hello World'"}],
+            messages=[
+                {"role": "system", "content": "You are a helpful assistant that exactly follows the user request."},
```
daavoo, line 25: @njbrake I guess this serves as a partial regression test for the anthropic fix, but it is missing the coverage for the "non-streaming" call.

Maybe some small unit tests could be added to cover both?

PR #97, opened 2025-08-07
test: Use get_provider_metadata to skip in completion and streaming.

daavoo:


njbrake:
I'm all for this change, but I think we actually need a better way to extract meta-data on a per provider basis. Maybe a helper method on the factory to make this easier?

daavoo:
> I'm all for this change, but I think we actually need a better way to extract meta-data on a per provider basis. Maybe a helper method on the factory to make this easier?

yeah. 
I was tempted to just refactor the get_provider_metadata as it seems overly complicated given that we could just check the constants `SUPPORTS_X`, but wanted to go with the quickest approach. Can do in a followup

PR #98, opened 2025-08-07
feat: Add support for VoyageAI. 

daavoo:
This provider only supports `embedding` so I added a `SUPPORTS_COMPLETION` feature flag.

Needs #97


daavoo:
@njbrake the Docs fail because we try to check the source code url against `main` branch. Not sure how much value there is in this check, but it will basically fail on any branch adding a new provider

njbrake:
> @njbrake the Docs fail because we try to check the source code url against `main` branch. Not sure how much value there is in this check, but it will basically fail on any branch adding a new provider

Yeah it's a known bug, I sort of just ignored because it didn't seem like a big problem, we won't add new providers very often I assume 🤷

PR #106, opened 2025-08-08
fix(anthropic): Drop duplicated `stream` + `response_format` check.

daavoo:
Check is already happening inside `verify_kwargs`.


daavoo:
Integration tests: https://github.com/mozilla-ai/any-llm/actions/runs/16829005416

Issue #111, opened 2025-08-08
Responses API for Fireworks AI


njbrake:
https://fireworks.ai/docs/guides/response-api

Shiva4113:
@njbrake I'd like to work on this if it's open. I've gone through the PR #108 and the [fireworks responses API documentation](https://fireworks.ai/docs/guides/response-api). Could you please let me know if there's any other resources I should go through before I can work on this?

njbrake:
Hi @Shiva4113 help would be great! To implement this, you'll need to make the call to the fireworks responses Api similar to how we make the call in https://github.com/mozilla-ai/any-llm/blob/main/src/any_llm/providers/openai/base.py#L149, but then you'll probably need to work through the conversion logic to turn the Fireworks Response pydantic object to conform to the OpenAI responses object, if they don't exactly match.

Since I haven't yet finished the conversion logic for any of the responses API providers yet, I don't have any examples that you can closely follow just yet. 

Any feedback about how it goes would be useful, so that we can update our Contributing.md file with any relevant tips.

PR #121, opened 2025-08-11
feat(fireworks): Responses API for Fireworks AI

Shiva4113:
## Related Issue

This Pull Request is related to the issue #111 .

## Summary

Added support for using Fireworks as a provider with the Responses API.

## Notes from Development

- During the build process, there are issues with the latest version of `fireworks-ai` (v0.19.18).
- When building with `any-llm-sdk` after running `pip install fireworks-ai`, a dependency issue arises:
  - `reward-kit` (v4.1) is incompatible with `openai` (v1.99.6).
- As an alternative, attempting to build with `pip install any-llm-sdk[fireworks]` results in a version of `fireworks-ai` that is **too old** to support the new Responses API feature.

## Additional Changes

- Added a **Response Wrapper** utility to adapt the Fireworks response object to match the **OpenAI Response** object format.


src/any_llm/providers/fireworks/fireworks.py
```diff
@@ -74,3 +76,21 @@ def _make_api_call(
...
+
+    def responses(self, model, input_data, **kwargs) -> Response | Iterator[ResponseStreamEvent]:
+        """Call Fireworks Responses API and normalize into ChatCompletion/Chunks."""
+        client = OpenAI(
+            base_url="https://api.fireworks.ai/inference/v1",
+            api_key=self.config.api_key,
+        )
+
+        response = client.responses.create(
+            model=f'accounts/fireworks/models/{model}',
```
njbrake: Our current approach is that the model must contain the entire string, like https://github.com/mozilla-ai/any-llm/blob/main/tests/conftest.py#L38.

It's not clear to me whether we can safely hardcode the `accounts/fireworks/models` bit.

src/any_llm/providers/fireworks/utils.py
```diff
@@ -54,3 +54,27 @@ def _create_openai_chunk_from_fireworks_chunk(fireworks_chunk: Any) -> ChatCompl
...
+                raw_output = self._response.output[-1].content[0].text
+                self._output_text = raw_output.split("</think>")[-1].strip()
+            except (IndexError, AttributeError):
+                self._output_text = ""
+        return self._output_text or ""
+
+
+def _create_response_with_output_text(fireworks_response: Any) -> Any:
+    """Wrap a Fireworks response to add output_text property."""
+    return ResponseWrapper(fireworks_response)
```
njbrake: From looking at https://github.com/openai/openai-python I think output_text is a helper property. Would it work to instead do something like OpenAIResponse.model_validate(fireworks_response.model_dump())

src/any_llm/providers/fireworks/utils.py
```diff
@@ -54,3 +54,27 @@ def _create_openai_chunk_from_fireworks_chunk(fireworks_chunk: Any) -> ChatCompl
...
+        if self._output_text is None and hasattr(self._response, 'output') and self._response.output:
+            try:
+                raw_output = self._response.output[-1].content[0].text
+                self._output_text = raw_output.split("</think>")[-1].strip()
+            except (IndexError, AttributeError):
+                self._output_text = ""
+        return self._output_text or ""
+
+
+def _create_response_with_output_text(fireworks_response: Any) -> Any:
```
njbrake: Needs proper typing

src/any_llm/providers/fireworks/fireworks.py
Shiva4113: Yeah sorry, it's totally not safe to do so

Just using `model = model` works completely fine

src/any_llm/providers/fireworks/utils.py
Shiva4113: Thank you for this suggestion, this completely eliminated the need of the utility `_create_response_with_output_text`

src/any_llm/providers/fireworks/utils.py
Shiva4113: The need for this utility was eliminated thanks to a suggestion above

Shiva4113:
Addressed the concerns previously raised, kindly let me know if any more changes are required.

src/any_llm/providers/fireworks/fireworks.py
```diff
@@ -74,3 +76,23 @@ def _make_api_call(
...
+            input=input_data,
+            **kwargs,
+        )
+
+        if not isinstance(response, (Response, Stream)):
+            raise ValueError(f"Responses API returned an unexpected type: {type(response)}")
+
+        response = Response.model_validate(response.model_dump())
+
+        return response
```
njbrake: Since currently the Any LLM response type is an alias to the OpenAI type, you may actually be able to make mypy happy by removing the model_validate line.
```suggestion
        return response
```

njbrake (changes requested):
Thanks! A few more changes needed. 

1. You'll need to merge in the latest changes from main
2. Run linting to surface the type errors. If you use uv: `uv run pre-commit run --all-files --verbose` this will run mypy and ruff on your code and surface a few of the missing typing issues
3. You need to handle both streaming and non streaming in your return. Your code works fine now for non-streaming, but if it's streaming, you'll need to re-yield each chunk as the any-llm type. I know it seems like a silly thing to do right now, but the idea is to decouple the fireworks output from the any-llm typing to make it flexible if fireworks decides to change their output information. You _may_ actually be able to directly return the `response`, it's possible that mypy will understand that the Any-LLM and OpenAI responses type are currently the same thing.

Shiva4113:
Merged the latest commits from main and ran linting to surface all type errors. 
Fixed the errors as well and also committed the requested change regarding model_validate().

Still working on the streaming responses.

src/any_llm/providers/fireworks/fireworks.py
```diff
@@ -95,3 +97,19 @@ def completion(
...
+        )
+        response = client.responses.create(
+            model=model,
+            input=input_data,
+            **kwargs,
+        )
+        if not isinstance(response, Response | Stream):
+            msg = f"Responses API returned an unexpected type: {type(response)}"
+            raise ValueError(msg)
+        return response
```
njbrake: Ok, I'm not entirely sure if mypy would be happy with this or not, but I think technically it might be fine to directly return the output. Not sure if mypy will be ok with that 😅 

```suggestion
        return client.responses.create(
            model=model,
            input=input_data,
            **kwargs,
        )
```

Shiva4113: Just a small issue, sorry for overlooking this earlier.
The returned here includes the part where the LLM thinks and we'll have to process this output in this way `response.output[-1].content[0].text.split("</think>")[-1]` to get only the relevant response.

Example response: 
`<think>`
LLM thinking content here
`</think>`

Actual response


Just wanted to confirm with you once whether we can allow this kind of a response(that includes the thought and the actual reply) or return only the reply. 

I will make the changes according to your reply

njbrake:
@Shiva4113  Thanks for the work on this! I'll make a few pushes to get it over the finish line and then will merge it in 🙏

PR #138, opened 2025-08-13
feat: Add typing for responses API

besaleli:
## Description

- Added typing for response API

## Testing
- Unit tests/integration tests pass

## Notes
- Python 3.11+ supports `TypeAlias` type hint, which we should probably use on any type aliases (unless we're targeting <=3.10 compatibility) in a separate PR. I'll raise an issue. (UPDATE: looks like we're >=3.11, I'll raise the issue.)
- I'm only adding aliases for the first level of OpenAI input models. Let me know if we want to go deeper than that

### DIFF
--- a/src/any_llm/api.py
+++ b/src/any_llm/api.py
@@ -6,7 +6,7 @@ from pydantic import BaseModel
 from any_llm.provider import ApiConfig, Provider, ProviderFactory
 from any_llm.tools import prepare_tools
 from any_llm.types.completion import ChatCompletion, ChatCompletionChunk, CreateEmbeddingResponse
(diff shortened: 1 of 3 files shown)


njbrake:
> I'm only adding aliases for the first level of OpenAI input models. Let me know if we want to go deeper than that

@besaleli I think this is perfect for the time being. If/when we start to encounter providers that aren't perfectly compliant with the responses API, we can assess whether we need to switch it from a typealias to an extension.

> TypeAlias

I'm on board with using typealias 👍

daavoo (approved):
🚀

PR #141, opened 2025-08-13
chore: Add `CONTRIBUTING.md`.

daavoo:


CONTRIBUTING.md
```diff
@@ -0,0 +1,69 @@
...
+
+Thank you for your interest in contributing to this repository! This project supports the mozilla.ai goal of empowering developers to integrate AI capabilities into their projects using open-source tools and models.
+
+We welcome all kinds of contributions, from improving customization, to extending capabilities, to fixing bugs. Whether you’re an experienced developer or just starting out, your support is highly appreciated.
+
+---
+
+## **How to Contribute**
+
+### **Customize for your use-case or Extend It** 🔧
```
daavoo: This is copy-pasta from I am not sure where. IMHO, we could drop it and keep this doc focused on instructions for install,lint,test,docs but I see most repos having this generic test that feels a little bit like LLM generated

njbrake: Yes lets drop this. This sounds more relevant for a blueprint.

CONTRIBUTING.md
```diff
@@ -0,0 +1,69 @@
+# Contributing to mozilla.ai any-llm
+
+Thank you for your interest in contributing to this repository! This project supports the mozilla.ai goal of empowering developers to integrate AI capabilities into their projects using open-source tools and models.
```
njbrake:
```suggestion
Thank you for your interest in contributing to this repository! This project supports the mozilla.ai goal of empowering choice and transparency when it comes to LLM usage and selection.
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,69 @@
...
+- Check the Issues page to see if there are any tasks you'd like to tackle.
+- Look for issues labeled **`good first issue`** if you're new to the project—they're a great place to start.
+
+### **Report Issues** 🐛
+- Found a bug? Open a Bug Report by clicking on 'New Issue'
+- Provide as much detail as possible, including the steps to reproduce the issue and Expected vs. actual behavior
+
+### **Suggest Features** 🚀
+- Have an idea for improving the project? Open a Feature Request by clicking on 'New Issue'
+- Share why the feature is important and any alternative solutions you’ve considered.
```
njbrake: To me it doesn't seem like a value add? Describing how github works
```suggestion
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,69 @@
...
+### **Guidelines for Contributions**
+
+**Install**
+
+We recommend to use [uv](https://docs.astral.sh/uv/getting-started/installation/):
+
+```
+uv venv
+source .venv/bin/activate
+uv sync --dev --extra all
```
njbrake: This is what I use 🤷 
```suggestion
uv sync --all-extras -U --python=3.13
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,41 @@
...
+pre-commit run --all-files
+```
+
+**Tests**
+
+Test changes locally to ensure functionality.
+
+```bash
+pytest -v tests
+```
```
njbrake:
```suggestion
```bash
pytest -v tests/unit
pytest -v tests/integration -n auto
```
```

File: CONTRIBUTING.md (hunk 4 above)
Comment (daavoo) [reply CONTRIBUTING.md]: Why not `--dev`? (installs lint, test and docs deps, which is relevant for local development)

And why `--python=3.13` in the `sync` instead of when creating the `uv venv`?

File: CONTRIBUTING.md [hunk 6]
```diff
@@ -0,0 +1,41 @@
...
+
+**Lint**
+
+Ensure all the checks pass:
+
+```bash
+pre-commit run --all-files
+```
+
```
besaleli, line 26: Add section on running formatting and linting for more verbose checks? I've been using:

```
uv run -m ruff check --fix
uv run -m ruff format
```

daavoo: > for more verbose checks?

Is there something specific missing from pre-commit output (with verbose flags like `-vv`)

---
 
What I have recommended in the past is to also run specific checks via `pre-commit` (just to reduce the potential mismatches between local and CI), like:

```
pre-commit run ruff
pre-commit run ruff-format
```

TBH I am not sure what is going on when, sometimes, you get different results from `uv run -m ruff` and `pre-commit run ruff`. 
I feel like there are several things that can go wrong like having different the `ruff` versions, failing to load `pyproject.toml` config, some annoying quirk in `pre-commit` cache, etc.

daavoo:
> Looks good otherwise! Do we have company-wide standards on contributing and public-facing guidelines? 

Afaik, we do not. 

> That might be a DevRel thing to flag for later

Totally agree

CONTRIBUTING.md
besaleli: @daavoo It might just be that I don't know how to read. Good catch

CONTRIBUTING.md
njbrake: When testing different python versions and the venv already existed, `--python` swapped the version for me on the fly, IIRC.

the `--all-extras` tag installs everything, both the `all` extra as well as everything else (which includes the dev deps). Not sure if that was your question

CONTRIBUTING.md
njbrake: +1 on this. There is still something messed up with how we run ruff, but I haven't prioritized figuring out exactly what the issue is. Any ideas about what the problem is are welcome 🏅

CONTRIBUTING.md
```diff
@@ -0,0 +1,41 @@
...
+source .venv/bin/activate
+uv sync --dev --extra all
+```
+
+**Lint**
+
+Ensure all the checks pass:
+
+```bash
+pre-commit run --all-files
```
njbrake: This is what I use 🤷 
```suggestion
uv run pre-commit run --all-files --verbose
```

CONTRIBUTING.md
daavoo: > When testing different python versions and the venv already existed, --python swapped the version for me on the fly, IIRC.

So you usually change the `--python=` argument? (I just use the default all the time) 

> the --all-extras tag installs everything, both the all extra as well as everything else (which includes the dev deps). Not sure if that was your question

Aah. I didn't know that 👍

CONTRIBUTING.md
daavoo: what does `uv run` do compared to `pre-commit run` in the active `venv`? 

FTR, I am fine with just running what we all agree to run and we document in CONTRIBUTING, just trying to understand what each thing does

njbrake: totally. I think this makes sure that we use the local install of pre-commit even if there is one installed globally? This way someone can run the pre-commit even if they didn't source the env yet.

CONTRIBUTING.md
daavoo: Created https://github.com/mozilla-ai/any-llm/issues/144

PR #143, opened 2025-08-13
feat: Add `TypeAlias` for imported OpenAI Types

besaleli:
Python 3.11+ supports `TypeAlias`, good to keep explicit

### DIFF
--- a/src/any_llm/types/completion.py
+++ b/src/any_llm/types/completion.py
@@ -54,7 +54,9 @@ class ChatCompletionChunk(OpenAIChatCompletionChunk):
 
 
 ChatCompletionMessageFunctionToolCall: TypeAlias = OpenAIChatCompletionMessageFunctionToolCall
-ChatCompletionMessageToolCall: TypeAlias = OpenAIChatCompletionMessageFunctionToolCall | OpenAIChatCompletionMessageToolCall
+ChatCompletionMessageToolCall: TypeAlias = (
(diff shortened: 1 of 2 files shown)


daavoo (approved):
No strong opinion, apparently it is deprecated in 3.12 😅 

> Deprecated since version 3.12: [TypeAlias](https://docs.python.org/3/library/typing.html#typing.TypeAlias) is deprecated in favor of the [type](https://docs.python.org/3/reference/simple_stmts.html#type) statement, which creates instances of [TypeAliasType](https://docs.python.org/3/library/typing.html#typing.TypeAliasType) and which natively supports forward references. Note that while [TypeAlias](https://docs.python.org/3/library/typing.html#typing.TypeAlias) and [TypeAliasType](https://docs.python.org/3/library/typing.html#typing.TypeAliasType) serve similar purposes and have similar names, they are distinct and the latter is not the type of the former. Removal of [TypeAlias](https://docs.python.org/3/library/typing.html#typing.TypeAlias) is not currently planned, but users are encouraged to migrate to [type](https://docs.python.org/3/reference/simple_stmts.html#type) statements.

From: https://docs.python.org/3/library/typing.html#typing.TypeAlias

besaleli:
@daavoo oh wow I lowkey love that. I don't think this is necessary then for Python3.11 compatibility

njbrake:
For whatever reason, openai sdk uses typealias quite a bit in their docs. https://github.com/openai/openai-python/commit/064910b115e21837dd793390e6cfbeddd07e5f9a

PR #145, opened 2025-08-13
fix(anthropic): Handle messages inside agent loop.

daavoo:
Integration tests: https://github.com/mozilla-ai/any-llm/actions/runs/16938318602

### DIFF
--- a/tests/unit/providers/test_anthropic_provider.py
+++ b/tests/unit/providers/test_anthropic_provider.py
@@ -182,7 +182,7 @@ def test_make_api_call_inside_agent_loop() -> None:
 
     with mock_anthropic_provider() as mock_anthropic:
         provider = AnthropicProvider(ApiConfig(api_key=api_key))
-        provider.completion(model, messages)
+        provider.completion(model, messages)  # type: ignore[arg-type]
 
         mock_anthropic.return_value.messages.create.assert_called_once_with(
             model=model,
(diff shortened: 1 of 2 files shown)


tests/unit/providers/test_anthropic_provider.py
```diff
@@ -163,3 +163,38 @@ def test_stream_with_response_format_raises() -> None:
...
+        {"role": "user", "content": "What is the weather like in Salvaterra?"},
+        {
+            "role": "assistant",
+            "content": "",
+            "tool_calls": [
+                {"id": "foo", "function": {"name": "get_weather", "arguments": '{"location": "Salvaterra"}'}}
+            ],
+        },
+        {"role": "tool", "tool_call_id": "foo", "content": "sunny"},
+    ]
```
daavoo, line 181: @njbrake this is kind of the "flow" that gets used in `any-agent` (or any other "agent loop"). 

We are and will be running integration tests there for different providers (that is how I found this anthropic thing) but do you think it is worth also testing this setup where messages include a `tool_call` and `tool_response`?

njbrake: Making sure I understand: you're saying the scenario where the provider used a tool remotely, vs a provider where the LLM is indicating that a tool should be used, but the tool is your responsibility to execute?

daavoo: > Making sure I understand: you're saying the scenario where the provider used a tool remotely, vs a provider where the LLM is indicating that a tool should be used, but the tool is your responsibility to execute?

I was talking about including tests where there is a LLM tool_call and a tool result (coming from wherever) as part of the input messages. 

Those 2 type of messages as input are not being currently tested (in our test_tool, we pass regular `user` message as input and assert on the given tool_call output).

njbrake: Ah ok I get it now 🙏

PR #154, opened 2025-08-14
chore: pydantic typed between api.completion <-> provider.completion

njbrake:
This PR would keep the typing of the args all the way in to each provider implementation, instead of turning into a generic dict right at the entry API.

Using this as a basis, we can now start to tie parameters directly into each provider SDK, and also evaluate which input parameters are irrelevant ( I think there are a few parameters that are not worth typing)

### DIFF
--- a/src/any_llm/api.py
+++ b/src/any_llm/api.py
@@ -5,8 +5,9 @@ from pydantic import BaseModel
 
 from any_llm.provider import ApiConfig, ProviderFactory
 from any_llm.tools import prepare_tools
-from any_llm.types.completion import ChatCompletion, ChatCompletionChunk, CompletionParams, CreateEmbeddingResponse
+from any_llm.types.completion import ChatCompletion, ChatCompletionChunk, CreateEmbeddingResponse
 from any_llm.types.responses import Response, ResponseInputParam, ResponseStreamEvent
+from any_llm.utils.api import _process_completion_params
 
 
 def completion(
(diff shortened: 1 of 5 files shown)


src/any_llm/types/completion.py
```diff
@@ -60,3 +62,65 @@ class ChatCompletionChunk(OpenAIChatCompletionChunk):
 Usage = OpenAIUsage
 ChoiceDeltaToolCall = OpenAIChoiceDeltaToolCall
 ChoiceDeltaToolCallFunction = OpenAIChoiceDeltaToolCallFunction
+
+
+class CompletionParams(BaseModel):
```
daavoo, line 67: @njbrake could we do the stuff like `stream and response_format not allowed` here using https://docs.pydantic.dev/latest/concepts/validators/#model-validators ?

src/any_llm/api.py
```diff
@@ -135,30 +58,41 @@ def completion(
...
+        messages=messages,
+        tools=prepared_tools,
         tool_choice=tool_choice,
-        max_turns=max_turns,
         temperature=temperature,
         top_p=top_p,
         max_tokens=max_tokens,
         response_format=response_format,
-        stream=stream,
+        stream=stream or False,
```
daavoo: Why `or False` here instead of handling default value in the pydantic object?

src/any_llm/api.py
```diff
@@ -212,30 +144,41 @@ async def acompletion(
...
+        config["api_base"] = str(api_base)
+    api_config = ApiConfig(**config)
+
+    provider = ProviderFactory.create_provider(provider_key, api_config)
+
+    prepared_tools: list[dict[str, Any]] | None = None
+    if tools:
+        prepared_tools = prepare_tools(tools)
+
+    completion_params = CompletionParams(
```
daavoo: Am I missing something or this whole codeblock is duplicated between completion and acompletion? (Doesn't need to be tackled in this PR)

njbrake: Totally. No problem, I'll consolidate, easy enough

src/any_llm/api.py
njbrake: I kept going back and forth about whether or not to allow stream=None. I'll make this consistent before I merge 👍

src/any_llm/types/completion.py
njbrake: I originally didn't do that because apparently there is a way that structured outputs and streaming actually do work together, like https://platform.openai.com/docs/guides/structured-outputs#streaming.

PR #157, opened 2025-08-14
feat: response_format for cohere provider

besaleli:
# Description
- Adds support for `response_format` for Cohere provider
- Enables `response_format` integration tests for Cohere

## Notes
- The way that Cohere handles structured outputs is significantly different than Fireworks so I wrote logic in the Cohere provider instead of making a universal helper function. I think that something like an abstract method or smaller formatting function for Pydantic models specifically could be helpful if/when we want to do things like validate desired JSON schemas on a per-provider basis.
- On that note, there's a ton of limitations in the way Cohere does structured outputs and I'm letting the Cohere client handle it.

### DIFF
--- a/src/any_llm/providers/cohere/cohere.py
+++ b/src/any_llm/providers/cohere/cohere.py
@@ -54,6 +54,23 @@ class CohereProvider(Provider):
 
         for chunk in cohere_stream:
             yield _create_openai_chunk_from_cohere_chunk(chunk)
+    
+    @staticmethod
+    def _preprocess_response_format(response_format: type[BaseModel] | dict) -> dict:
+        # if response format is a BaseModel, generate model json schema
+        if isinstance(response_format, type) and issubclass(response_format, BaseModel):
+            return {"type": "json_object", "schema": response_format.model_json_schema()}
+        # can either be json schema already in dict
+        # or {"type": "json_object"} to just generate *a* JSON (JSON mode)
+        # see docs here: https://docs.cohere.com/docs/structured-outputs#json-mode
+        elif isinstance(response_format, dict):
+            return response_format
+        # For now, let Cohere API handle invalid schemas.
+        # Note that Cohere has a bunch of limitations on JSON schemas (e.g., no oneOf, numeric/str ranges, weird regex limitations)
+        # see docs here: https://docs.cohere.com/docs/structured-outputs#unsupported-schema-features
+        else:
+            # Validation logic could/would eventually go here
+            return response_format
 
     def completion(
         self,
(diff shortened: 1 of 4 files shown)


besaleli:
Unit tests are failing because of invalid API key 😅:

```
FAILED tests/unit/providers/test_cohere_provider.py::test_response_format_raises_for_non_streaming - cohere.errors.unauthorized_error.UnauthorizedError: headers: {'cache-control': 'no-cache, no-store, no-transform, must-revalidate, private, max-age=0', 'content-type': 'application/json', 'expires': 'Thu, 01 Jan 1970 00:00:00 GMT', 'pragma': 'no-cache', 'vary': 'Origin', 'x-accel-expires': '0', 'date': 'Thu, 14 Aug 2025 07:58:17 GMT', 'content-length': '75', 'x-envoy-upstream-service-time': '8', 'server': 'envoy', 'via': '1.1 google', 'alt-svc': 'h3=":443"; ma=2592000,h3-29=":443"; ma=2592000'}, status_code: 401, body: {'id': 'b069dc5a-f02f-4538-bf27-2ce6908b5b4c', 'message': 'invalid api token'}
```

src/any_llm/providers/cohere/cohere.py
```diff
@@ -61,18 +63,34 @@ def completion(
...
             raise UnsupportedParameterError(msg, self.PROVIDER_NAME)
 
-        if kwargs.get("stream", False):
-            kwargs.pop("stream")
+        stream = kwargs.pop("stream", False)
+
+        if stream:
             return self._stream_completion(model, messages, **kwargs)
+
+        # note: ClientV2.chat does not have a `stream` parameter
```
daavoo, line 96: Double checking: this is a separate, unrelated?, fix that was not being catch because we didn't have `CO_API_KEY` for integration tests, right?

pd. I am fine with having it in this P.R.

tests/integration/test_response_format.py
```diff
@@ -21,9 +21,9 @@ def test_response_format(
     if not cls.SUPPORTS_COMPLETION:
         pytest.skip(f"{provider.value} does not support response_format, skipping")
     """Test that all supported providers can be loaded successfully."""
-    if provider in [ProviderName.COHERE]:
-        pytest.skip(f"{provider.value} does not support response_format")
-        return
+    # if provider in [ProviderName.COHERE]:
+    #     pytest.skip(f"{provider.value} does not support response_format")
+    #     return
```
daavoo:
```suggestion
```

tests/unit/providers/test_cohere_provider.py
```diff
@@ -13,15 +14,27 @@ def _mk_provider() -> Any:
     return CohereProvider(ApiConfig(api_key="test-api-key"))
 
 
-def test_response_format_raises_for_non_streaming() -> None:
+def test_preprocess_response_format() -> None:
```
daavoo, line 17: 👍

daavoo (approved):
CI Integration tests: https://github.com/mozilla-ai/any-llm/actions/runs/16960111414

src/any_llm/providers/cohere/cohere.py
besaleli: @daavoo `stream` was only being popped from the kwargs in the case that it was True, which was a problem since the ClientV2.chat does not have a `stream` parameter (was throwing an error otherwise). Here, we pop it from the start—other than that, nothing more than a tiny refactor

PR #159, opened 2025-08-14
fix(google): Handle multiple input `messages` appropriately.

daavoo:
We were concatenating the user messages into a single string (not sure why) and leaving out other messages (i.e. assistant responses and tool call results).

This was an incorrect implementation for both agent loops and multi-turn conversations (because assistant responses were not being fed back to the model).

See https://ai.google.dev/gemini-api/docs/text-generation#rest_6 for reference.

### DIFF
--- a/src/any_llm/providers/google/google.py
+++ b/src/any_llm/providers/google/google.py
@@ -123,29 +123,17 @@ class GoogleProvider(Provider):
 
         formatted_messages = _convert_messages(messages)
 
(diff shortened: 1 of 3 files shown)


daavoo:
Integration tests: https://github.com/mozilla-ai/any-llm/actions/runs/16960756917

njbrake (approved):
Thank you for the find and fix!

PR #160, opened 2025-08-14
fix(google): Extract and pass `system_instruction`.

daavoo:
Instead of converting it to `user` message.

See https://ai.google.dev/gemini-api/docs/text-generation#system-instructions.


daavoo:
Integration tests: https://github.com/mozilla-ai/any-llm/actions/runs/16962012033

PR #163, opened 2025-08-14
fix(xai): Preserve tool_call id from client response.

daavoo:
We were discarding it and replacing with a random uuid.


daavoo:
Integration tests: https://github.com/mozilla-ai/any-llm/actions/runs/16965817849/job/48089342552

PR #165, opened 2025-08-14
fix(xai): Handle messages inside agent loop.

daavoo:


daavoo:
Integration tests: https://github.com/mozilla-ai/any-llm/actions/runs/16966829442/job/48092896088

PR #176, opened 2025-08-14
feat(provider): add Perplexity (OpenAI-compatible) — chat + streaming

### LINKED ISSUE #127: Add Perplexity provider (perplexity/*) — completion + streaming (MVP)
I’d like to add a new `perplexity` provider using Perplexity’s OpenAI-compatible Chat Completions API. V1 will support the standard OpenAI fields only (`model`, `messages`, `max_tokens`, `temperature`, `top_p`, `stream`) with both non-stream and SSE streaming. 

Would love to work on this - if this can be assigned to me!

Scope (MVP)
- Endpoint: POST https://api.perplexity.ai/chat/completions
- Auth: `PERPLEXITY_API_KEY` (Bearer); optional `PERPLEXITY_BASE_URL` override
- Features: non-stream + SSE streaming; proper `finish_reason`
- Return: OpenAI-style response object

I’ve checked existing issues and didn’t find Perplexity—happy to proceed once assigned.

mihirahuja1:
Adds a new Perplexity provider using their OpenAI-compatible Chat Completions API.

Auth / Config:
- PERPLEXITY_API_KEY (required)
- Optional base URL override via PERPLEXITY_BASE_URL env or ApiConfig(api_base=...)
Default base: https://api.perplexity.ai

Supported:
- Chat Completions (non-stream + SSE streaming)
- Proper finish_reason

Not in scope:
- Responses API
- Embeddings

Usage:
```python
from any_llm.provider import ProviderFactory, ApiConfig

p = ProviderFactory.create_provider(
    "perplexity",
    ApiConfig(api_key="YOUR_KEY")  # optionally api_base="https://proxy.."
)
out = p.completion(
    model="perplexity/llama-3.1-sonar-small-128k-chat",
    messages=[{"role": "user", "content": "hello"}],
) 
```
Closes #127


src/any_llm/providers/perplexity/perplexity.py
```diff
@@ -0,0 +1,39 @@
...
+            "PERPLEXITY_BASE_URL", self.API_BASE
+        )
+
+    def completion(
+        self,
+        model: str,
+        messages: list[dict[str, Any]],
+        **kwargs: Any,
+    ) -> ChatCompletion | Iterator[ChatCompletionChunk]:
+        return super().completion(model, messages, **kwargs)
```
daavoo: Unless I am missing something, this is not needed 

```suggestion
```

daavoo (changes requested):
Thanks for the contribution @mihirahuja1 🚀 !

It seems that the lint and tests are failing: https://github.com/mozilla-ai/any-llm/pull/176/checks 

You can find info about our recommended setup here https://github.com/mozilla-ai/any-llm/blob/main/CONTRIBUTING.md . Please let us know if there is some information missing.

Aside from that, you need to add a new dependency group and include it in the `all` under:

https://github.com/mozilla-ai/any-llm/blob/a3b0c06409e98368d380911592530579da73d854/pyproject.toml#L19

src/any_llm/providers/perplexity/perplexity.py
```diff
@@ -0,0 +1,39 @@
...
+    SUPPORTS_COMPLETION = True
+    SUPPORTS_RESPONSES = False
+    SUPPORTS_COMPLETION_REASONING = False
+    SUPPORTS_EMBEDDING = False
+
+    def __init__(self, config: ApiConfig) -> None:
+        super().__init__(config)
+        # prefer explicit ApiConfig, else env override, else default.
+        self.API_BASE = config.api_base or os.getenv(
+            "PERPLEXITY_BASE_URL", self.API_BASE
```
daavoo: btw, is this like a common pattern and env var name recommended or used elsewhere?

src/any_llm/providers/perplexity/perplexity.py
mihirahuja1: Thanks for pointing it out! Yep this is an overkill. Will remove it.

mihirahuja1:
@daavoo  Thanks for the thorough review! I've addressed all the issues you raised:

Fixed lint issues - Removed unused imports (Iterator, Any, ChatCompletion, ChatCompletionChunk)
Tests passing - All unit tests now pass 
Removed unnecessary override - Eliminated the redundant completion() method as suggested
Dependency group configured - Already added perplexity = [] 
Environment variable - Using PERPLEXITY_API_BASE for consistency with other providers (like OPENAI_API_BASE)
Integration tests updated - Added Perplexity to provider_model_map in tests/conftest.py with llama-3.1-sonar-small-128k-chat model

mihirahuja1:
Since Perplexity uses an OpenAI-compatible API through BaseOpenAIProvider, no additional dependencies are needed (hence the empty dependency list, similar to moonshot/nebius/databricks providers).
All changes have been pushed. Could you please re-run the CI checks? Everything should pass now.
Thanks again for your guidance!

src/any_llm/providers/perplexity/perplexity.py
mihirahuja1: @daavoo Good catch! I've actually already updated this to use
  PERPLEXITY_API_BASE instead of PERPLEXITY_BASE_URL for consistency
  with other providers in the codebase.

  Looking at the pattern across other providers:
  - OpenAI uses OPENAI_API_BASE
  - Other OpenAI-compatible providers follow the same
  {PROVIDER}_API_BASE pattern

mihirahuja1:
@daavoo - Looks like most of the tests that previously failed passed except Lint and Docs. 
As I understand - Docs will fail if the provider is being added for the first time. Right?
I'm checking the Lint issue - hope to have a fix for it fairly quickly.

njbrake:
Hi @mihirahuja1 ! I'll help take this over the finish line. That's correct, the docs tests will fail initially when adding a new provider.

mihirahuja1:
Thanks @njbrake for taking it over the finish line - really appreciate it! I see the lint fix you did, makes sense, will keep that in mind for future PRs.

PR #179, opened 2025-08-15
fix(huggingface): Handle `config.api_base`.

daavoo:
Argument was being ignored. Needs to be passed as `base_url` to the client.

### DIFF
--- a/src/any_llm/providers/huggingface/huggingface.py
+++ b/src/any_llm/providers/huggingface/huggingface.py
@@ -35,7 +35,7 @@ class HuggingfaceProvider(Provider):
 
     PROVIDER_NAME = "huggingface"
     ENV_API_KEY_NAME = "HF_TOKEN"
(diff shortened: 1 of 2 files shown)


src/any_llm/providers/huggingface/huggingface.py
```diff
@@ -35,7 +35,7 @@ class HuggingfaceProvider(Provider):
 
     PROVIDER_NAME = "huggingface"
     ENV_API_KEY_NAME = "HF_TOKEN"
-    PROVIDER_DOCUMENTATION_URL = "https://huggingface.co/inference-endpoints"
+    PROVIDER_DOCUMENTATION_URL = "https://huggingface.co/docs/huggingface_hub/package_reference/inference_client"
```
daavoo, line 38: The `InferenceClient` can be used for different usages beyond inference-endpoints like inference providers and local endpoints (see updated link)

daavoo:
Integration tests: https://github.com/mozilla-ai/any-llm/actions/runs/16986116299

PR #181, opened 2025-08-15
tests: Enable `huggingface` integration tests.

daavoo:
https://github.com/mozilla-ai/any-llm/actions/runs/16986708879/job/48157076250

image

### DIFF
--- a/tests/integration/test_response_format.py
+++ b/tests/integration/test_response_format.py
@@ -16,6 +16,8 @@ def test_response_format(
     provider_model_map: dict[ProviderName, str],
     provider_extra_kwargs_map: dict[ProviderName, dict[str, Any]],
 ) -> None:
+    if provider == ProviderName.HUGGINGFACE:
+        pytest.xfail("response_format is not currently working for HuggingFace")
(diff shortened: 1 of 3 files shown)


tests/integration/test_response_format.py
```diff
@@ -16,6 +16,8 @@ def test_response_format(
     provider_model_map: dict[ProviderName, str],
     provider_extra_kwargs_map: dict[ProviderName, dict[str, Any]],
 ) -> None:
+    if provider == ProviderName.HUGGINGFACE:
+        pytest.xfail("response_format is not currently working for HuggingFace")
```
daavoo, line 20: I was having issues in the any-agent integration tests but it turns out this was also not working here for this test

njbrake: Do you think we should add a new provider attribute for `SUPPORTS_RESPONSE_FORMAT`?

daavoo: I think all (our current) providers do? I just need to figure out why it is not working with hf

njbrake: Llamafile doesn't either :(

PR #185, opened 2025-08-15
fix(huggingface): Use openai's `type_to_response_format_param` to han…

daavoo:
…dle `response_format`.

Drop `_convert_pydantic_to_huggingface_json` modification of messages.

### DIFF
--- a/src/any_llm/providers/huggingface/huggingface.py
+++ b/src/any_llm/providers/huggingface/huggingface.py
@@ -3,16 +3,14 @@ from typing import TYPE_CHECKING, Any
 
 try:
     from huggingface_hub import InferenceClient
+    from openai.lib._parsing import type_to_response_format_param
 
     PACKAGES_INSTALLED = True
 except ImportError:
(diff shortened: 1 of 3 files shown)


daavoo:
Integration tests: https://github.com/mozilla-ai/any-llm/actions/runs/16989284273/job/48164718681

image

njbrake:
Can you explain a little more? It looks like you adjusted the behavior of response_format handling in huggingface, but then in the integration test you skip it as not working?

If the implementation isn't working should we remove the huggingface code that attempts to handle it, and throw an exception there too?

daavoo:
> Can you explain a little more? It looks like you adjusted the behavior of response_format handling in huggingface, but then in the integration test you skip it as not working?
> 
> If the implementation isn't working should we remove the huggingface code that attempts to handle it, and throw an exception there too?

I skipped it in a previous PR because I enabled integration tests and the implementation was failing.

In this P.R., I fixed the implementation and removed the skip (check the link and/or see the screenshot attached)

Issue #177, opened 2025-08-14
Create an exhaustive test that verifies each individual completion parameter


njbrake:
Using the built in provider iterators,

Make an exhaustive test suite that iterates over all of the possible parameters that are a part of our completion signatures to make sure they're all supported. 

This shouldn't happen every merge to main or be a part of our regular integration tests, but it should be something that we run either manually locally or as a part of creating a new release?

mihirahuja1:
@njbrake - Happy to take this if there are no takers already.

njbrake:
Sure thing! When you implement this, a few things to consider:

1. You'll need to decide whether the tests should be its own folder, (e.g. `tests/release`) or they should be a new folder in the integration tests folder (e.g. `tests/integration/parameters`) that are only activated by a specific flag, that way we won't auto run them. I lean towards the adding `tests/integration/parameters` approach, so if you think there is a better design, make sure you explain why :)
2. The code should automatically identify the parameters to test, or at least there should be some sort of meta-test to make sure that the test is checking all of the available parameters. We need to make sure that if someone adds a parameter later on, something is there to automatically check the new parameter, or to flag that the parameter needs to be added to these parameter tests
3. Refer to `tests/conftest.py` to see the existing fixtures that we have, and reuse those when possible, or feel free to improve the design. I haven't yet put too much time into optimizing the pytest fixture design.

Thanks!

njbrake:
This turned out to be a kind of crazy big task to have in a single ticket. We're gonna go with the approach where we write tests for parameters that will test the parameter across all providers as we add new features or as bugs are reported. Thanks!

PR #188, opened 2025-08-15
fix(Groq): Remove unsupported params from integration tests

### LINKED ISSUE #186: Fix Groq Unknown Field
https://github.com/mozilla-ai/any-llm/blob/e8dd6783c654855a6789035623446397eaa2dfa5/tests/integration/test_responses.py#L30

FAILED tests/integration/test_responses.py::test_responses[groq] - openai.BadRequestError: Error code: 400 - {'error': {'message': 'unknown field `max_tool_calls` in request body', 'type': 'invalid_request_error'}}

besaleli:
## Description

This PR removes unsupported fields (e.g., `max_tool_calls`, `text`, etc.) from Responses API integration tests. See https://github.com/mozilla-ai/any-llm/issues/177 for discussion on what parameters to test for each provider.

## PR Type

🐛 Bug Fix

## Relevant issues

Fixes #186

## Checklist

### DIFF
--- a/tests/integration/test_responses.py
+++ b/tests/integration/test_responses.py
@@ -27,12 +27,6 @@ def test_responses(
             **extra_kwargs,
             input_data="What's the capital of France? Please think step by step.",
             instructions="Talk like a pirate.",
-            max_tool_calls=3,
-            parallel_tool_calls=True,
-            reasoning={"effort": "medium"},
-            text={
-                "verbosity": "low",
-            },
         )
     except MissingApiKeyError:
         pytest.skip(f"{provider.value} API key not provided, skipping")


njbrake (changes requested):
Hmm. If they're unsupported, Can we instead throw the specific exception from within the provider? I like that we're testing the args, but we need some handling for the providers that don't support them. 

We need to understand:

1. Does groq not support the arg at all?
2. Does groq support the arg but they name it something different that we need to map it to

njbrake:
>> See https://github.com/mozilla-ai/any-llm/issues/177 for discussion on what parameters to test for each provider.

That PR Issue is more specifically to handle testing of Completion API parameters, not responses Api parameters.

besaleli:
@njbrake Groq responses API does not support the argument yet: https://console.groq.com/docs/api-reference#responses-create

Groq's native client doesn't have access to the responses API so the provider is using its OpenAI-compatible library, which doesn't throw an error if it's an unrecognized argument (error will come back in the REST API error). Two paths I see:

1. Continuing to fail as expected and maybe update docs
2. Hardcoding parameters we know won't work and raising an error before the client method is called

What do you think?

njbrake:
@besaleli Cool! I think we should add a check in the provider implementation https://github.com/mozilla-ai/any-llm/blob/main/src/any_llm/providers/groq/groq.py#L73 and throw the any_llm custom exception (UnsupportedParameter). 

Then the question is how do we handle it in the integration test. How about for the immediate term:

1. In this PR, Add the param check and exception throwing in the groq provider implementation as mentioned above. 
2. In this PR, drop the extra params from the responses integration test (which is what this PR already has done)
3. Create Github Issue with the task of testing all parameters for the responses API, and in that PR we'll handle the finding and management of the parameter support matrix, following whatever design comes out of #177

besaleli:
@njbrake Sounds good! here is what I did:
- [x] added param check for max_tool_calls (double checked with the docs, that's the only one it doesn't support)
- [x] dropped extra params from integration test
- [x] See attached #194

njbrake (approved):
❤️  Looks great!

PR #189, opened 2025-08-15
chore: Wake up HuggingFace endpoint before running integration tests.

daavoo:
If the endpoint has been scaled down to zero because of inactivity, the tests fail with 503 service unavailable.

### DIFF
--- a/.github/workflows/tests-integration.yaml
+++ b/.github/workflows/tests-integration.yaml
@@ -27,6 +27,10 @@ jobs:
         run: |
           uv sync --group tests --extra all
 
+      - env:
+          HF_TOKEN: ${{ secrets.HF_TOKEN }}
+        run: python scripts/wake_up_hf_endpoint.py
+
       # cache the ~/.ollama/models directory
       - uses: actions/cache@v4
         with:
@@ -42,6 +46,10 @@ jobs:
           ollama serve &
           ollama pull llama3.2:1b
           ollama pull qwen3:0.6b
+
+      - env:
+          HF_TOKEN: ${{ secrets.HF_TOKEN }}
+        run: python scripts/wake_up_hf_endpoint.py --retry=5
 
       - name: Run Integration tests (parallel with xdist)
         env:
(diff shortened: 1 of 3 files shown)


daavoo:
Integration tests: https://github.com/mozilla-ai/any-llm/actions/runs/16991639737

.github/workflows/tests-integration.yaml
```diff
@@ -23,6 +23,22 @@ jobs:
...
+          -H "Content-Type: application/json" \
+          -d '{
+              "model": "tgi",
+              "messages": [
+                  {
+                      "role": "user",
+                      "content": "Wake up!"
+                  }
+              ]
+          }'
```
njbrake: Can we move this into something that I can easily run locally too? Maybe into the scripts folder?

Also, should we change the behavior so that it will wait til the server responds 200 before finishing?

njbrake (changes requested):
I'd like the logic to go outside of the github action, and also to reference a variable for the endpoint instead of a hardcoded string

.github/workflows/tests-integration.yaml
```diff
@@ -27,6 +27,8 @@ jobs:
         run: |
           uv sync --group tests --extra all
 
+      - run: python scripts/wake_up_hf_endpoint.py
```
daavoo: This calls with no retry to trigger the wake up but don't keep waiting while we install and run ollama

daavoo:
Integration tests: https://github.com/mozilla-ai/any-llm/actions/runs/17160930810

scripts/wake_up_hf_endpoint.py
```diff
@@ -0,0 +1,33 @@
...
+from any_llm import completion
+
+HF_ENDPOINT = "https://y0okp71n85ezo5nr.us-east-1.aws.endpoints.huggingface.cloud/v1/"
+
+
+def wake_up_hf_endpoint(retry: int = 0):
+    while True:
+        try:
+            completion(
+                model="huggingface:tgi", messages=[{"role": "user", "content": "Are you awake?"}], api_base=HF_ENDPOINT
```
njbrake, line 15: 😆 🥇

PR #198, opened 2025-08-15
chore: add codecov.yml

besaleli:
Anything else we want besides ignore?


njbrake (approved):
I assume that we will want to add things later, but this is all I wanted at the moment!

PR #223, opened 2025-08-15
feat: Update BaseProvider to async-first implementation

### LINKED ISSUE #201: Update `BaseProvider` to async-first implementation
To be done on completion of #199 and #200

besaleli:
## Description
This PR updates the return type of `acompletion` and `aresponse` to `AsyncIterator` when `stream=True`.

‼️ THIS IS A BREAKING CHANGE ‼️

We need this in order to implement async-native `acompletion` and `aresponse` functions (it turns out that there is no easy way to do `AsyncIterator` -> `Iterator`).

## PR Type

🆕 New Feature

## Relevant issues

Closes #201

### DIFF
--- a/src/any_llm/api.py
+++ b/src/any_llm/api.py
@@ -1,4 +1,4 @@
-from collections.abc import Callable, Iterator
+from collections.abc import AsyncIterator, Callable, Iterator
 from typing import Any, Literal
 
 from pydantic import BaseModel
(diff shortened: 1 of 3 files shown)


njbrake (approved):
Makes total sense 👍🏼

anivar:
Hey @besaleli! 

I've been diving deep into the async migration and had an idea that might help avoid the breaking changes while still getting us proper async streaming.

What if instead of changing the return types, we add new methods alongside the existing ones? Like:
- `complete_async()` → returns `ChatCompletion` 
- `complete_stream_async()` → returns `AsyncIterator[ChatCompletionChunk]`

This way:
- No breaking changes (existing code keeps working)
- Clean return types (no unions in provider methods!)  
- Providers can migrate gradually
- We can deprecate the old methods later

The API layer would just check if a provider has the new methods and use them, otherwise fall back to current behavior. 

What do you think? Happy to help implement this if you like the approach!

Issue #226, opened 2025-08-17
Return reasoning content from anthropic provider


njbrake:
Make it so that SUPPORTS_COMPLETION_REASONING is true for anthropic

Issue #225, opened 2025-08-17
Return reasoning content from Google Provider


njbrake:
Make it so that SUPPORTS_COMPLETION_REASONING is true for google

PR #239, opened 2025-08-18
fix(api): Set default `reasoning_effort="medium"`.

daavoo:
See https://platform.openai.com/docs/api-reference/responses/create.

> effort string or null Optional Defaults to medium

### DIFF
--- a/src/any_llm/providers/cohere/cohere.py
+++ b/src/any_llm/providers/cohere/cohere.py
@@ -103,7 +103,9 @@ class CohereProvider(Provider):
         response = self.client.chat(
             model=params.model_id,
             messages=patched_messages,  # type: ignore[arg-type]
-            **params.model_dump(exclude_none=True, exclude={"model_id", "messages", "stream", "reasoning_effort", "response_format"}),
+            **params.model_dump(
+                exclude_none=True, exclude={"model_id", "messages", "stream", "reasoning_effort", "response_format"}
+            ),
             **kwargs,
         )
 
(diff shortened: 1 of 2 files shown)


daavoo:
This surfaced the issue that the param was not being handled by providers not inheriting from OpenAIBaseProvider. Taking a look

njbrake:
You're running into a task that is on my backburner but I want to get to it at some point: I'd like to have a direct mapping from our CompletionArgs model into each SDK args, instead of the hacky model_dump syntax that we currently have. Do you think I should create a github issue to track doing that?

daavoo:
> Will this mess up all the providers that don't support reasoning_effort? 

Yes, it requires to drop the value from the model_dump in providers that don't support it. I have pushed a commit.

> Its not clear to me what the motivator is for setting the default.

First, consistency with OpenAI API. 
Second, in providers where the parameter actually needs to be handled like #225 #226 (reasoning effort needs to be mapped to "thinking budget") users would need to manually set a value to explicitly enable the return of reasoning parts.

daavoo:
> I'd like to have a direct mapping from our CompletionArgs model into each SDK args, instead of the hacky model_dump syntax that we currently have. Do you think I should create a github issue to track doing that?

Yes, that would be great

njbrake:
> > I'd like to have a direct mapping from our CompletionArgs model into each SDK args, instead of the hacky model_dump syntax that we currently have. Do you think I should create a github issue to track doing that?
> 
> Yes, that would be great
Done 👍 
https://github.com/mozilla-ai/any-llm/issues/241#issue-3329916167

daavoo:
Integration tests: https://github.com/mozilla-ai/any-llm/actions/runs/17036877089

PR #250, opened 2025-08-18
feat: reasoning_effort -> map to provider defaults

njbrake:
## Description
Use the default provider settings for reasoning, by adding a "auto" literal.

This will allow us to correctly throw exceptions if a user tries to set reasoning effort on a provider that doesn't support it.

## PR Type

🆕 New Feature
🐛 Bug Fix
💅 Refactor

## Checklist

### DIFF
--- a/tests/integration/test_reasoning.py
+++ b/tests/integration/test_reasoning.py
@@ -20,7 +20,7 @@ def test_completion_reasoning(
     cls = ProviderFactory.get_provider_class(provider)
     if not cls.SUPPORTS_COMPLETION_REASONING:
         pytest.skip(f"{provider.value} does not support completion reasoning, skipping")
-        
+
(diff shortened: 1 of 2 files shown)


njbrake:
> You need to drop all the usages in model_dump like:

Sorry, can you give an example? For Portkey and Cerebras, since I set it to None, then it gets automaticlaly dropped since we have exclude_none=True in our model_dumps.

daavoo:
> Sorry, can you give an example? For Portkey and Cerebras, since I set it to None, then it gets automaticlaly dropped since we have exclude_none=True in our model_dumps.

The line I linked in the previous comment. 
I added the additional `reasoning_effort` key to `exclude` to all providers (except those that I missed that when changing the default to `medium`. 

If you now leave that key after this PR, users won't get an explicit error when setting a value different to `auto`.

PR #247, opened 2025-08-18
feat(async): Migrate Together.ai provider to async

besaleli:
## Description

NOTE: Untested until I get my hands on a Together.ai API key 🥀 they rate limited me

PR Type: 🆕 New Feature

## Relevant issues

Closes #217

## Checklist
- [x] I have added unit tests that prove my fix/feature works
- [x] New and existing tests pass locally
- [x] Documentation was updated where necessary
- [x] I have read and followed the [contribution guidelines](https://github.com/mozilla-ai/any-llm/blob/main/CONTRIBUTING.md)

PR #253, opened 2025-08-18
feat(together): Add native async support

anivar:
Adds native async support to Together provider using `AsyncTogether` client.

- Removes asyncio.to_thread overhead  
- Adds async streaming support
- Extracts common code to utils

Part of #217


anivar:
Closing in favor of #247. Suggestion: Consider extracting the response conversion logic to a utility function to avoid duplication between sync and async methods.

PR #254, opened 2025-08-18
feat(openai): Add support for custom httpx client

anivar:
Enables passing custom httpx clients to OpenAI-based providers for connection pooling.

- Added http_client parameter support to BaseOpenAIProvider methods
- Works with all providers inheriting from BaseOpenAIProvider
- Includes tests and usage example

Addresses #236

PR #255, opened 2025-08-18
feat(native-providers): Add httpx client support for native SDK providers

anivar:
Extends httpx client support to providers with native SDK support.

**Providers updated:**
- Anthropic: Uses  parameter 
- Groq: Uses  parameter
- Mistral: Uses  parameter
- Cohere: Uses  parameter

**Benefits:**
- Connection pooling across API calls
- Custom timeout/retry configuration
- HTTP/2 support where available

Builds on #254 to provide comprehensive httpx client support.


besaleli:
We appreciate the enthusiasm here. Right now we haven’t finalized whether and how we want to extend `httpx` client support across providers (see #236). Before code changes, we need consensus on design and API surface. For now we’ll close this PR and keep the discussion in the issue.

src/any_llm/providers/cohere/cohere.py
```diff
@@ -39,12 +39,19 @@ class CohereProvider(Provider):
     def __init__(self, config: ApiConfig) -> None:
         """Initialize Cohere provider."""
         super().__init__(config)
-        self.client = cohere.ClientV2(api_key=config.api_key)
+        # Note: httpx_client must be passed at method call time for Cohere
```
daavoo, line 44: Nice. 

I think this surfaces a good point for discussion. 

In this case, it seems that cohere was the outlier being implemented inconsistently but I do wonder if we should put the client instantiation during __init__ instead of inside each method (and handle kwargs appropiately) for all providers.

My intuition thinks that the internal http_client is what matters, and we would be handle it with this PR,  but I am not sure what is the overhead of recreating the "native sdk client" inside each `completion/embedding/x` call.

tests/unit/providers/test_native_httpx_client.py
```diff
@@ -0,0 +1,189 @@
...
+        messages=[{"role": "user", "content": "test"}],
+    )
+
+    # Patch the client constructor to verify http_client is passed
+    with patch(module_path) as mock_client_class:
+        mock_client = MagicMock()
+        mock_client_class.return_value = mock_client
+
+        # Set up mock response based on provider
+        if provider == ProviderName.ANTHROPIC:
```
daavoo: Not to be handled in this PR: It would be probably nice to have some shared fixtures in conftest or something mocking "native provider responses" to be reused in the unit tests

tests/unit/providers/test_native_httpx_client.py
```diff
@@ -0,0 +1,189 @@
...
+        model_id="test-model",
+        messages=[{"role": "user", "content": "test"}],
+    )
+
+    # Patch the client constructor to verify http_client is passed
+    with patch(module_path) as mock_client_class:
+        mock_client = MagicMock()
+        mock_client_class.return_value = mock_client
+
+        # Set up mock response based on provider
```
daavoo: Since we are only caring about the client instantiation, what we could do is to mock the `_convert_response` or whatever from each provider, so we don't have to create a mock response

tests/unit/providers/test_native_httpx_client.py
daavoo, line 1: Not a strong opinion but we could take and move the test to each providers `test_unit_{provider}.py` where the internal client mocks are already implemented and can be reused.

anivar:
Thanks for the review @daavoo\! I've investigated the missing providers as requested:

**Providers with native httpx support:**
- Anthropic: http_client parameter
- Groq: http_client parameter  
- Mistral: client parameter (different name)
- Cohere: httpx_client parameter (different name)
- Google: HttpOptions class with httpx by default, configurable via http_options parameter
- Watson: httpx_client parameter in ModelInference (accepts httpx.Client or HttpClientConfig)

**Providers without direct httpx support:**
- AWS (boto3): Uses urllib3 internally. No direct httpx support in standard boto3, but aiobotocore has experimental httpx support via http_session_cls parameter
- XAI: Uses gRPC protocol, not HTTP/httpx

Key observations:
1. Most native SDK providers (6/8) support custom httpx clients
2. Parameter names vary: http_client, httpx_client, client, http_options
3. Google uses httpx by default in their new GenAI SDK
4. AWS's async variant (aiobotocore) has experimental httpx support

This confirms that almost all providers expose similar functionality for connection pooling and custom HTTP client configuration.

Regarding the client instantiation point you raised - that's a great observation. The overhead of recreating SDK clients on each call could indeed be significant. Cohere is currently the outlier requiring instantiation at method call time. I'm happy to explore moving client instantiation to __init__ if that's the preferred direction.

src/any_llm/providers/cohere/cohere.py
anivar: Good point about the overhead. I checked Cohere's SDK and it seems they don't support passing httpx_client at initialization time - it must be passed when creating the client. This is why I had to instantiate it in each method.

However, we could explore caching the client instance with a given httpx_client to avoid recreation. Something like storing a mapping of httpx_client -> cohere.Client instances. What do you think?

tests/unit/providers/test_native_httpx_client.py
anivar: Agreed, shared fixtures would be helpful for future tests. We could create a conftest.py with fixtures for common response patterns across providers separately

tests/unit/providers/test_native_httpx_client.py
anivar: Good suggestion! You're right that we only need to verify the http_client parameter is passed correctly to the SDK client constructor. I can update the tests to mock the _convert_response functions and remove the mock response setup. This will make the tests simpler and more focused on what we're actually testing - the client instantiation

anivar:
@daavoo @njbrake I've addressed all the feedback and investigated the additional providers as requested. The analysis shows that 6/8 native SDK providers support custom httpx clients.

Happy to explore moving client instantiation to __init__ if that's the preferred approach. Would appreciate your thoughts on the design direction\!

PR #260, opened 2025-08-19
chore(anthropic): Increase `DEFAULT_MAX_TOKENS` from `4096` to `8192`.

daavoo:
The current default is way below the capacity of any of the provided models:

https://docs.anthropic.com/en/docs/about-claude/models/overview#model-comparison-table

I updated the value to match the smallest model capacity (claude-3-5-haiku).

## Checklist


daavoo:
I found this while working on #226 because `budget_tokens` for thinking can't be higher than `max_tokens`

PR #345, opened 2025-08-24
Feat/openrouter reasoning 291

### LINKED ISSUE #291: OpenRouter Completion API Reasoning Support

mihirahuja1:
## Description
Adds **reasoning support** to the OpenRouter provider.

- Supports unified `reasoning` object (`effort`, `max_tokens`, `exclude`, `enabled`)
- Legacy `include_reasoning=True|False` toggle still works
- Full support for **streaming reasoning deltas**
- Uses `extra_body["reasoning"]` with explicit opt-in to avoid extra cost/latency

## PR Type
🆕 New Feature

## Relevant issues
Fixes #291

## Usage
```python
await acompletion(
  model="openrouter:deepseek/deepseek-r1:free",
  messages=[{"role":"user","content":"Is 97 prime? Think step by step."}],
  reasoning={"effort": "medium", "max_tokens": 512}
)

# Legacy toggle
await acompletion(model="openrouter:…", messages=[…], include_reasoning=True)

# Exclude reasoning
await acompletion(model="openrouter:…", messages=[…], include_reasoning=False)
```

## Checklist


mihirahuja1:
Fixes #291

njbrake (changes requested):
Hi @mihirahuja1 thanks for the work on this! This needs a little bit more work and maybe discussion as well.

the acompletion api exposes a `reasoning_effort` parameter, which should be used here. See the other providers for an example of how it's used: we default it to "auto" which lets you set the default for the provider as seems reasonable (for openrouter, I'm thinking that in the "auto" scenario the reasoning info should be excluded).

Please avoid getattr type calls whenever possible and use properly typed functions instead. 

If there are small extra configurations that openrouter has available that aren't common to other providers, any-llm will already pass them through to openrouter through the kwargs dict, so I think as long as we focus on connecting the existing `reasoning_effort` parameter correctly, we can let the users pass in any more custom openrouter logic through the **kwargs.

Thanks!

mihirahuja1:
Thanks @njbrake , appreciate the feedback! I want to make sure I’m aligned before I rework things:
- For OpenRouter, would you expect reasoning_effort="auto" to always exclude reasoning by default, and then only include it if set explicitly (e.g. "medium", "high")?
-  Do we still want to support the legacy include_reasoning flag for backward compatibility, or should we move entirely to the unified reasoning_effort approach?

njbrake:
>     * For OpenRouter, would you expect reasoning_effort="auto" to always exclude reasoning by default, and then only include it if set explicitly (e.g. "medium", "high")?

Yes, I think this would be a reasonable default, if you think that makes sense?

>     * Do we still want to support the legacy include_reasoning flag for backward compatibility, or should we move entirely to the unified reasoning_effort approach?

My gut reaction would be to say we should drop the `include_reasoning` flag, unless you have a compelling reason why you think it should stay. My general approach recommendation would be to make this initial approach as simple as possible.


Thanks for helping out on this!

mihirahuja1:
Not stale — changes are in progress, pushing fixes for conflicts and coverage soon.

mihirahuja1:
@njbrake 
- Removed legacy `include_reasoning`
- Use only `reasoning_effort`; default `auto` excludes reasoning (no extra_body sent)
- No getattr; typed helper builds OpenRouter reasoning
- Streaming path untouched and tested
- Repo-wide ruff/mypy clean; added tests covering auto/no-extra-body, effort levels, explicit override, and extra_body merge

Rebased on main

mihirahuja1:
Will work on resolving conflicts.

daavoo:
hei @mihirahuja1 , thanks for the contribution!
I will take care of solving the final conflicts to get this merged

mihirahuja1:
> hei @mihirahuja1 , thanks for the contribution! I will take care of solving the final conflicts to get this merged

Thanks @daavoo for resolving the conflicts.

PR #348, opened 2025-08-25
Fix(provider): Correctly parse model names with tags

### LINKED ISSUE #347: [BUG] Model parsing logic fails for models that included a tag
**`any-llm` version:** `0.16.0`
**Python Version**: `3.11.10`

### Describe the bug

Model parsing logic fails for models that included  a tag, such as `ollama/gemma3:4b`. The string is split at the first colon, incorrectly separating the model name and tag. In this example, the provider would be `ollama/gemma3`—leading to an "unsupported provider" error—while the model would be `4b`.
    
We should fix this until we deprecate `/` as a provider-model separator.

dpoulopoulos:
## Description

This pull request addresses a bug in the model parsing logic that caused incorrect parsing of model names that include a tag, such as `ollama/gemma3:4b`.

Problem:

The previous implementation split the model string at the first colon (:), which resulted in the provider being incorrectly identified as `ollama/gemma3` and the model name as `4b`.

Solution:

This patch modifies the parsing order to first check for the legacy `provider/model` separator (``/) before checking for the `provider:model` separator (`:`). This ensures that model names in the format `<provider><model_name>:<model_tag>` are parsed correctly.

This change is backward compatible and does not affect the parsing of model names that do not include a tag.

## PR Type

🐛 Bug Fix

## Relevant issues

Fixes #347

## Checklist

### DIFF
--- a/src/any_llm/provider.py
+++ b/src/any_llm/provider.py
@@ -344,9 +344,7 @@ class ProviderFactory:
         The legacy format will be deprecated in version 1.0.
         """
         # Check for new colon syntax first
-        if ":" in model:
-            provider, model_name = model.split(":", 1)
-        elif "/" in model:
+        if "/" in model:
             # Legacy slash syntax with deprecation warning
             warnings.warn(
                 f"Model format 'provider/model' is deprecated and will be removed in version 1.0. "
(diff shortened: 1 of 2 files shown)


besaleli (approved):
lgtm 🚀

PR #349, opened 2025-08-25
refactor: Replace `PACKAGES_INSTALLED` with `MISSING_PACKAGES_ERROR`.

daavoo:
In order to preserve the original exception that caused the ImportError.

We were hiding potentially useful information to the users facing the error. For example, in https://github.com/mozilla-ai/any-llm/pull/238 I had to run with the debugger to discover what was wrong because just running the recommended command didn't work.

- Added `test_providers_raise_ImportError_from_original`.

Along the way, I fixed several inconsistencies:

- Wrap `utils` imports inside MISSING_PACKAGES_ERROR block.
- Use `if TYPE_CHECKING` blocks.
- Don't raise separate ImportErrors from `utils`.
- Extract `_verify_no_missing_packages` to be used in providers that don't call `super().__init__` (these providers were not raising the ImportErrors as expected before this PR).


daavoo:
Need to figure out why the test passes if you run it individually but fails when running alongside the whole suite

njbrake (dismissed):
Excellent improvement 🙏

daavoo:
Integration tests: https://github.com/mozilla-ai/any-llm/actions/runs/17206608870

PR #360, opened 2025-08-26
fix(aio): Don't raise RuntimeError if `INSIDE_NOTEBOOK`.

daavoo:
- Move `RuntimeError` handling to inside `run_async_in_sync`. No need to repeat the `get_running_loop` block.


daavoo:
Integration tests: https://github.com/mozilla-ai/any-llm/actions/runs/17236034904

PR #362, opened 2025-08-26
tests: Add `AZUREOPENAI` placeholder.

daavoo:
It is failing on main because of missing key: https://github.com/mozilla-ai/any-llm/actions/runs/17237406658

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -53,6 +53,7 @@ def provider_model_map() -> dict[ProviderName, str]:
         ProviderName.PORTKEY: "@first-integrati-d8a10f/gpt-4.1-mini",  # Owned by njbrake in portkey UI
         ProviderName.LLAMA: "Llama-4-Maverick-17B-128E-Instruct-FP8",
         ProviderName.AZURE: "openai/gpt-4.1-nano",
+        ProviderName.AZUREOPENAI: "azure/<your_deployment_name>",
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -53,6 +53,7 @@ def provider_model_map() -> dict[ProviderName, str]:
         ProviderName.PORTKEY: "@first-integrati-d8a10f/gpt-4.1-mini",  # Owned by njbrake in portkey UI
         ProviderName.LLAMA: "Llama-4-Maverick-17B-128E-Instruct-FP8",
         ProviderName.AZURE: "openai/gpt-4.1-nano",
+        ProviderName.AZUREOPENAI: "azure/<your_deployment_name>",
```
daavoo, line 57: @njbrake / @besaleli not super critical but it feels annoying to have this separate dicts in the conftest. I don't know if it make sense to have like this values stored at the class level like `TEST_COMPLETION_MODEL` or something so when someone adds a new provider they can set that and if it's not set then we skip the test?

daavoo:
Integration tests: https://github.com/mozilla-ai/any-llm/actions/runs/17238881195

image

tests/conftest.py
njbrake: The conftest needs some cleanup for sure. It's on the backburner but I'll make a ticket (maybe we'll get an outside contributor willing to help!)

PR #371, opened 2025-09-01
refactor(api): Expose and handle `client_args`.

daavoo:
Users were not able to pass custom args to the client instantiation.

Make it part of `CompletionParams` for `(a)completion` and a explicit kwarg for the rest of methods.

### DIFF
--- a/src/any_llm/api.py
+++ b/src/any_llm/api.py
@@ -39,6 +39,7 @@ def completion(
     stream_options: dict[str, Any] | None = None,
     max_completion_tokens: int | None = None,
     reasoning_effort: Literal["minimal", "low", "medium", "high", "auto"] | None = "auto",
+    client_args: dict[str, Any] | None = None,
     **kwargs: Any,
 ) -> ChatCompletion | Iterator[ChatCompletionChunk]:
     """Create a chat completion.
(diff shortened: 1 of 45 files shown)


daavoo:
integration tests: https://github.com/mozilla-ai/any-llm/actions/runs/17378036728

njbrake:
@daavoo Would it be a better design to adjust the ApiConfig object to put the client args in that instead of the CompletionArgs? 

In my head that seems like a cleaner design but I'm not sure if you know of a reason why that is a bad idea.

daavoo:
> @daavoo Would it be a better design to adjust the ApiConfig object to put the client args in that instead of the CompletionArgs?
> 
> In my head that seems like a cleaner design but I'm not sure if you know of a reason why that is a bad idea.

You are absolutely right. I guess it might make sense to rename it to `ClientConfig` instead?

njbrake:
> I guess it might make sense to rename it to ClientConfig instead?

Either that or ApiClientConfig, whichever you think is more appropriate.

daavoo:
> > I guess it might make sense to rename it to ClientConfig instead?
> 
> Either that or ApiClientConfig, whichever you think is more appropriate.

went with `ClientConfig` just beacuse it is shorter, please TAL

daavoo:
Integration tests: https://github.com/mozilla-ai/any-llm/actions/runs/17402536695

PR #376, opened 2025-09-04
Fix the separation between reasoning and answer for Ollama

JosephChataignon:
The separation was done wrong, saving the part that is BEFORE the <think> tag instead of the part AFTER the </think> closing tag. As a result, the message content was empty.

Note: this relies on the assumption that model reasoning always follows the pattern "\<think\> some reasoning \<\/think\> answer". To my knowledge this is the case everywhere.

Note 2: maybe it would be useful to have some kind of escape mechanism, as there are cases where this would break.

## PR Type

🐛 Bug Fix

## Checklist

I wrote a fix quickly because the issue in the logic is evident and it didn't work in my real-use case. I haven't run the tests locally, and I don't understand how this passed the reasoning integration test in the first place.


njbrake:
> I haven't run the tests locally, and I don't understand how this passed the reasoning integration test in the first place.

It passed the integration test because of a quirk in ollama: with many of their models (e.g. gpt-oss) ollama correctly parses the reasoning content into their "thinking" chunk. Our logic with the bug that you discovered was our "fallback" method to try to catch the reasoning content if ollama missed it. Unfortunately we didn't have any testing around this fallback method, which is why you were the unlucky person to encounter it first. Since this is bit of an annoying bug, I can pick this up to help you get it over the finish line. Thanks for finding the issue and reporting it! 🙏

JosephChataignon:
@njbrake  Hi, I couldn't respond last week but actually I'd be interested in contributing to the repo. Is there something simple I can do, like other unit tests ?

njbrake:
Hi @JosephChataignon , help is welcome! See this page: https://github.com/mozilla-ai/any-llm/issues?q=is%3Aissue%20state%3Aopen%20no%3Aassignee if you comment on the issue that you would like to work on, we can assign it to you and appreciate contributions 🙏 . Be sure to read https://github.com/mozilla-ai/any-llm/blob/main/CONTRIBUTING.md for an understanding of contributing guidelines.

PR #378, opened 2025-09-04
manually run can pass args

njbrake:
## Description

Sometime you only make a change for one provider and you want to run the integration test just on that feature or provider.

## Checklist

### DIFF
--- a/.github/workflows/tests-integration.yaml
+++ b/.github/workflows/tests-integration.yaml
@@ -9,6 +9,10 @@ on:
       - '.github/workflows/**'
       - 'pyproject.toml'
   workflow_dispatch:
+    inputs:
+      filter:
+        description: the arg you want to pass to pytest filter (-k)
+        type: string
 
 jobs:
   run-integration-tests:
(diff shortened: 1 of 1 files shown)


.github/workflows/tests-integration.yaml
```diff
@@ -66,7 +70,12 @@ jobs:
           VOYAGE_API_KEY: ${{ secrets.VOYAGE_API_KEY }}
           WATSONX_API_KEY: ${{ secrets.WATSONX_API_KEY }}
           XAI_API_KEY: ${{ secrets.XAI_API_KEY }}
-        run: pytest tests/integration -v -n auto --cov --cov-report=xml
+        run: |
+          if [ -n "${{ inputs.filter }}" ]; then
+            pytest tests/integration -v -n auto --cov --cov-report=xml -k "${{ inputs.filter }}"
+          else
+            pytest tests/integration -v -n auto --cov --cov-report=xml
+          fi
```
daavoo, line 78: Probably doesn't matter but I think this would also work because empty string matches everything

```suggestion
        run: |
            pytest tests/integration -v -n auto --cov --cov-report=xml -k "${{ inputs.filter }}"
```

PR #379, opened 2025-09-04
fix(aws): `kwargs` were not being handled.

daavoo:
- Refactor all logic into `_convert_params`.
- Fix `inferenceConfig` args were not being handled properly (i.e. `max_tokens` -> `maxTokens`).
- Fix `tool_choice` not handled.

### DIFF
--- a/src/any_llm/providers/aws/aws.py
+++ b/src/any_llm/providers/aws/aws.py
@@ -20,8 +20,7 @@ try:
     import instructor
 
     from .utils import (
-        _convert_kwargs,
-        _convert_messages,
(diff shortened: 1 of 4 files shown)


daavoo:
Came here because kwargs were not working, ended up doing some humanization/refactor

njbrake (approved):
LGTM I assume this bug was missed because we don't have that integration test that tries out and verifies all the arguments for all the providers

daavoo:
> I assume this bug was missed because we don't have that integration test that tries out and verifies all the arguments for all the providers

correct

PR #401, opened 2025-09-05
feat: Split google into gemini and vertexai

njbrake:
## Description

We had the "google" provider grouping both gemini and vertex ai. This PR splits them into being separate providers

## Checklist


njbrake:
Integration tests: https://github.com/mozilla-ai/any-llm/actions/runs/17496041001 

We are flying blind for VertexAI because we don't have the API keys for that yet. But we do have them for Gemini so at least that part is tested in the integration test.

This PR should also make it easier to help users debug, since then we'll know right away whether they're using vertex or gemini

PR #409, opened 2025-09-08
fix: accept an expected_providers env var

njbrake:
## Description

This PR adds support for an env var that gets passed between integration test and the github action to not let tests fail if the key is expected to be there. This helps prevent tests being silently skipped. 

## Checklist


njbrake:
Integration Tests: https://github.com/mozilla-ai/any-llm/actions/runs/17546365677 

WatsonX tests fail from rate limiting but all else is fine

PR #410, opened 2025-09-08
fix(gemin): Implement `_verify_and_set_api_key`.

daavoo:
In #401 we stopped overriding `Provider.__init__` and thus we started to rely again on the default `Provider._verify_and_set_api_key`.

We need to provide a custom implementation that checks for both API key options instead of the plain string with a `/`.

- Added `test_gemini_initialization_with_env_var_api_key` and `test_vertexai_initialization_with_env_var_api_key`.

### DIFF
--- a/src/any_llm/providers/gemini/gemini.py
+++ b/src/any_llm/providers/gemini/gemini.py
@@ -15,12 +15,16 @@ class GeminiProvider(GoogleProvider):
     PROVIDER_DOCUMENTATION_URL = "https://ai.google.dev/gemini-api/docs"
     ENV_API_KEY_NAME = "GEMINI_API_KEY/GOOGLE_API_KEY"
 
+    def _verify_and_set_api_key(self, config: ClientConfig) -> ClientConfig:
+        # Standardized API key handling. Splitting into its own function so that providers
+        # Can easily override this method if they don't want verification (for instance, LMStudio)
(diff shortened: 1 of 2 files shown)


src/any_llm/providers/gemini/gemini.py
```diff
@@ -15,12 +15,16 @@ class GeminiProvider(GoogleProvider):
     PROVIDER_DOCUMENTATION_URL = "https://ai.google.dev/gemini-api/docs"
     ENV_API_KEY_NAME = "GEMINI_API_KEY/GOOGLE_API_KEY"
 
-    def _get_client(self, config: ClientConfig) -> "genai.Client":
-        """Get Gemini API client."""
-        api_key = getattr(config, "api_key", None) or os.getenv("GEMINI_API_KEY") or os.getenv("GOOGLE_API_KEY")
+    def _verify_and_set_api_key(self, config: ClientConfig) -> ClientConfig:
+        # Standardized API key handling. Splitting into its own function so that providers
+        # Can easily override this method if they don't want verification (for instance, LMStudio)
```
njbrake:
```suggestion
```

daavoo:
Integration tests: https://github.com/mozilla-ai/any-llm/actions/runs/17545926927/job/49827279927

image

PR #417, opened 2025-09-09
feat: support for passing images into completion api

njbrake:
## Description

Add a new attribute to notate whether a provider allows you to pass an image into their completion api.

Some providers don't support it because they don't host any image capable models (e.g. Deepseek, xAI) while others may support it but their APIs diverge from the OpenAI Spec so  it will take more work to get them up and running (e.g. Anthropic).

I made new github issues for those providers that I think do support images but need more work to have their APIs become compatible with ours.

## Checklist


besaleli (approved):

njbrake:
Integration tests: https://github.com/mozilla-ai/any-llm/actions/runs/17577763576

PR #425, opened 2025-09-09
fix: parallel tool call handling

njbrake:
## Description

Fix bug where parallel tool calls weren't being added to the response

## Checklist

### DIFF
--- a/src/any_llm/providers/gemini/gemini.py
+++ b/src/any_llm/providers/gemini/gemini.py
@@ -1,9 +1,12 @@
 import os
+from collections.abc import AsyncIterator
+from typing import Any
 
 from google import genai
 
 from any_llm.exceptions import MissingApiKeyError
 from any_llm.provider import ClientConfig
+from any_llm.types.completion import ChatCompletion, ChatCompletionChunk, CompletionParams
 
 from .base import GoogleProvider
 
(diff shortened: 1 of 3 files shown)


src/any_llm/providers/gemini/gemini.py
```diff
@@ -26,3 +29,10 @@ def _verify_and_set_api_key(self, config: ClientConfig) -> ClientConfig:
     def _get_client(self, config: ClientConfig) -> "genai.Client":
         """Get Gemini API client."""
         return genai.Client(api_key=config.api_key, **(config.client_args if config.client_args else {}))
+
+    async def acompletion(
```
daavoo: I feel like I am missing something 😅

daavoo (approved):
Afaik we don't have an integration tests for multiple/parallel tool calls  (would it be worth adding?), but the existing unit test makes sense to me

src/any_llm/providers/gemini/gemini.py
njbrake: Dangit, I refactored and forgot to remove the now uneccessary call 😰 Sorry about that

njbrake:
Integration tests: https://github.com/mozilla-ai/any-llm/actions/runs/17588156428

PR #435, opened 2025-09-11
tests: Avoid waking up hf endpoint if `filter` was provided and doesn't need it.

daavoo:


daavoo:
(with `openai` as filter) https://github.com/mozilla-ai/any-llm/actions/runs/17644522051/job/50139131190

image

PR #441, opened 2025-09-16
chore: Drop `run-mypy.sh` in favor of `un run mypy`.

daavoo:


njbrake:
@daavoo can you explain what this change did? I'm trying to wrap my head around what happened 😅  

On my side now if I run `uv run pre-commit run --all-files` and then try again I get

```
(any-llm) *[543-azure-jsonschemaformat][~/scm/any-llm]$ uv run pre-commit run --all-files
error: Failed to parse `uv.lock`
  Caused by: TOML parse error at line 181, column 1
    |
181 | [[package]]
    | ^^^^^^^^^^^
missing field `version`
```

Not sure if this PR is what started it being a problem for me.

daavoo:
> @daavoo can you explain what this change did? I'm trying to wrap my head around what happened 😅

Before the change, a separate script was being run via bash and installing the full environment (run-mypy.sh)

After the change, mypy runs via UV using the same environment.

> On my side now if I run `uv run pre-commit run --all-files` and then try again I get
> 
> ```
> (any-llm) *[543-azure-jsonschemaformat][~/scm/any-llm]$ uv run pre-commit run --all-files
> error: Failed to parse `uv.lock`
>   Caused by: TOML parse error at line 181, column 1
>     |
> 181 | [[package]]
>     | ^^^^^^^^^^^
> missing field `version`
> ```
> 
> Not sure if this PR is what started it being a problem for me.

I have no idea where this error is coming from, do you have a local `uv.lock`? Did you recreating the `.venv` ?

njbrake:
I did recreate the .venv. I'll try removing my pre-commit cache to see if that fixes things 🤷 . I'll let you know if i still get stuck but overall swapping to be able to use the same mypy env sounds like a good move forward 👍

PR #444, opened 2025-09-16
refactor: Introduce `_init_client` for instantiating client once.

daavoo:
Instead of re-instantiating the client on each call.


daavoo:
Integration tests: https://github.com/mozilla-ai/any-llm/actions/runs/17795463479

PR #454, opened 2025-09-17
refactor(together): Drop `instructor` and use native response_format …

daavoo:
…instead.

According to https://docs.together.ai/docs/json-mode


daavoo:
Can't run integration tests until https://github.com/mozilla-ai/infrastructure/issues/50

PR #455, opened 2025-09-17
refactor(sambanova): Drop `instructor` usage.

daavoo:
Native `response_format` is supposed to work with the openai-compatible API: https://docs.sambanova.ai/docs/en/api-reference/endpoints/chat#function-calling-parameters

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -72,10 +72,6 @@ together = [
   "together",
 ]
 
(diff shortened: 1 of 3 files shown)


daavoo:
Can't run integration tests until https://github.com/mozilla-ai/infrastructure/issues/49

PR #463, opened 2025-09-18
fix(azureopenai): Drop unused `_get_client`. Use `super()._init_clien…

daavoo:
…t()` instead.

- Drop `AZURE_OPENAI_API_BASE` env var.

### DIFF
--- a/src/any_llm/providers/azureopenai/azureopenai.py
+++ b/src/any_llm/providers/azureopenai/azureopenai.py
@@ -1,9 +1,3 @@
-"""Azure OpenAI AnyLLM."""
-
-import os
-
-from openai import AsyncOpenAI, OpenAI
-
 from any_llm.providers.openai.base import BaseOpenAIProvider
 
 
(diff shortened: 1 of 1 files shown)


src/any_llm/providers/azureopenai/azureopenai.py
```diff
@@ -16,11 +10,8 @@ class AzureopenaiProvider(BaseOpenAIProvider):
...
-        return _client_class(
-            base_url=self.config.api_base or self.API_BASE or os.getenv("AZURE_OPENAI_API_BASE"),
-            api_key=self.config.api_key,
-            **(self.config.client_args if self.config.client_args else {}),
-            default_query={"api-version": "preview"},
-        )
+    def _init_client(self) -> None:
+        if not self.config.client_args:
+            self.config.client_args = {}
+        self.config.client_args["default_query"] = {"api-version": "preview"}
```
daavoo, line 16: I am not really sure where this argument is coming from but I left it to match previous behavior.

PR #479, opened 2025-09-19
tests: Fix `test_list_models` for `huggingface`.

daavoo:
We don't want to use the `api_base` used for completion.


daavoo:
Integration tests: https://github.com/mozilla-ai/any-llm/actions/runs/17853465593

PR #493, opened 2025-09-29
fix(openrouter): Only override `_convert_completion_params`.

daavoo:
Fixes usage with `response_format` and `tools`.


daavoo:
Integration tests: https://github.com/mozilla-ai/any-llm/actions/runs/18092990118/job/51477711864

Issue #490, opened 2025-09-24
gemini _convert_tool_spec function not compatible with gemini google_search tool format


Elevioux:
gemini doc: https://ai.google.dev/gemini-api/docs/google-search

daavoo:
hi @Elevioux thanks for the report, this should be fixed in https://github.com/mozilla-ai/any-llm/pull/499 and will be available on the next release:

```python
from any_llm import AnyLLM
from google.genai import types

llm = AnyLLM.create("gemini", api_key=api_key)

model = "gemini-2.5-flash"
messages = [{"role": "user", "content": "Search the web to find what is the any-llm python library"}]
tools = [types.Tool(google_search=types.GoogleSearch())]

response = llm.completion(model=model, messages=messages, tools=tools)
```

PR #497, opened 2025-09-30
feat: Add support for `BUILT_IN_TOOLS`.

daavoo:
Expose AnyLLM.BUILT_IN_TOOLS class var to allow providers to specify classes that will not be converted during prepare_tools.

This is pre-requisite for #490


daavoo:
Integration tests: https://github.com/mozilla-ai/any-llm/actions/runs/18125605327

PR #535, opened 2025-10-10
feat(mistral): add support for OpenAI-style response format dictionaries with Mistral

HareeshBahuleyan:
Current support is limited to `params.response_format` being a Pydantic Model. This PR enables using openAI schema. Will be useful when migrating openai agents framework to use any-llm: https://github.com/mozilla-ai/any-agent/pull/828

### DIFF
--- a/tests/unit/providers/test_mistral_provider.py
+++ b/tests/unit/providers/test_mistral_provider.py
@@ -103,31 +103,38 @@ class StructuredOutput(BaseModel):
     bar: int
 
 
+openai_json_schema = {
+    "type": "json_schema",
+    "json_schema": {
+        "name": "StructuredOutput",
+        "schema": {**StructuredOutput.model_json_schema(), "additionalProperties": False},
+        "strict": True,
+    },
+}
+
+
 @pytest.mark.asyncio
 @pytest.mark.parametrize(
-    ("response_format", "expected_converter"),
+    "response_format",
     [
-        (StructuredOutput, "response_format_from_pydantic_model"),
-        ({"type": "json_object", "schema": StructuredOutput.model_json_schema()}, "ResponseFormat"),
+        StructuredOutput,
+        openai_json_schema,
     ],
     ids=["pydantic_model", "openai_json_schema"],
 )
-async def test_response_format_conversion(response_format: Any, expected_converter: str) -> None:
+async def test_response_format(response_format: Any) -> None:
     """Test that response_format is properly converted for both Pydantic and dict formats."""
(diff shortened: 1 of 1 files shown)


HareeshBahuleyan:
Ran CI tests:
https://github.com/mozilla-ai/any-llm/actions/runs/18411357927/job/52464763632
The failures observed there I guess are unrelated - they seem to be for other providers, not Mistral, which is being changed in this PR.

HareeshBahuleyan:
@daavoo A unit test has been added now 👍

tests/unit/providers/test_mistral_provider.py
```diff
@@ -91,3 +96,45 @@ def test_patch_messages_with_multiple_valid_tool_calls() -> None:
...
+    from any_llm.providers.mistral.mistral import MistralProvider
+
+    class StructuredOutput(BaseModel):
+        foo: str
+        bar: int
+
+    with (
+        patch("any_llm.providers.mistral.mistral.Mistral") as mocked_mistral,
+        patch("any_llm.providers.mistral.mistral.response_format_from_pydantic_model") as mocked_pydantic_converter,
+        patch("any_llm.providers.mistral.mistral.ResponseFormat") as mocked_response_format,
```
daavoo: @HareeshBahuleyan I think we should not be mocking this mistral functions, we should actually try to see if they work with the inputs we are passing.

Then, we can assert on the arguments passed to `mocked_mistralchat.complete_async`
```suggestion
```

tests/unit/providers/test_mistral_provider.py
```diff
@@ -91,3 +96,45 @@ def test_patch_messages_with_multiple_valid_tool_calls() -> None:
         {"role": "assistant", "content": "OK"},
         {"role": "user", "content": "u1"},
     ]
+
+
+@pytest.mark.asyncio
+async def test_response_format_accepts_pydantic_and_openai_json_schema() -> None:
```
daavoo: not a blocker: we could use pytest.parametrize instead of having the double calls inside.

Because right now both cases run under the same test, the signal regarding where is the problem gets less clear

HareeshBahuleyan: Thanks, done

tests/unit/providers/test_mistral_provider.py
HareeshBahuleyan: Sure, updated the code to actually call the functions and verify that they pass the `response_format` argument with the expected schema to the completion() function.

PR #544, opened 2025-10-15
feat(azure): support dict input with json_schema

njbrake:
## Description

Azure requires the JsonSchemaFormat in all cases, it doesn't accept json

## Checklist

### DIFF
--- a/src/any_llm/providers/azure/utils.py
+++ b/src/any_llm/providers/azure/utils.py
@@ -38,12 +38,12 @@ def _convert_response_format(
             raise ValueError(err_msg)
         json_schema: dict[str, Any] = response_format.get("json_schema", {})
 
-        if not json_schema:
-            err_msg = "Response format must be a Pydantic model or a dict with a json_schema key"
+        if not json_schema or "schema" not in json_schema:
+            err_msg = "Response format must be a Pydantic model or a dict with a json_schema key containing a schema"
             raise ValueError(err_msg)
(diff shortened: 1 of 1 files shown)


src/any_llm/providers/azure/utils.py
```diff
@@ -33,7 +33,20 @@ def _convert_response_format(
...
+        json_schema = response_format.get("json_schema")
+
+        if not json_schema:
+            err_msg = "Response format must be a Pydantic model or a dict with a json_schema key"
+            raise ValueError(err_msg)
+        return JsonSchemaFormat(
+            name=response_format.get("json_schema", {}).get("name", ""),
+            schema=response_format.get("json_schema", {}).get("schema"),
+            description=response_format.get("json_schema", {}).get("description", ""),
+            strict=response_format.get("json_schema", {}).get("strict", True),
```
daavoo: 💅 

```suggestion
        json_schema: dict[str, Any] = response_format.get("json_schema", {})

        if not json_schema:
            err_msg = "Response format must be a Pydantic model or a dict with a json_schema key"
            raise ValueError(err_msg)
        return JsonSchemaFormat(
            name=json_schema.get("name", ""),
            schema=json_schema.get("schema"),
            description=json_schema.get("description", ""),
            strict=json_schema.get("strict", True),
```

daavoo (dismissed):

PR #558, opened 2025-10-17
feat(huggingface): support reasoning tokens

njbrake:
## Checklist

### DIFF
--- a/tests/unit/providers/test_huggingface_provider.py
+++ b/tests/unit/providers/test_huggingface_provider.py
@@ -1,10 +1,12 @@
+from collections.abc import AsyncGenerator
 from contextlib import contextmanager
+from typing import Any
 from unittest.mock import AsyncMock, patch
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


njbrake:
Integration tests: https://github.com/mozilla-ai/any-llm/actions/runs/18596097174

src/any_llm/providers/openai/utils.py
```diff
@@ -13,11 +14,7 @@ def _normalize_reasoning_on_message(message_dict: dict[str, Any]) -> None:
     if isinstance(message_dict.get("reasoning"), dict) and "content" in message_dict["reasoning"]:
         return
 
-    possible_fields = [
-        "reasoning_content",
-        "thinking",
-        "chain_of_thought",
-    ]
+    possible_fields = REASONING_FIELD_NAMES
     value: Any | None = None
```
njbrake, line 18: Note that the only change here is that now "think" is also a possible field name. It seemed reasonable to me to add it.

PR #578, opened 2025-10-28
feat(batch): Support for Native Batch API

njbrake:
## Description

Initial support for batch. This PR implements only openai batch, but sets up repo for more implementations. I added an experimental decorator to make it clear that the Batch API may need a few changes to the interface as we expand the implementations and usage.

The `any-llm` batch API requires you to pass a **path to a local JSONL file** containing your batch requests. The provider implementation automatically handles uploading and file management as needed.

Different providers handle batch processing differently:

- **OpenAI**: Requires uploading a file first, then creating a batch with the file ID
- **Anthropic** (future): Expects file content passed directly in the request
- **Other providers**: May have their own unique requirements

By accepting a local file path, `any-llm` abstracts these provider differences and handles the implementation details automatically.

## Checklist

### DIFF
--- a/src/any_llm/any_llm.py
+++ b/src/any_llm/any_llm.py
@@ -536,7 +536,6 @@ class AnyLLM(ABC):
 
         Args:
             input_file_path: Path to a local file containing batch requests in JSONL format.
-                The file will be automatically uploaded before creating the batch.
             endpoint: The endpoint to be used for all requests (e.g., '/v1/chat/completions')
             completion_window: The time frame within which the batch should be processed (default: '24h')
             metadata: Optional custom metadata for the batch
--- a/src/any_llm/api.py
+++ b/src/any_llm/api.py
@@ -461,7 +461,6 @@ def create_batch(
     Args:
         provider: Provider name to use for the request (e.g., 'openai', 'mistral')
         input_file_path: Path to a local file containing batch requests in JSONL format.
-            The file will be automatically uploaded before creating the batch.
         endpoint: The endpoint to be used for all requests (e.g., '/v1/chat/completions')
         completion_window: The time frame within which the batch should be processed (default: '24h')
         metadata: Optional custom metadata for the batch
@@ -502,7 +501,6 @@ async def acreate_batch(
     Args:
         provider: Provider name to use for the request (e.g., 'openai', 'mistral')
         input_file_path: Path to a local file containing batch requests in JSONL format.
-            The file will be automatically uploaded before creating the batch.
         endpoint: The endpoint to be used for all requests (e.g., '/v1/chat/completions')
         completion_window: The time frame within which the batch should be processed (default: '24h')
         metadata: Optional custom metadata for the batch


tests/integration/test_batch.py
```diff
@@ -0,0 +1,208 @@
...
+                "custom_id": "request-2",
+                "method": "POST",
+                "url": "/v1/chat/completions",
+                "body": {
+                    "model": model_id,
+                    "messages": [{"role": "user", "content": "What is the capital of France?"}],
+                    "max_tokens": 100,
+                },
+            },
+        ]
```
njbrake, line 52: Format follows spec here: https://platform.openai.com/docs/guides/batch We may need to make updates as we add more providers, but I tried to avoid making this PR tooo massive, since all the new apis will be marked with the experimental decorator in any_llm, i think that buy us some grace to make tweaks and further harden the api as we continue implementation (or hopefully, get some community contribution for the implementations 🤞 )

src/any_llm/providers/openai/base.py
```diff
@@ -187,3 +192,78 @@ async def _alist_models(self, **kwargs: Any) -> Sequence[Model]:
...
+
+        file_path = Path(input_file_path)
+        file_content = await asyncio.to_thread(file_path.read_bytes)
+
+        file_obj = BytesIO(file_content)
+        file_obj.name = file_path.name
+
+        uploaded_file = await self.client.files.create(file=file_obj, purpose="batch")
+
+        valid_endpoint = cast(
```
besaleli, line 220: should we have more explicit error handling/validation here?

src/any_llm/utils/decorators.py
```diff
@@ -0,0 +1,46 @@
...
+from functools import wraps
+from typing import Any, TypeVar
+
+F = TypeVar("F", bound=Callable[..., Any])
+
+BATCH_API_EXPERIMENTAL_MESSAGE = "The Batch API is experimental and may have breaking changes in future versions."
+DEFAULT_EXPERIMENTAL_MESSAGE = "This API is experimental and subject to breaking changes."
+
+
+def experimental(message: str = DEFAULT_EXPERIMENTAL_MESSAGE) -> Callable[[F], F]:
```
besaleli, line 14: slay

src/any_llm/providers/openai/base.py
njbrake: Eventually, absolutely. My hesitancy was to go too crazy on validation until I find another provider that uses the OpenAI api and supports batch. Once we go through the providers and gather the state of batch api across everybody then I think we can for sure improve this code and clean up all the casting (that I did to make mypy happy tbh)

src/any_llm/utils/decorators.py
njbrake: ⚔️ 🚀 💪

njbrake:
Integration tests; https://github.com/mozilla-ai/any-llm/actions/runs/18878613217

docs/api/batch.md
```diff
@@ -0,0 +1,27 @@
+# Batch
+
+!!! warning "Experimental API"
+    The Batch API is experimental and subject to breaking changes in future versions. Use with caution in production environments.
+
+The Batch API allows you to process multiple requests asynchronously at a lower cost.
+
+## File Path Interface
+
+The `any-llm` batch API requires you to pass a **path to a local JSONL file** containing your batch requests. The provider implementation automatically handles uploading and file management as needed.
```
daavoo, line 10: Do we like this API or is more like what openai uses and we follow? 

I feel like we are not providing the "any" value because each provider batch has an specific format (i.e. https://platform.openai.com/docs/guides/batch#1-prepare-your-batch-file VS https://ai.google.dev/gemini-api/docs/batch-api#input-file ).

I think it would not be that hard to create the file "on the fly" and have an API that doesn't deviate that much from the rest.

Like, take a list of `completion` args as input that we convert to tempfile:

```python
# Dumps to temp file and makes the create batch request
any_llm.batch_completion(
    [
        {"model_id": "foo", "messages": ...},
        {"model_id": "bar", "messages": ...},
        ...
    ]
)
```

And/or, if we want to stick with the file as input, we could still provide some helper to create the batch file from the current API:

```python
@to_batch("openai_batch.jsonl")
any_llm.completion(...)
...

any_llm.create_batch("openai_batch.jsonl")
```

njbrake: Totally: I think the `any` value will come as we expand to more providers and better generalize the api to be the best implementation for everybody. I guess I was hoping to do this in a step by step manner so that I don't make any individual PR too insanely big. 

I think we will need to type out some standardized definition for exactly what format the list of batch completion requests will need to be. 

Initially I went with passing a file path, because theoretically running a batch request would use a lot of requests, so then there would be the inefficiency of loading it into memory before our process would be responsible to write the string from memory into a file. I get that its probably not going to be a massive amount of memory in most cases but it did seem a bit inefficient.

RE: your idea about `we could still provide some helper to create the batch file from the current API:`, I don't really understand why you would want to use a decorator from a completion response to write to a batch.jsonl file and then pass that into the create_batch function. Maybe you can explain a bit more which use case you're thinking of? I don't follow.

daavoo: >  Maybe you can explain a bit more which use case you're thinking of? I don't follow.

If someone is using the completion API and wants to move to the batch API, they could add the decorator and then have a separate job or something that creates the batch request at the end of the day.

I don't know 🤷 trying to think what is the value we could provide comparted to people calling the openai client directly

daavoo: > unning a batch request would use a lot of requests, so then there would be the inefficiency of loading it into memory before our process would be responsible to write the string from memory into a file

I guess I see the usage of batch API not only as "having a lot of request" but potentially also as "I don't need immediate answer" for a dozen of requests and I can save some money by sending them in a batch waiting 1 day

njbrake:
Integration tests: https://github.com/mozilla-ai/any-llm/actions/runs/18911326735/job/53986774991

Batch test for openai ✅  David and I discussed his comments on slack and he is comfortable with the merge for this to get initial support out there. We'll expand and refine the API as we gather input from the community. Thanks!

PR #587, opened 2025-10-29
feat(gateway): add gateway provider

njbrake:
## Description

(untested) Eventually I think we would be able to add it to the integration test suite, once the gateway is a public image so that we can easily run it in a github action without needing to configure authentication etc.

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -199,6 +199,8 @@ filterwarnings = [
   "ignore:Pydantic V1 style `@validator` validators are deprecated:pydantic.warnings.PydanticDeprecatedSince20:fireworks\\.client\\.image_api",
   # Silence fireworks generated protos deprecation about BaseModelDetails.tunable only
   "ignore:BaseModelDetails\\.tunable is deprecated:DeprecationWarning:fireworks\\.control_plane\\.generated\\.protos\\.gateway.*",
+  # Silence model format deprecation warnings
+  "ignore:Model format 'provider/model' is deprecated:DeprecationWarning",
 ]
 
 [tool.coverage.run]
--- a/tests/unit/providers/test_deepseek_provider.py
+++ b/tests/unit/providers/test_deepseek_provider.py
@@ -5,7 +5,7 @@ from any_llm.providers.deepseek.utils import _preprocess_messages
 from any_llm.types.completion import CompletionParams
 
 
-class TestResponseFormat(BaseModel):
+class PersonResponseFormat(BaseModel):
     name: str
     age: int
 
@@ -17,7 +17,7 @@ async def test_preprocess_messages_with_pydantic_model() -> None:
     params = CompletionParams(
         model_id="deepseek-chat",
         messages=messages,
-        response_format=TestResponseFormat,
+        response_format=PersonResponseFormat,
     )
 
     processed_params = _preprocess_messages(params)
--- a/tests/unit/providers/test_gateway_provider.py
+++ b/tests/unit/providers/test_gateway_provider.py
@@ -1,0 +1,137 @@
+import os
+from unittest.mock import AsyncMock, MagicMock, patch
+
+import pytest
+
+from any_llm.providers.gateway.gateway import GATEWAY_HEADER_NAME, GatewayProvider
+
+
+def test_gateway_init_requires_api_base() -> None:
+    with pytest.raises(ValueError, match="api_base is required"):
+        GatewayProvider(api_key="test-key")
+
+
+@patch("any_llm.providers.openai.base.AsyncOpenAI")
+def test_gateway_init_with_api_key(mock_openai_class: MagicMock) -> None:
+    mock_client = AsyncMock()
+    mock_openai_class.return_value = mock_client
+
+    GatewayProvider(api_key="test-key", api_base="https://gateway.example.com")
+
+    mock_openai_class.assert_called_once()
+    call_kwargs = mock_openai_class.call_args[1]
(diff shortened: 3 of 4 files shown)


src/any_llm/providers/gateway/gateway.py
```diff
@@ -0,0 +1,26 @@
+from typing import Any
+
+from any_llm.providers.openai.base import BaseOpenAIProvider
+
+
+class GatewayProvider(BaseOpenAIProvider):
+    ENV_API_KEY_NAME = "None"
```
daavoo: Shouldn't we:

```suggestion
    ENV_API_KEY_NAME = "GATEWAY_MASTER_KEY"
```

src/any_llm/providers/gateway/gateway.py
```diff
@@ -0,0 +1,26 @@
...
+    SUPPORTS_COMPLETION = True
+    SUPPORTS_RESPONSES = True
+    SUPPORTS_COMPLETION_REASONING = True
+    SUPPORTS_COMPLETION_IMAGE = True
+    SUPPORTS_COMPLETION_PDF = True
+    SUPPORTS_EMBEDDING = True
+    SUPPORTS_LIST_MODELS = True
+    SUPPORTS_BATCH = True
+
+    def __init__(self, api_key: str | None = None, api_base: str | None = None, **kwargs: Any) -> None:
```
daavoo, line 26: It would be nice if we override prepare_completion args or something to include the:

```
extra_headers={"X-AnyLLM-Key": f"Bearer {self.api_key}"},
```

Assuming `self.api_key` holds `GATEWAY_MASTER_KEY` (per the comment above)

src/any_llm/providers/gateway/gateway.py
njbrake: My gut reaction is to say initially no, we should leave this blank, because you can use the gateway either with the master key or with the gateway created virtual key. There's also the question about whether we should add the header X-Anyllm-header as an arg before creating the client, but I decided not to because that header name is a configurable value within the gateway, so people may not be naming it that when they're using the gateway interface.

njbrake: Or maybe we very genetically name at Gateway API key

daavoo: > Or maybe we very genetically name at Gateway API key

Let's do this

daavoo: > that header name is a configurable value within the gateway

what is the use case for configuring it?

njbrake: Cool, I'll use GATEWAY_API_KEY, but I'll make all the keys optional so that people can still use the connection if they dangerously have no auth configured, how does that sound?

> what is the use case for configuring it?
Good point.... My thought was maybe they want to customize it to their own branding, but maybe this is a case of uneccessary added flexibility. I'm fine with removing that feature. Let me know what you think and I can add a ticket in any-llm-gateway to drop the feature

src/any_llm/providers/gateway/gateway.py
```diff
@@ -0,0 +1,43 @@
...
+        if not api_base:
+            msg = "For any-llm gateway, api_base is required"
+            raise ValueError(msg)
+        api_key = self._verify_and_set_api_key(api_key)
+        if api_key:
+            if not kwargs:
+                # if passed into client init, the parameters
+                kwargs = {"default_headers": {}}
```
njbrake, line 34: @daavoo  when passed into the AsyncOpenAI constructor it expects `default_headers`. If we wanted to inject the header right before the call then we would use `extra_headers`. I chose to implement it in default headers that way I didn't need to override each api call definition. Let me know if you think otherwise

pyproject.toml
```diff
@@ -19,7 +19,7 @@ dependencies = [
 [project.optional-dependencies]
 
 all = [
-  "any-llm-sdk[mistral,anthropic,huggingface,gemini,vertexai,cohere,cerebras,fireworks,groq,bedrock,azure,azureopenai,watsonx,together,sambanova,ollama,moonshot,nebius,xai,databricks,deepseek,inception,openai,openrouter,portkey,lmstudio,llama,voyage,perplexity,llamafile,llamacpp,sagemaker]"
+  "any-llm-sdk[mistral,anthropic,huggingface,gemini,vertexai,cohere,cerebras,fireworks,groq,bedrock,azure,azureopenai,watsonx,together,sambanova,ollama,moonshot,nebius,xai,databricks,deepseek,inception,openai,openrouter,portkey,lmstudio,llama,voyage,perplexity,llamafile,llamacpp,sagemaker, gateway]"
```
daavoo: Maybe I am just paranoid about the extra whitespace but just in case:

```suggestion
  "any-llm-sdk[mistral,anthropic,huggingface,gemini,vertexai,cohere,cerebras,fireworks,groq,bedrock,azure,azureopenai,watsonx,together,sambanova,ollama,moonshot,nebius,xai,databricks,deepseek,inception,openai,openrouter,portkey,lmstudio,llama,voyage,perplexity,llamafile,llamacpp,sagemaker,gateway]"
```

src/any_llm/providers/gateway/gateway.py
daavoo: default_headers sounds even better to me

njbrake:
> Not critical, but some simple unit tests would be nice to have

Good point. I'll add a few unit tests and then I'll merge it 👍

PR #598, opened 2025-11-04
feat: any-llm-gateway

njbrake:
## Description

The any-llm-gateway is a new feature of any-llm that allows you to host any-llm as a proxy, giving you access to budget and user management systems.

To check out the docs (which probably need some help), git checkout the branch, then

```
uv sync --all-extras
mkdocs serve
```
open up localhost:8000 and take a look

## Checklist

### DIFF
--- a/docs/gateway/overview.md
+++ b/docs/gateway/overview.md
@@ -11,17 +11,17 @@ Managing LLM costs and access at scale is challenging. Give users unrestricted a
 - **Cost Control**: Set budgets that automatically enforce or track spending limits
 - **Access Management**: Issue, revoke, and monitor API keys generated for user access without exposing provider credentials
 - **Complete Visibility**: Track every request with full token counts, costs, and metadata
-- **Production Ready**: Deploy with Docker and Postgres, Kubernetes-ready
+- **Production-Ready**: Deploy with Docker and Postgres, Kubernetes-ready
 
 ## How it works
 
 The gateway exposes an OpenAI-compatible Completions API that works with any supported provider. Your applications connect to the gateway instead of directly to LLM providers, and the gateway handles:
 
-1. **Authentication**: Validates requests using master keys or virtual API keys
-2. **Budget Enforcement**: Checks spending limits before forwarding requests
-3. **Provider Routing**: Routes requests to the appropriate LLM provider using the `provider:model` format (e.g., `openai:gpt-4o-mini`, `anthropic:claude-3-5-sonnet-20241022`)
-4. **Usage Tracking**: Logs all requests with token counts and costs
-5. **Streaming Support**: Handles streaming responses with automatic token tracking
+- **Authentication**: Validates requests using master keys or virtual API keys
+- **Budget Enforcement**: Checks spending limits before forwarding requests
+- **Provider Routing**: Routes requests to the appropriate LLM provider using the `provider:model` format (e.g., `openai:gpt-4o-mini`, `anthropic:claude-3-5-sonnet-20241022`)
+- **Usage Tracking**: Logs all requests with token counts and costs
+- **Streaming Support**: Handles streaming responses with automatic token tracking
 
 ## Key Features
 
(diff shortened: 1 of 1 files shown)


src/any_llm/gateway/__init__.py
```diff
@@ -0,0 +1,6 @@
+from importlib.metadata import PackageNotFoundError, version
+
+try:
+    __version__ = version("any-llm-sdk")
+except PackageNotFoundError:
+    __version__ = "0.0.0-dev"
```
njbrake: This means that the versioning of the gateway is tied to the any_llm package. They share the same pyproject.toml and this code imports that version.

docker/docker-entrypoint.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -e
+
+if [ "$AUTO_MIGRATE" = "true" ]; then
```
daavoo: Can we make this the default (if env var is not set)?

docs/gateway/api-reference.md
```diff
@@ -0,0 +1,42 @@
...
+To regenerate the OpenAPI spec after making changes to the API routes, run:
+```bash
+python scripts/generate_openapi.py
+```
+
+## Chat Completions
+
+All chat completion endpoints require the use of a client key: these are keys created by someone using the master key.
+
+Look for the routes in `src/any_llm/gateway/chat.py`
```
daavoo: We could look into something like https://www.neoteroi.dev/mkdocs-plugins/web/oad/ to autogenerate endpoint docs

src/any_llm/gateway/alembic/versions/28d153c22616_initial_migration_with_all_models.py
daavoo, line 1: This initial migration doesn't work on a fresh DB, the `api_keys` and `usage_logs` tables are not being created.

I am guessing that the root issue is `init_db` running `Base.metadata.create_all(bind=_engine)`, which should not be there if we meant to use alembic. 

At some point, `init_db` was manually run (creating all models defined at that point) and later the alembic revision was generated, skipping the creation of existing tables (`api_keys`, `usage_logs`)

We should:

- Default to always run alembic migrations.
- Drop `Base.metadata.create_all(bind=_engine)`.
- Fix this initial migration.

daavoo (changes requested):
alembic migrations and `init_db` need to be fixed.

See https://github.com/mozilla-ai/any-llm-gateway/commit/777c89a3ab5379c9bc67a2d3ce19b450770bde78#diff-55d1bf7a4ed5781996c8e8bfba516cad32ca140432ef29fa6d2019ca497db093

docs/gateway/api-reference.md
njbrake: Yep absolutely we should render the openapi spec better than pointing to a JSON. I'll create a Github Issue

njbrake: https://github.com/mozilla-ai/any-llm/issues/599

src/any_llm/gateway/alembic/versions/28d153c22616_initial_migration_with_all_models.py
njbrake: Ok, done in https://github.com/mozilla-ai/any-llm/pull/598/commits/5698e39fc1e022ce7cd902dfab132052d29b8b4b I think with this change, now that init_db runs the migration, we can completely delete the docker-entrypoint.sh since init_db is run on server startup. 

I updated so that auto_migrate is now a setting in the GatewayConfig and it defaults to True: https://github.com/mozilla-ai/any-llm/pull/598/commits/593c535e13e718b9204339ed1167109e0ca456ee

docs/gateway/quickstart.md
```diff
@@ -0,0 +1,159 @@
...
+> The following assumes that you have cloned this repository and `cd` into its root.
+
+### Option 1: Docker compose
+
+First, create a `config.yaml` file with your configuration, using config.example.yaml as a template.
+
+Then run the Docker containers:
+
+```bash
+docker-compose up -d --build
```
daavoo: This command doesn't work. See #600 

side note, I feel like I would drop the docker-compose from the repo

njbrake: I'd rather keep docker-compose atm so that it gives people an easy way to spin up postgres alongside the gateway

daavoo (approved):

PR #600, opened 2025-11-05
fix(docker): Add `VERSION` arg.

daavoo:
Because we use `setuptools_scm` in our any-llm pyproject.toml the `pip install --no-cache-dir .[gateway]` inside docker fails (`.git` dir is not available there).

Copying the `.git` inside docker feels suboptimal so instead I added the `VERSION` arg:

```bash
VERSION=$(git describe --tags --always | sed 's/-\([0-9]\+\)-g\([0-9a-f]\+\)/.dev\1+g\2/') \
docker-compose -f docker/docker-compose.yml up --build
```

```bash
docker build \
--build-arg VERSION=$(git describe --tags --always | sed 's/-\([0-9]\+\)-g\([0-9a-f]\+\)/.dev\1+g\2/') \
-f docker/Dockerfile .
```


docker/Dockerfile
```diff
@@ -7,8 +7,12 @@ COPY src/ ./src/
 COPY alembic/ ./alembic/
 COPY alembic.ini ./
 
+# Set version explicitly for setuptools-scm
+ARG VERSION=0.0.0
+ENV SETUPTOOLS_SCM_PRETEND_VERSION=${VERSION}
+
 RUN pip install --no-cache-dir --upgrade pip && \
-    pip install --no-cache-dir -e .[gateway]
+    pip install --no-cache-dir .[gateway]
```
daavoo, line 15: Also removed the `-e` because I don't see why we would want editable mode inside docker

njbrake:
Fixed in the gateway branch, with slightly different (but very similar) approach

PR #618, opened 2025-11-10
feat: Integrate any-llm-platform for key management and usage tracking

dpoulopoulos:
Integrate any-llm with the any-llm-platform.

Key changes include:

- Introduce a mechanism to securely fetch provider keys from the any-llm-platform using a challenge-response authentication method.
- Send completion usage data (e.g., token counts) to the platform.

## PR Type

🆕 New Feature

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,13 +13,19 @@ dependencies = [
   "openai>=1.99.3",
   "rich",
   "httpx",
-  "typing-extensions"
 ]
 
 [project.optional-dependencies]
 
 all = [
-  "any-llm-sdk[mistral,anthropic,huggingface,gemini,vertexai,cohere,cerebras,fireworks,groq,bedrock,azure,azureopenai,watsonx,together,sambanova,ollama,moonshot,nebius,xai,databricks,deepseek,inception,openai,openrouter,portkey,lmstudio,llama,voyage,perplexity,llamafile,llamacpp,sagemaker,gateway,zai,minimax]"
+  "any-llm-sdk[mistral,anthropic,huggingface,gemini,vertexai,cohere,cerebras,fireworks,groq,bedrock,azure,azureopenai,watsonx,together,sambanova,ollama,moonshot,nebius,xai,databricks,deepseek,inception,openai,openrouter,portkey,lmstudio,llama,voyage,perplexity,platform,llamafile,llamacpp,sagemaker,gateway,zai,minimax]"
+]
+
+platform = [
+  "typing-extensions",
+  "bcrypt>=5.0.0",
+  "cryptography>=46.0.3",
+  "pynacl>=1.6.0",
 ]
 
 perplexity = []
--- a/src/any_llm/any_llm.py
+++ b/src/any_llm/any_llm.py
@@ -11,7 +11,7 @@ from any_llm.constants import INSIDE_NOTEBOOK, LLMProvider
 from any_llm.exceptions import MissingApiKeyError, UnsupportedProviderError
 from any_llm.tools import prepare_tools
 from any_llm.types.completion import ChatCompletion, ChatCompletionMessage, CompletionParams
-from any_llm.types.provider import ProviderMetadata
+from any_llm.types.provider import PlatformKey, ProviderMetadata
 from any_llm.types.responses import Response, ResponseInputParam, ResponsesParams, ResponseStreamEvent
 from any_llm.utils.aio import async_iter_to_sync_iter, run_async_in_sync
 from any_llm.utils.decorators import BATCH_API_EXPERIMENTAL_MESSAGE, experimental
@@ -87,6 +87,8 @@ class AnyLLM(ABC):
     For example, in `gemini` provider, this could include `google.genai.types.Tool`.
     """
 
+    ANY_LLM_KEY: str = "ANY_LLM_KEY"
+
     def __init__(self, api_key: str | None = None, api_base: str | None = None, **kwargs: Any) -> None:
         self._verify_no_missing_packages()
         self._init_client(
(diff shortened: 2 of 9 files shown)


pyproject.toml
```diff
@@ -13,7 +13,10 @@ dependencies = [
   "openai>=1.99.3",
   "rich",
   "httpx",
-  "typing-extensions"
+  "typing-extensions",
+  "bcrypt>=5.0.0",
+  "cryptography>=46.0.3",
+  "pynacl>=1.6.0",
```
daavoo: Not a strong opinion, but these dependencies should probably go to a separate group in extras (i.e. `platform`)

src/any_llm/any_llm.py
```diff
@@ -147,6 +152,24 @@ def _create_provider(
...
+                AnyAPIKey(api_key=api_key)
+
+                # Set the ANY_LLM_KEY environment variable for downstream usage.
+                # We need this to check whether to post usage events to any-llm platform.
+                os.environ[cls.ANY_LLM_KEY] = api_key
+
+                return AnyLLMPlatformProvider.prepare(
+                    provider_class, any_llm_key=api_key, api_base=api_base, **kwargs)
+            except ValueError:
+                msg = ("The provided API key does not conform with the any-api format. "
```
daavoo: Not really sure about having this log

src/any_llm/any_llm.py
```diff
@@ -147,6 +152,24 @@ def _create_provider(
             raise ImportError(msg) from e
 
         provider_class: type[AnyLLM] = getattr(module, provider_class_name)
+
+        if api_key:
```
daavoo, line 156: I think at this point we should also check for `os.environ[cls.ANY_LLM_KEY]`, so user setting that env var also serves as activation of the platform.

```suggestion
        if api_key:
```

src/any_llm/any_llm.py
```diff
@@ -395,7 +418,26 @@ async def acompletion(
                 messages[i] = message.model_dump(exclude_none=True, exclude={"reasoning"})
         all_args["messages"] = messages
 
-        return await self._acompletion(CompletionParams(**all_args), **kwargs)
+        completion = await self._acompletion(CompletionParams(**all_args), **kwargs)
+
+        # TODO: Handle AsyncIterator[ChatCompletionChunk] case
+        if isinstance(completion, ChatCompletion):
```
daavoo: I think this logic should not be part of `AnyLLM` base class but rather part of the new "wrapper" provider we are introducing in the PR.

It should implement a `_completion` where it calls the original `provider_class._completion` and then apply the new logic

src/any_llm/providers/anyllmplatform/anyllmplatform.py
daavoo, line 1: Not a strong opinion, we could probably just name this module and class `platform` (similar to how we name `gateway` the `any-llm-gateway` logic)

src/any_llm/providers/anyllmplatform/anyllmplatform.py
```diff
@@ -0,0 +1,27 @@
...
+class AnyLLMPlatformProvider:
+    @staticmethod
+    def prepare(provider_class: type[AnyLLM], any_llm_key: str, api_base: str | None = None, **kwargs) -> AnyLLM:
+        provider_key = get_provider_key(
+            any_llm_key=any_llm_key,
+            provider=provider_class,
+        )
+
+        provider_kwargs = {**kwargs}
+        sig = inspect.signature(provider_class.__init__)
```
daavoo: Not sure why we are doing this inspect, afaik all provider classes follow the same signature

src/any_llm/providers/platform/utils.py
```diff
@@ -0,0 +1,261 @@
...
+    The client has to create and solve two challenges, one to prove ownership of the private key
+    that decrypts the encrypted provider key, without sharing it, and one to prove ownership of the project
+    it wants to post data to.
+
+    Args:
+        any_llm_key: The Any LLM platform key, tied to a specific project.
+        provider: The name of the LLM provider.
+        completion: The LLM response.
+    """
+    # Parse the `ANY_LLM_KEY`, load the private key and extract the public key from it.
```
daavoo, line 209: Do we need to run all these operations on each post_usage?

Can we do this one inside the `PlatformProvider` and consider the provider authenticated?

src/any_llm/providers/anyllmplatform/utils.py
```diff
@@ -0,0 +1,261 @@
...
+        "provider": provider,
+        "model": model,
+        "data": {
+            "input_tokens": str(input_tokens),
+            "output_tokens": str(output_tokens),
+        },
+        "id": event_id
+    }
+
+    async with httpx.AsyncClient(timeout=10.0) as client:
```
daavoo: minor: we could probably instantiate and create a single client inside the `PlatformProvider` instance to reuse across calls

src/any_llm/providers/platform/utils.py
```diff
@@ -0,0 +1,261 @@
...
+import requests
+
+from any_llm.logging import logger
+
+if TYPE_CHECKING:
+    from any_llm.any_llm import AnyLLM
+    from any_llm.types.completion import ChatCompletion
+
+
+ANY_LLM_PLATFORM_URL = "http://localhost:8000/api/v1"
```
agpituk: should this point to the real URL before we merge it? Ideally we should be able to override this with an environment variable, so we can point to different environments

src/any_llm/providers/platform/utils.py
dpoulopoulos: Yes, currently we should do this on each `post_usage` request, since the endpoint expects a (new) solved challenge in the headers. We may revisit this as we're building the "decryptor" SDK that will replace this bit of code in any-llm.

src/any_llm/types/provider.py
```diff
@@ -15,3 +19,16 @@ class ProviderMetadata(BaseModel):
     class_name: str
     list_models: bool
     batch_completion: bool
+
+
+class AnyAPIKey(BaseModel):
```
daavoo: for consistency:

```suggestion
class PlatformKey(BaseModel):
```

src/any_llm/providers/platform/utils.py
```diff
@@ -0,0 +1,264 @@
...
+
+
+def _extract_public_key(private_key) -> str:
+    """Extract public key as base64 from X25519 private key."""
+    public_key_bytes = bytes(private_key.public_key)
+    return base64.b64encode(public_key_bytes).decode("utf-8")
+
+
+def _create_challenge(public_key: str, any_api_url: str) -> dict:
+    """Step 1: Create an authentication challenge."""
```
HareeshBahuleyan:
```suggestion
    """Create an authentication challenge."""
```

PR #623, opened 2025-11-10
OpenAPI spec updates

angpt:
## Description
Adding more information around authentication in the openapi spec. 

## PR Type

<!-- Delete the types that don't apply --!>

📚 Documentation

## Checklist
- [X] I have added unit tests that prove my fix/feature works
- [X] I have read and followed the [contribution guidelines](https://github.com/mozilla-ai/any-llm/blob/main/CONTRIBUTING.md)```

### DIFF
--- a/.github/workflows/gateway-docker.yml
+++ b/.github/workflows/gateway-docker.yml
@@ -27,7 +27,13 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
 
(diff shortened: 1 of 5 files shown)


scripts/generate_openapi.py
```diff
@@ -13,10 +13,95 @@
 import sys
 from pathlib import Path
 
-from any_llm.gateway.config import GatewayConfig
+from any_llm.gateway.config import API_KEY_HEADER, GatewayConfig
 from any_llm.gateway.server import create_app
 
 
+def _add_security_schemes(spec: dict) -> dict:
```
daavoo: Instead of manually updating the openapi spec, I think we could use the built-in `fastapi.security` (which should then automatically generate appropriate scheme):

https://fastapi.tiangolo.com/reference/security/#fastapi.security.APIKeyHeader

https://fastapi.tiangolo.com/reference/security/#fastapi.security.HTTPBearer

PR #627, opened 2025-11-17
Resolve Azure OpenAI ResourceNotFoundError

sujith-vasudev:
✔️ Fixes included

Corrected how the AzureOpenAI client resolves the api version.

🧪 Testing

Verified successful calls against valid Azure OpenAI deployments.


njbrake:
Hi @sujith-vasudev, please create a Github Issue to describe the problem you are encountering so that we can come to an agreement on the correct plan of action, thanks 🙏

PR #649, opened 2025-12-02
Support reading the gateway key from the `Authorization` header

jtpio:
## Description

As mentioned in https://mozilla-ai.github.io/any-llm/gateway/quickstart/, the `X-AnyLLM-Key` header must be provided to specify the master key when using the any-llm gateway:

```
  -H "X-AnyLLM-Key: Bearer ${GATEWAY_MASTER_KEY}" \
```

However it may be impractical to provide with existing generic OpenAI providers.

So it would make sense to also support reading the key from the `Authorization` header. 

## PR Type

🆕 New Feature

## Checklist

### DIFF
--- a/tests/gateway/test_key_management.py
+++ b/tests/gateway/test_key_management.py
@@ -246,9 +246,7 @@ def test_inactive_api_key_rejected(
     assert response.status_code == 401
 
 
-def test_authorization_header_accepted(
-    client: TestClient, test_config: GatewayConfig
-) -> None:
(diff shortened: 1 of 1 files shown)


jtpio:
The last commit should have addressed the lint issues.

Marking as ready so folks could take a first look already.

daavoo (approved):
Thanks @jtpio ! Changes make sense to me.
I have tested manually and it works.

Will add a test actually using the OpenAI client to do a completions request in a followup

PR #657, opened 2025-12-04
fix(minimax): Skip chunk contents with no delta.

daavoo:
Apparently the minimax API returns a final chunk for the `finish_reason` where the `delta` is None, which was causing our code to break while trying to validate the Pydantic model.


daavoo:
https://github.com/mozilla-ai/any-llm/actions/runs/19930559403/job/57141258726

PR #673, opened 2025-12-18
fix(anthropic): Handle parallel tool calls.

daavoo:


daavoo:
Integration tests: https://github.com/mozilla-ai/any-llm/actions/runs/20333518049/job/58414324842

PR #717, opened 2026-01-06
fix: together API expects non empty tool_calls list, upgrade to together sdk v2

njbrake:
## Description

together api started rejecting provision of an empty list of tool_calls. So, if the tool calls are empty, we should delete the key. The key value pair is kept around if it's not empty.

## PR Type

- 🐛 Bug Fix

## Checklist
- [x] **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

- AI Model used: Opus 4.5
- AI Developer Tool used: Cursor to fix, Claude Code to debug and find the problem
- Any other info you'd like to share: Had to go through a few iterations to make it usable: Opus wanted the cleanup logic to be a separate static member function of the TogetherAI class which was way overkill for the simple removal of an empty list.  Additionally, Claude Code found the problem but along the way found 3 other things that it claimed were the problem but were irrelevant. Since I gave it the tools to verify the fix, it did eventually find the problem, but it failed to undo all the things along the way that it hypothesized were problems but in reality were irrelevant.

### DIFF
--- a/src/any_llm/providers/together/together.py
+++ b/src/any_llm/providers/together/together.py
@@ -132,28 +132,14 @@ class TogetherProvider(AnyLLM):
         async for chunk in response:
             yield self._convert_completion_chunk_response(chunk)
 
-    @staticmethod
-    def _clean_messages_for_api(messages: list[dict[str, Any]]) -> list[dict[str, Any]]:
-        """Clean messages before sending to Together API.
-
-        Together API rejects messages with empty tool_calls arrays,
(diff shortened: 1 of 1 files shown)


njbrake:
Integration tests: https://github.com/mozilla-ai/any-llm/actions/runs/20760766999

verified that test now passes: [gw0] [ 88%] PASSED tests/integration/test_agent_loop.py::test_agent_loop_sequential_tool_calls[together]

PR #718, opened 2026-01-07
feat(azureopenai): AzureOpenAI use AzureOpenAI instead of OpenAI

njbrake:
## Description

- Introduced support for custom API versions and defaulted to "2024-02-01".
- Added Azure AD token authentication capability.
- Updated tests to validate new features and ensure compatibility with existing functionality.
- Adjusted provider configuration in tests to reflect new model and endpoint settings.

## PR Type

- 🆕 New Feature
- 💅 Refactor

## Checklist
- [ ] **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

- AI Model used: Opus 4.5 🚀 
- Any other info you'd like to share: Started with plan mode, passing in the docs for the openai pypi package (https://pypi.org/project/openai/).

Prompt: 

```
> My AzureOpenAI provider class currently only simply extends the base openai class. However, I want to leverage the AzureOpenAI class that exists in the OpenAI library instead. See these docs:

  [Pasted text #1 +749 lines]

  Can you think hard and figure out what would be the best way to update my code to use the AsyncAzureOpenAI class?
```

That got me to a usable first pass. After it did the initial implementation then I added in my credentials, ran our integration test suite, and had claude fix a few lingering issues (bad default api version, no completion pdf support, no test handling that aligns azureopenai with how we handle vertex and bedrock, etc).

---

> [!NOTE]
> Introduces a proper Azure OpenAI client and updates CI/tests to include it.
> 
> - Implement `AzureopenaiProvider` using `AsyncAzureOpenAI`; map `api_base` to `azure_endpoint`, default `api_version` to `"preview"` (overridable via `OPENAI_API_VERSION` or `api_version`), support `azure_ad_token`, set `SUPPORTS_COMPLETION_PDF = False`, and require `AZURE_OPENAI_ENDPOINT` when not provided
> - CI: add `azureopenai` to expected providers and wire `AZURE_OPENAI_API_KEY` in integration workflow
> - Tests: add unit tests covering endpoint mapping, API-version (default/custom/env), AAD token, and extra kwargs; update `conftest.py` model maps and `provider_client_config` for `AZUREOPENAI`; adjust generic provider-loading tests to pass `api_base` for `azureopenai`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit cca8b4de1a528dcbf4b5c9639bb3803c7c7a8653. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/any_llm/providers/azureopenai/azureopenai.py
+++ b/src/any_llm/providers/azureopenai/azureopenai.py
@@ -17,12 +17,12 @@ class AzureopenaiProvider(BaseOpenAIProvider):
     SUPPORTS_LIST_MODELS = True
     SUPPORTS_COMPLETION_PDF = False
 
-    DEFAULT_API_VERSION = "2024-02-01"
+    DEFAULT_API_VERSION = "preview"
 
     client: AsyncAzureOpenAI
 
     def _verify_and_set_api_key(self, api_key: str | None = None) -> str | None:
-        return api_key or os.getenv("AZURE_OPENAI_API_KEY")
+        return api_key or os.getenv(self.ENV_API_KEY_NAME)
 
     def _init_client(
         self,
@@ -32,7 +32,6 @@ class AzureopenaiProvider(BaseOpenAIProvider):
     ) -> None:
         api_version = kwargs.pop("api_version", None) or os.getenv("OPENAI_API_VERSION", self.DEFAULT_API_VERSION)
         azure_ad_token = kwargs.pop("azure_ad_token", None) or os.getenv("AZURE_OPENAI_AD_TOKEN")
-        azure_ad_token_provider = kwargs.pop("azure_ad_token_provider", None)
 
         azure_endpoint = api_base or kwargs.pop("azure_endpoint", None) or os.getenv("AZURE_OPENAI_ENDPOINT")
         if not azure_endpoint:
@@ -43,6 +42,5 @@ class AzureopenaiProvider(BaseOpenAIProvider):
             azure_endpoint=azure_endpoint,
             api_version=api_version,
             azure_ad_token=azure_ad_token,
-            azure_ad_token_provider=azure_ad_token_provider,
             **kwargs,
         )
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -115,6 +115,7 @@ def embedding_provider_model_map() -> dict[LLMProvider, str]:
         LLMProvider.VOYAGE: "voyage-3.5-lite",
         LLMProvider.LLAMACPP: "N/A",
         LLMProvider.GATEWAY: "text-embedding-ada-002",
+        LLMProvider.AZUREOPENAI: "gpt-4.1-nano",  # Not an embedding model but it's the only one we have deployed in Azure OpenAI
         LLMProvider.OPENROUTER: "qwen/qwen3-embedding-8b",
     }
 
--- a/tests/unit/providers/test_azureopenai_provider.py
+++ b/tests/unit/providers/test_azureopenai_provider.py
@@ -54,7 +54,7 @@ async def test_azureopenai_api_version_default() -> None:
         call_args = mock_azure_client.call_args
         assert call_args is not None
         _, kwargs = call_args
-        assert kwargs["api_version"] == "2024-02-01"
+        assert kwargs["api_version"] == "preview"
 
 
 @pytest.mark.asyncio


njbrake:
Integration tests for azureopenai: https://github.com/mozilla-ai/any-llm/actions/runs/20783059244

src/any_llm/providers/azureopenai/azureopenai.py
```diff
@@ -11,9 +15,34 @@ class AzureopenaiProvider(BaseOpenAIProvider):
...
     SUPPORTS_RESPONSES = True
     SUPPORTS_LIST_MODELS = True
+    SUPPORTS_COMPLETION_PDF = False
+
+    DEFAULT_API_VERSION = "2024-02-01"
+
+    client: AsyncAzureOpenAI
+
+    def _verify_and_set_api_key(self, api_key: str | None = None) -> str | None:
+        return api_key or os.getenv("AZURE_OPENAI_API_KEY")
```
daavoo: Why are we overriding this?

src/any_llm/providers/azureopenai/azureopenai.py
```diff
@@ -11,9 +15,34 @@ class AzureopenaiProvider(BaseOpenAIProvider):
...
+        api_base: str | None = None,
+        **kwargs: Any,
+    ) -> None:
+        api_version = kwargs.pop("api_version", None) or os.getenv("OPENAI_API_VERSION", self.DEFAULT_API_VERSION)
+        azure_ad_token = kwargs.pop("azure_ad_token", None) or os.getenv("AZURE_OPENAI_AD_TOKEN")
+        azure_ad_token_provider = kwargs.pop("azure_ad_token_provider", None)
+
+        azure_endpoint = api_base or kwargs.pop("azure_endpoint", None) or os.getenv("AZURE_OPENAI_ENDPOINT")
+        if not azure_endpoint:
+            raise MissingApiKeyError(self.PROVIDER_NAME, "AZURE_OPENAI_ENDPOINT")
```
daavoo, line 38: nit: I think MissingApiKeyError might be a little misleading here

src/any_llm/providers/azureopenai/azureopenai.py
```diff
@@ -11,9 +15,34 @@ class AzureopenaiProvider(BaseOpenAIProvider):
...
+
+    def _init_client(
+        self,
+        api_key: str | None = None,
+        api_base: str | None = None,
+        **kwargs: Any,
+    ) -> None:
+        api_version = kwargs.pop("api_version", None) or os.getenv("OPENAI_API_VERSION", self.DEFAULT_API_VERSION)
+        azure_ad_token = kwargs.pop("azure_ad_token", None) or os.getenv("AZURE_OPENAI_AD_TOKEN")
+        azure_ad_token_provider = kwargs.pop("azure_ad_token_provider", None)
```
daavoo: This is not needed, unless I am missing something

src/any_llm/providers/azureopenai/azureopenai.py
```diff
@@ -11,9 +15,34 @@ class AzureopenaiProvider(BaseOpenAIProvider):
     PROVIDER_DOCUMENTATION_URL = "https://learn.microsoft.com/en-us/azure/ai-foundry/"
     SUPPORTS_RESPONSES = True
     SUPPORTS_LIST_MODELS = True
+    SUPPORTS_COMPLETION_PDF = False
+
+    DEFAULT_API_VERSION = "2024-02-01"
```
daavoo: Is there a reason for this default value?

njbrake: Claude recommended we default to an older stable version instead of defaulting to 'preview'. I'm also ok to switch back to using 'preview' if nothing is specified, though then it risks a client using our endpoint having it silently break if 'preview' makes some big change.

src/any_llm/providers/azureopenai/azureopenai.py
njbrake: Because with Azure you don't need the openai_api_key, you could also use the ad_token. Though this did make me realize that I should have accessed the env var via `self.ENV_API_KEY_NAME` instead of re-typing the string. I'll update that.

src/any_llm/providers/azureopenai/azureopenai.py
njbrake: Ah I see what you mean, we don't need it because it would make it into the AsyncAzureOpenAI anyways in the `**kwargs`

src/any_llm/providers/azureopenai/azureopenai.py
njbrake: Yeah it's not ideal, but it is what we also do with VertexAI so it matches the current design 🤷 

```
        if self.client._api_client.project is None:
            msg = "vertexai"
            raise MissingApiKeyError(msg, "GOOGLE_CLOUD_PROJECT")
        if self.client._api_client.location is None:
            msg = "vertexai"
            raise MissingApiKeyError(msg, "GOOGLE_CLOUD_LOCATION")
```

src/any_llm/providers/azureopenai/azureopenai.py
njbrake: I'll stick this at 'preview' to keep it consistent with current behavior, and we can evaluate changing it via a separate PR

src/any_llm/providers/azureopenai/azureopenai.py
daavoo: > , you could also use the ad_token.

I guess then we should explicitly check for that ad_token here and raise an error if None of the 2 alternatives are present

njbrake: Let's handle this later. TBH I'm a little worried about throwing exceptions in the case where there's another valid configuration we don't know about due to our lack of experience working with Azure. Once we get some more runtime with using azure we should be able to make some more informed design decisions, if that's ok.

PR #723, opened 2026-01-08
refactor: streamline version handling in pyproject.toml

njbrake:
## Description

If you try to run stuff like `uv run pytest`, uv gets angry because the uv.lock for any-llm-sdk doesn't have a version set, which gives you this very annoying error message everytime the dynamic git based versioning tag changes:
```
     error: Failed to parse `uv.lock`
       Caused by: TOML parse error at line 218, column 1
         |
     218 | [[package]]
         | ^^^^^^^^^^^
     missing field `version`
```

It's because of the dynamic versioning happening in the pyproject.toml. We tried to be fancy with the dynamic versioning setting, but this PR simplifies this to a more "industry standard" design, where the github action that creates the release swaps out the version at build time. 

- Removed `setuptools_scm` dependency and adjusted versioning logic in `pyproject.toml`.
- Updated the GitHub Actions workflow to set the package version from the release tag using a regex.
- Simplified version retrieval in `__init__.py` files across the package to directly use `importlib.metadata.version`.

## PR Type

- 🐛 Bug Fix
- 💅 Refactor

## Checklist
- [ ] **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

- AI Model used: Opus 4.5
- Any other info you'd like to share: I put it in plan mode and it gave me a few different options to resolve the issue. The solution I settled one was my favorite of the three options it gave me. I had to do some additional prompting after the initial solution in order to have it clean up the __init__ files and also add testing that the version change was done correctly (as well as adding comments into the code to explain the linkage between the github action regex and the pyproject.toml version spec)

---

> [!NOTE]
> Replaces dynamic SCM-based versioning with explicit version injection at build/release time and aligns packaging, Docker, and module version exports.
> 
> - Sets static `version = "0.0.0-dev"` in `pyproject.toml`; removes `setuptools_scm` config/dependency
> - `release.yaml`: injects version from release tag into `pyproject.toml`, builds, verifies wheel `Version` matches tag, then publishes to PyPI
> - `docker/Dockerfile`: substitutes `ARG VERSION` into `pyproject.toml` and verifies substitution; removes `SETUPTOOLS_SCM_PRETEND_VERSION`
> - `src/any_llm/__init__.py`: resolves `__version__` via `importlib.metadata.version` with fallback to `"0.0.0-dev"`; `gateway/__init__.py` re-exports from root
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f510be739859ccd6e69d16ea7e5f446cc8a1d6b7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -15,6 +15,7 @@
 <!-- e.g. "Fixes #123" -->
 
 ## Checklist
+<!-- If this checklist is deleted from the PR submission it will be immediately closed -->
 - [ ] I understand the code I am submitting.
 - [ ] I have added unit tests that prove my fix/feature works
 - [ ] I have run this code locally and verified it fixes the issue.
@@ -32,3 +33,5 @@
 - AI Model used:
 - AI Developer Tool used:
 - Any other info you'd like to share:
+
+When answering questions by the reviewer, please respond yourself, do not copy/paste the reviewer comments into an AI system and paste back its answer. We want to discuss with you, not your AI :) 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -32,6 +32,7 @@ This repo uses `uv` for local dev (Python 3.11+). For the full, up-to-date comma
 - Python indentation: 4 spaces; formatting/linting via `ruff` (line length 120) and `pre-commit`.
 - Type hints: required; `mypy` runs in strict mode for library code (see `pyproject.toml`).
 - Provider code lives under `src/any_llm/providers/<provider>/` (keep provider-specific behavior isolated there).
+- Please add code comments if you find them helpful to accomplish your objective. However, please remove any comments you added that describe obvious behavior before finishing your task.
 
 ## Testing Guidelines
 
--- a/src/any_llm/providers/platform/platform.py
+++ b/src/any_llm/providers/platform/platform.py
@@ -8,6 +8,7 @@ from any_llm_platform_client import AnyLLMPlatformClient
 from httpx import AsyncClient
 
 from any_llm.any_llm import AnyLLM
+from any_llm.constants import LLMProvider
 from any_llm.logging import logger
 from any_llm.types.completion import (
     ChatCompletion,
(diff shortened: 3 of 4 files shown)


daavoo:
@njbrake I can't reproduce this:

> If you try to run stuff like uv run pytest, uv gets angry because the uv.lock for any-llm-sdk doesn't have a version set, which gives you this very annoying error message everytime the dynamic git based versioning tag changes:

     error: Failed to parse `uv.lock`
       Caused by: TOML parse error at line 218, column 1
         |
     218 | [[package]]
         | ^^^^^^^^^^^
     missing field `version`

It works just fine for me.

How/Where are you running the uv run pytest? Which UV version do you use?

njbrake:
> @njbrake I can't reproduce this:
> 
> > If you try to run stuff like uv run pytest, uv gets angry because the uv.lock for any-llm-sdk doesn't have a version set, which gives you this very annoying error message everytime the dynamic git based versioning tag changes:
> 
> ```
>  error: Failed to parse `uv.lock`
>    Caused by: TOML parse error at line 218, column 1
>      |
>  218 | [[package]]
>      | ^^^^^^^^^^^
>  missing field `version`
> ```
> 
> It works just fine for me.
> 
> How/Where are you running the uv run pytest? Which UV version do you use?
@daavoo 
It's a really goofy corner case: because the build version only gets calculated at build time, if the package isn't built then the version is empty. For instance, this reproduction inside VsCode terminal, where a different virtual env is active even though it doesn't show as being active, results in the uv.lock for any-llm-sdk missing the version spec.

```main][/tmp/one_llm]$ uv run python -c 'print("hi")'
warning: `VIRTUAL_ENV=/Users/nbrake/scm/any-llm/.venv` does not match the project environment path `.venv` and will be ignored; use `--active` to target the active environment instead
Using CPython 3.12.8
Creating virtual environment at: .venv
      Built any-llm-sdk @ file:///private/tmp/one_llm
Installed 130 packages in 502ms
hi
[main][/tmp/one_llm]$ grep -A 10 'any-llm-sdk' uv.lock
name = "any-llm-sdk"
##### THERE SHOULD BE A VERSION SPEC HERE!!!!!! #######
source = { editable = "." }
dependencies = [
    { name = "httpx" },
    { name = "openai" },
    { name = "pydantic" },
    { name = "rich" },
]

[package.optional-dependencies]
```

Screenshot 2026-01-08 at 8 39 08 AM

I'm not entirely sure all of the different scenarios where it might happen, but it does happen to me on a semi-weekly basis

daavoo:
> It's a really goofy corner case: because the build version only gets calculated at build time, if the package isn't built then the version is empty.

That sounds to me like some bug with UV (should be building any-llm as part of uv run) and/or something clunky going on on the venv setup.
Considering setuptools_scm  "fancy" and/or incompatible with uv feels like a stretch (I assume some LLM made these statements)

I am not 100% on favor of changing the release system mainly because the alternative doesn't look simpler, less clunkier (it's using `sed` in 4 different places, etc) nor is more standard than setuptools_scm.

njbrake:
> I am not 100% on favor of changing the release system mainly because the alternative doesn't look simpler, less clunkier (it's using `sed` in 4 different places, etc) nor is more standard than setuptools_scm.

Lol the "fancy" statement was actually my own, by "fancy" I meant "a higher level of abstraction". The sed method is more code, but is a little clearer imo about what's happening. My thinking that sed would be simpler is also because all the python libs we developed in my previous experience we used the `sed` replacement approach and we never had any issues 🤷 .


> That sounds to me like some bug with UV (should be building any-llm as part of uv run) and/or something clunky going on on the venv setup. Considering setuptools_scm "fancy" and/or incompatible with uv feels like a stretch (I assume some LLM made these statements)

You're right: this is likely a bug or corner case within uv or cursor or vscode (or something that I am messing up in my usage of the tools). In my thinking, the most direct way to avoid the corner case issues is to use the approach that I'd used in the past and had been working well for me with no issues. I think comically enough the whole "dynamic" build versioning was something that Cursor (Clause 3.5 Sonnet at the time) had suggested originally.  

How strong are your feelings about the change?

njbrake:
Doing some research about what other people do:

CrewAI does dynamic: https://github.com/crewAIInc/crewAI/blob/1.8.0/lib/crewai/pyproject.toml 
OpenAI pins it: https://github.com/openai/openai-agents-python/blob/main/pyproject.toml#L3 
Google ADK pins it in version.py: https://github.com/google/adk-python/blob/v1.21.0/src/google/adk/version.py

daavoo:
> How strong are your feelings about the change?

Not strong at all. As long as a release process works, I am fine with it.

My point was that changing the release process as a way of bypassing the downstream bug doesn't feel quite right, but if you consider this alternative release process also better, ship it

PR #724, opened 2026-01-08
fix: exclude stream_options parameter from non-OpenAI providers

### LINKED ISSUE #720: [BUG]: Using `mistral` through PlatformProvider fails with unexpected keyword argument 'stream_options'
**`any-llm` version**: 1.6.0
**Python Version**: 3.11.11

### Describe the bug
Error message:
```
  File "/Users/hareeshbahuleyan/Desktop/mzai_github/any-llm/src/any_llm/providers/platform/platform.py", line 132, in _stream_with_usage_tracking
    async for chunk in stream:
  File "/Users/hareeshbahuleyan/Desktop/mzai_github/any-llm/src/any_llm/providers/mistral/mistral.py", line 116, in _stream_completion_async
    mistral_stream = await self.client.chat.stream_async(model=model, messages=messages, **kwargs)  # type: ignore[arg-type]
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: Chat.stream_async() got an unexpected keyword argument 'stream_options'
```
I believe that `stream_options` is not a parameter supported by mistral? Possibly some other providers too?

### Example w/ Code
Run Platform Provider with streaming:
```
import dotenv
from any_llm import AnyLLM
dotenv.load_dotenv()

llm = AnyLLM.create("mistral")

response = llm.completion(
    model="mistral-medium-latest",
    messages=[{"role": "user", "content": "Fun things to do on New Year's Day"}],
    stream=True,
)

# Accumulate the streamed content
full_content = ""
for chunk in response:
    if hasattr(chunk, 'choices') and chunk.choices:
        content = chunk.choices[0].delta.content
        if content:
            full_content += content
            print(content, end="")
```

tbille:
## Description

Fixes `TypeError` when using non-OpenAI providers through PlatformProvider with streaming enabled.

PlatformProvider was automatically setting `stream_options={"include_usage": True}` for all providers to enable usage tracking in streaming responses. However, this OpenAI-specific parameter is not supported by several providers (Anthropic, Cerebras, Cohere, Gemini, Mistral, Ollama, Together), causing runtime errors.

This PR adds intelligent `stream_options` handling that:
- **Auto-enables** `stream_options` for supported providers (OpenAI, HuggingFace, etc.)
- **Temporarily removes** `stream_options` for unsupported providers during API calls
- **Restores** original values after provider calls to prevent mutation
- **Maintains** explicit errors for direct provider usage (no silent parameter dropping)

### Technical Implementation

The solution centralizes compatibility logic in the PlatformProvider wrapper:

```python
PROVIDERS_WITHOUT_STREAM_OPTIONS = {
    "anthropic", "cerebras", "cohere", "gemini",
    "mistral", "ollama", "together"
}
```

For unsupported providers, `stream_options` is temporarily set to `None` before the API call and restored afterward, preventing parameter mutation while fixing the compatibility issue.

### Provider Verification

Online documentation research confirmed:
- **Unsupported:** Anthropic, Cerebras, Cohere, Gemini, Mistral, Ollama, Together
- **Supported:** OpenAI, HuggingFace (via [OpenAI-compatible API](https://huggingface.co/docs/huggingface_hub/en/package_reference/inference_client))

## PR Type

- 🐛 Bug Fix
- ✨ Enhancement

## Relevant issues

Fixes #720

## Test Coverage

Added **16 comprehensive unit tests**:
- Auto-enable for supported providers (1 test)
- Removal for unsupported providers (7 parameterized tests)
- User-specified preservation (1 test)
- Value restoration after calls (7 parameterized tests)

Created `mock_streaming_chunks` fixture for test code reuse.

**Total:** 41 unit tests (25 existing + 16 new) - ✅ All passing

```bash
# Run all platform provider tests
uv run pytest tests/unit/providers/test_platform_provider.py -v

# Run only stream_options tests
uv run pytest tests/unit/providers/test_platform_provider.py -k "stream_options" -v
```

## Checklist
- [ ] **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

- **AI Model used:** Claude Sonnet 4.5 (1M context)
- **AI Developer Tool used:** Claude Code CLI
- **Details:** AI assisted with implementation design, comprehensive test writing, online provider documentation research to verify `stream_options` support, test refactoring with fixtures, and identifying all 7 affected providers

### DIFF
--- a/tests/unit/providers/test_platform_provider.py
+++ b/tests/unit/providers/test_platform_provider.py
@@ -1,3 +1,4 @@
+from collections.abc import AsyncIterator
 from datetime import datetime
 from unittest.mock import AsyncMock, Mock, patch
 from uuid import UUID
@@ -7,6 +8,7 @@ import pytest
 from any_llm_platform_client import DecryptedProviderKey
 from pydantic import ValidationError
 
+from any_llm.constants import LLMProvider
 from any_llm.exceptions import MissingApiKeyError
 from any_llm.providers.openai import OpenaiProvider
 from any_llm.providers.platform import PlatformProvider
@@ -88,6 +90,44 @@ def mock_platform_client() -> Mock:
     mock_client.aget_solved_challenge = AsyncMock(return_value=UUID("550e8400-e29b-41d4-a716-446655440000"))
     mock_client.get_public_key = Mock(return_value="mock-public-key")
     return mock_client
+
+
+@pytest.fixture
+def mock_streaming_chunks() -> list[ChatCompletionChunk]:
+    """Fixture for mock streaming chunks with usage data."""
+    return [
+        ChatCompletionChunk(
+            id="chatcmpl-123",
+            model="test-model",
+            created=1234567890,
+            object="chat.completion.chunk",
+            choices=[
+                ChunkChoice(
+                    index=0,
+                    delta=ChoiceDelta(role="assistant", content="Hello"),
+                    finish_reason=None,
+                )
+            ],
+        ),
+        ChatCompletionChunk(
+            id="chatcmpl-123",
+            model="test-model",
+            created=1234567890,
+            object="chat.completion.chunk",
+            choices=[
+                ChunkChoice(
+                    index=0,
+                    delta=ChoiceDelta(),
+                    finish_reason="stop",
+                )
+            ],
+            usage=CompletionUsage(
+                prompt_tokens=10,
+                completion_tokens=5,
+                total_tokens=15,
+            ),
+        ),
+    ]
 
 
 def test_platform_key_valid_format() -> None:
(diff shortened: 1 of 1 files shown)


src/any_llm/providers/cerebras/cerebras.py
```diff
@@ -51,8 +51,10 @@ class CerebrasProvider(AnyLLM):
 
     @staticmethod
     def _convert_completion_params(params: CompletionParams, **kwargs: Any) -> dict[str, Any]:
-        # Cerebras does not support providing reasoning effort
-        converted_params = params.model_dump(exclude_none=True, exclude={"model_id", "messages", "stream"})
+        # Cerebras does not support providing reasoning effort or stream_options
+        converted_params = params.model_dump(
+            exclude_none=True, exclude={"model_id", "messages", "stream", "stream_options"}
```
HareeshBahuleyan: Q: Do we still get usage data by default (when streaming) for those providers who do not support this param?

tbille: Yes, the stream_options parameter is not required for the posting to work.

tbille:
@njbrake I think it makes sense. I reverted the PR and the logic to be in the platform provider. I let claude make some research of the providers that doesn't support the streaming option for now and captured them in a hardcoded list (in the platform provider) + added a bunch of tests so we don't revert this by mistake :) 

Let me know what you think of this

src/any_llm/providers/platform/platform.py
```diff
@@ -94,11 +94,40 @@ async def _acompletion(
...
+        # 2. We isolate the workaround to where it's needed (platform wrapper)
+        # 3. We don't silently ignore user-specified parameters
+        providers_without_stream_options = {
+            "anthropic",
+            "cerebras",
+            "cohere",
+            "gemini",
+            "mistral",
+            "ollama",
+            "together",
```
njbrake: Please use the `constants.LLMProvider` enum values, not strings

src/any_llm/providers/platform/platform.py
```diff
@@ -94,11 +94,40 @@ async def _acompletion(
...
-        if params.stream and params.stream_options is None:
-            params.stream_options = {"include_usage": True}
-
-        completion = await self.provider._acompletion(params=params, **kwargs)
+        # List of providers that don't support stream_options
+        # Instead of modifying individual providers to silently drop this parameter,
+        # we handle it here in the platform wrapper. This way:
+        # 1. Users calling providers directly get clear errors if unsupported
+        # 2. We isolate the workaround to where it's needed (platform wrapper)
+        # 3. We don't silently ignore user-specified parameters
```
njbrake:
```suggestion
        # List of providers that don't support stream_options and automatically return token usage. This list may need to be updated if a provider updates its usage of stream options
```

src/any_llm/providers/platform/platform.py
```diff
@@ -94,11 +94,40 @@ async def _acompletion(
...
+            "mistral",
+            "ollama",
+            "together",
+        }
+
+        # Enable usage data in streaming responses for providers that support it
+        if params.stream:
+            provider_name = self.provider.PROVIDER_NAME
+
+            if provider_name in providers_without_stream_options:
```
njbrake:
```suggestion
            if self.provider.PROVIDER_NAME in providers_without_stream_options:
```

src/any_llm/providers/platform/platform.py
```diff
@@ -94,11 +94,40 @@ async def _acompletion(
...
+            "anthropic",
+            "cerebras",
+            "cohere",
+            "gemini",
+            "mistral",
+            "ollama",
+            "together",
+        }
+
+        # Enable usage data in streaming responses for providers that support it
```
njbrake:
```suggestion
```
Preference thing (feel free to not accept this suggestion), I tend to not care about comments that explain simple operations of the code. Idk if this was AI generated but I found Claude likes to comment every little piece of logic, I have a custom extra rule in Cursor/CC that says "after you write the code to solve the problem, go back and delete all the comments that are verbose and explaining obvious functionality". 

Some misc ML background is that Claude writing those comments into the code helps it generate the code better (due to the auto-regressive nature of LLM generation) so it makes sense to allow it to initially generate the comment. But imo after the code is written to implement the comment, the comment is no longer necessary.

src/any_llm/providers/platform/platform.py
```diff
@@ -94,11 +94,40 @@ async def _acompletion(
...
+        if params.stream:
+            provider_name = self.provider.PROVIDER_NAME
+
+            if provider_name in providers_without_stream_options:
+                # Remove stream_options for providers that don't support it
+                original_stream_options = params.stream_options
+                params.stream_options = None
+                completion = await self.provider._acompletion(params=params, **kwargs)
+                # Restore original value after the call
+                params.stream_options = original_stream_options
```
njbrake: Why do we need to restore the value?

src/any_llm/providers/platform/platform.py
```diff
@@ -94,11 +94,40 @@ async def _acompletion(
...
+
+        # Enable usage data in streaming responses for providers that support it
+        if params.stream:
+            provider_name = self.provider.PROVIDER_NAME
+
+            if provider_name in providers_without_stream_options:
+                # Remove stream_options for providers that don't support it
+                original_stream_options = params.stream_options
+                params.stream_options = None
+                completion = await self.provider._acompletion(params=params, **kwargs)
```
njbrake:
```suggestion
                if params.stream_options is not None:
                    logger.warning("You have set stream_options to True but this provider does not support the setting, dropping the setting")
                params.stream_options = None
                completion = await self.provider._acompletion(params=params, **kwargs)
```
Need a warning. The syntax isn't quite right but this is the general idea

tests/unit/providers/test_platform_provider.py
```diff
@@ -90,6 +91,44 @@ def mock_platform_client() -> Mock:
...
+                    finish_reason="stop",
+                )
+            ],
+            usage=CompletionUsage(
+                prompt_tokens=10,
+                completion_tokens=5,
+                total_tokens=15,
+            ),
+        ),
+    ]
```
njbrake, line 130: Does this fixture exist somewhere else? Seems like a common mock that we might already have elsewhere.

tests/unit/providers/test_platform_provider.py
```diff
@@ -872,3 +911,239 @@ async def test_non_streaming_includes_total_duration(
     call_args = mock_post_usage.call_args
     assert call_args.kwargs["total_duration_ms"] is not None
     assert call_args.kwargs["total_duration_ms"] > 0
+
+
+# Tests for stream_options handling
+
+
```
njbrake:
```suggestion
```

tests/unit/providers/test_platform_provider.py
```diff
@@ -872,3 +911,239 @@ async def test_non_streaming_includes_total_duration(
...
+@patch("any_llm_platform_client.AnyLLMPlatformClient.get_decrypted_provider_key")
+@patch("any_llm.providers.platform.platform.post_completion_usage_event")
+async def test_stream_options_auto_enabled_for_supported_provider(
+    mock_post_usage: AsyncMock,
+    mock_get_decrypted_provider_key: Mock,
+    any_llm_key: str,
+    mock_decrypted_provider_key: DecryptedProviderKey,
+    mock_streaming_chunks: list[ChatCompletionChunk],
+) -> None:
+    """Test that stream_options is auto-enabled for providers that support it (e.g., OpenAI)."""
```
njbrake: This should probably iterate over all the supported providers I think?

tests/unit/providers/test_platform_provider.py
```diff
@@ -872,3 +911,239 @@ async def test_non_streaming_includes_total_duration(
...
+    [
+        "anthropic",
+        "cerebras",
+        "cohere",
+        "gemini",
+        "mistral",
+        "ollama",
+        "together",
+    ],
+)
```
njbrake: Hmm. These two tests (this one and the other one that was added) can probably be combined into a single one. If you look at the conftest.py you'll see the fixture we have that iterates over all the providers. Then, you can have this list in order to decide whether the stream_options should be dropped vs not dropped.

tests/unit/providers/test_platform_provider.py
```diff
@@ -872,3 +911,239 @@ async def test_non_streaming_includes_total_duration(
...
+
+    # Verify that stream_options was None when passed to the provider (i.e., it was removed)
+    assert captured_stream_options is None
+
+
+@pytest.mark.asyncio
+@patch("any_llm_platform_client.AnyLLMPlatformClient.get_decrypted_provider_key")
+@patch("any_llm.providers.platform.platform.post_completion_usage_event")
+async def test_stream_options_preserved_when_user_specifies_it(
+    mock_post_usage: AsyncMock,
```
njbrake, line 1014: Hmm ok I think this answers my earlier question about why we restore the stream_options key after we dropped it. I think as long as we log a clear warning that we are ignoring the value it is ok.

I'm also thinking that maybe we should  make a copy of params in the platform instead of modifying params in place. Like what happens if multiple python functions are running in async and sharing a ref to that object? If we drop stream_options from params even for just a few cycles, there's a race condition where the other threads using the params object may accidentally use the value without stream_options set, which would be unintended.

tests/unit/providers/test_platform_provider.py
```diff
@@ -872,3 +911,239 @@ async def test_non_streaming_includes_total_duration(
...
+        "anthropic",
+        "cerebras",
+        "cohere",
+        "gemini",
+        "mistral",
+        "ollama",
+        "together",
+    ],
+)
+async def test_stream_options_restored_after_unsupported_provider_call(
```
njbrake: Could this test be combined with the test above it?

tests/unit/providers/test_platform_provider.py
tbille: Great catch on the race condition! I've addressed both points:

1. ✅ **Warning added**: Now logging a clear warning when user-specified `stream_options` must be ignored:
   ```python
   if params.stream_options is not None:
       logger.warning(
           f"stream_options was set but {self.provider.PROVIDER_NAME} does not support it. "
           "The parameter will be ignored for this request."
       )
   ```

2. ✅ **Using model_copy()**: Changed from mutating params in place to creating copies to prevent race conditions:
   ```python
   params_copy = params.model_copy()
   params_copy.stream_options = None
   completion = await self.provider._acompletion(params=params_copy, **kwargs)
   ```

This ensures thread safety - the original `params` object is never mutated, so concurrent async operations won't see unexpected state changes.

tbille: @njbrake just tested this comment reply with claude 😅

src/any_llm/providers/platform/platform.py
tbille: Updated to make a copy of the parameters 🫡

tests/unit/providers/test_platform_provider.py
tbille: After checking, it's it doens't

src/any_llm/providers/platform/platform.py
```diff
@@ -94,11 +95,37 @@ async def _acompletion(
...
+        if params.stream:
+            if self.provider.PROVIDER_NAME in providers_without_stream_options:
+                if params.stream_options is not None:
+                    logger.warning(
+                        f"stream_options was set but {self.provider.PROVIDER_NAME} does not support it. "
+                        "The parameter will be ignored for this request."
+                    )
+                params_copy = params.model_copy()
+                params_copy.stream_options = None
+                completion = await self.provider._acompletion(params=params_copy, **kwargs)
```
njbrake, line 119: I forget if this would work or not, and it's fine to ignore this, but if it works this might be a cleaner approach idk
```suggestion
                params_copy = params.model_copy()
                params_copy.stream_options = None
                completion = await self.provider._acompletion(params=params.model_dump(exclude=['stream_options'], **kwargs)
```

tests/unit/providers/test_platform_provider.py
```diff
@@ -872,3 +912,152 @@ async def test_non_streaming_includes_total_duration(
...
+        # Unsupported providers
+        LLMProvider.ANTHROPIC,
+        LLMProvider.CEREBRAS,
+        LLMProvider.COHERE,
+        LLMProvider.GEMINI,
+        LLMProvider.MISTRAL,
+        LLMProvider.OLLAMA,
+        LLMProvider.TOGETHER,
+    ],
+)
```
njbrake, line 937: Only slight concern is that this is neither a comprehensive list of our supported providers, and has no handling for when new providers are added.

tbille: Maybe we could have some checklist when adding a new provider?

src/any_llm/providers/platform/platform.py
tbille: Trying this out

tbille: No, it has a syntax error and model_dump() returns a dict but _acompletion() expects a CompletionParams object. The current code is correct.

PR #735, opened 2026-01-15
perf: queue usage events to platform to be sent in batches

khaledosman:
## Description

This PR queues usage events to be sent in batches in a single network request instead of making one http request for each usage event after every turn.. it will make a bulk api request to send queued events once the batch size (currently set to 50) or a time duration (currently set to 5 seconds) have elapsed instead.

### Why?
This should significantly improve performance, scalability and cut down infrastructure costs due to less network requests hitting the platform, less delay between every llm turn waiting for the network request and less network overhead with TCP connections, TLS handshakes and going through the network, as well as less server and db load and costs overall due to less (spikey) requests per client

## PR Type

- 🆕 New Feature
- 💅 Refactor

## Checklist

- [] I have run this code locally and verified it fixes the issue.
- [x] **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

- AI Model used: Claude Sonnet 4.5
- AI Developer Tool used: Copilot + VSCode

### DIFF
--- a/src/any_llm/providers/platform/utils.py
+++ b/src/any_llm/providers/platform/utils.py
@@ -9,8 +9,9 @@ from any_llm_platform_client import (
     AnyLLMPlatformClient,  # noqa: TC002
 )
 
-from .batch_queue import UsageEventBatchQueue
 from any_llm import __version__
+
+from .batch_queue import UsageEventBatchQueue
 
 if TYPE_CHECKING:
     from any_llm.types.completion import ChatCompletion


src/any_llm/providers/platform/utils.py
```diff
@@ -9,14 +9,88 @@
...
+        time_to_last_token_ms: Time to last token in milliseconds (streaming only).
+        total_duration_ms: Total request duration in milliseconds.
+        tokens_per_second: Average token generation throughput.
+        chunks_received: Number of chunks received (streaming only).
+        avg_chunk_size: Average tokens per chunk (streaming only).
+        inter_chunk_latency_variance_ms: Inter-chunk latency variance (streaming only).
+        batch_size: Maximum number of events per batch (default: 50).
+        flush_interval: Maximum seconds to wait before flushing (default: 5.0).
+    """
+    batch_queue = get_global_batch_queue(
```
daavoo: I am not fully understanding why we go with the global singleton approach in general (I would prefer to make the queue part of the platform provider instance) specially because here we are "binding" the global queue to a specific platform_client which comes from the platform provider instance🤸  ?¿

docs/usage-event-batching.md
daavoo, line 1: We are not really documenting platform behavior at all in the docs here (cc @tbille probably good to add a link to any-llm.ai docs or something)

src/any_llm/providers/platform/batch_queue.py
```diff
@@ -0,0 +1,354 @@
...
+
+from any_llm.logging import logger
+
+if TYPE_CHECKING:
+    from collections.abc import Sequence
+
+    from any_llm.types.completion import ChatCompletion
+
+
+class UsageEventBatch:
```
daavoo: I would put this is any_llm_platform_client (but I am not sure the class is really worth it)

Issue #622, opened 2025-11-10
[BUG] Reported token usage is wrong when doing Anthropic prompt caching


tsplude:
**`any-llm` version:** local version of 1.2.1.dev2+gfb98903eb
**Python Version**: Python 3.12.11

### Describe the bug
Token usage field values in the response from `any_llm.acompletion(...)` are wrong when doing Anthropic prompt caching via Anthropic provider directly or through OpenRouter provider.

### Anthropic Example

Anthropic example:
```python
async def get_any_llm_streamed_response():
    data = load_messages_and_tools()
    stream = await acompletion(
        model="claude-haiku-4-5",
        provider="anthropic",
        messages=data["messages"], # include cache_control breakpoint(s) as per Anthropic docs
        tools=data["tools"], # must inline $defs / $ref
        stream=True,
    )

    async for chunk in stream:
        print(chunk)

asyncio.run(get_any_llm_streamed_response())
```

Add raw response logging to AnthropicProvider._stream_completion_async:
```python
    async def _stream_completion_async(self, **kwargs: Any) -> AsyncIterator[ChatCompletionChunk]:
        """Handle streaming completion - extracted to avoid generator issues."""
        logger.debug("=== Anthropic Streaming Request ===")
        logger.debug(f"Stream kwargs: {kwargs}")

        async with self.client.messages.stream(
            **kwargs,
        ) as anthropic_stream:
            async for event in anthropic_stream:
                logger.debug("=== Raw Anthropic Stream Event ===")
                logger.debug(f"Event type: {type(event)}")
                logger.debug(f"Event: {event.model_dump() if hasattr(event, 'model_dump') else event}")
                yield self._convert_completion_chunk_response(event, model_id=kwargs.get("model", "unknown"))
```

<details>
<summary>Cache WRITE response</summary>

```bash
                    DEBUG    === Raw Anthropic Stream Event ===                                                                                                 anthropic.py:105
                    DEBUG    Event type: <class 'anthropic.lib.streaming._types.MessageStopEvent'>                                                              anthropic.py:106
                    DEBUG    Event: {'type': 'message_stop', 'message': {'id': '<removed>', 'content': [{'citations': None, 'text': "Hey     anthropic.py:107
                             Vader! 👋 Great to see you. I'm here to help you get the insights you need from your data.\n\nHere's what I can do for you:\n\n-
                             **Search existing dashboards and reports** to find answers that are already built\n- **Create custom queries and visualizations**
[...]

PR #750, opened 2026-01-20
fix: add anthropic prompt caching

njbrake:
## Description

Our Anthropic provider had the wrong calculation for prompt token usage. This PR fixes that bug and also adds a new field to the usage type to include that information. 

## PR Type

- 🆕 New Feature
- 🐛 Bug Fix
- 💅 Refactor
## Checklist

- [ ] I have read and followed the [contribution guidelines](https://github.com/mozilla-ai/any-llm/blob/main/CONTRIBUTING.md)
- [ ] **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

- AI Model used: Opus 4.5
- Any other info you'd like to share: It kept on trying to use getattr and setattr which I hate because then what's the point of typing lol. So this PR also suggests an addition to the AGENTS.md to tell the agent to stop trying to do that. The unit tests feel verbose to me but I suppose they're all in pursuit of codecov happiness so I left them.

### DIFF
--- a/src/any_llm/types/completion.py
+++ b/src/any_llm/types/completion.py
@@ -19,6 +19,7 @@ from openai.types.chat.chat_completion_message_function_tool_call import (
 )
 from openai.types.chat.chat_completion_message_function_tool_call import Function as OpenAIFunction
 from openai.types.completion_usage import CompletionUsage as OpenAICompletionUsage
+from openai.types.completion_usage import PromptTokensDetails as OpenAIPromptTokensDetails
 from openai.types.create_embedding_response import Usage as OpenAIUsage
 from openai.types.embedding import Embedding as OpenAIEmbedding
 from pydantic import BaseModel, ConfigDict, field_validator
@@ -73,6 +74,7 @@ class ChatCompletionMessageFunctionToolCall(OpenAIChatCompletionMessageFunctionT
 ChatCompletionMessageToolCall = ChatCompletionMessageFunctionToolCall | OpenAIChatCompletionMessageToolCall
 Function = OpenAIFunction
 CompletionUsage = OpenAICompletionUsage
+PromptTokensDetails = OpenAIPromptTokensDetails
 CreateEmbeddingResponse = OpenAICreateEmbeddingResponse
 Embedding = OpenAIEmbedding
 Usage = OpenAIUsage
--- a/src/any_llm/providers/anthropic/utils.py
+++ b/src/any_llm/providers/anthropic/utils.py
@@ -9,7 +9,6 @@ from anthropic.types import (
     MessageStopEvent,
 )
 from anthropic.types.model_info import ModelInfo as AnthropicModelInfo
-from openai.types.completion_usage import PromptTokensDetails
 
 from any_llm.exceptions import UnsupportedParameterError
 from any_llm.logging import logger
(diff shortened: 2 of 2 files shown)


njbrake:
Integration tests: https://github.com/mozilla-ai/any-llm/actions/runs/21183852088/job/60932943608

src/any_llm/types/completion.py
```diff
@@ -73,6 +74,7 @@ class ChatCompletionMessageFunctionToolCall(OpenAIChatCompletionMessageFunctionT
 ChatCompletionMessageToolCall = ChatCompletionMessageFunctionToolCall | OpenAIChatCompletionMessageToolCall
 Function = OpenAIFunction
 CompletionUsage = OpenAICompletionUsage
+PromptTokensDetails = OpenAIPromptTokensDetails
```
njbrake, line 77: This is probably the possibly most controversial part of the PR? Though I don't think it's incredibly controversial, but basically we use openai's existing setup for adding details about the prompts.

tests/unit/providers/test_anthropic_provider.py
njbrake, line 1: Really just skimmed these, Claude put these in here to get the needed code coverage

src/any_llm/types/completion.py
daavoo: yeah... I guess this technically fixes the bug but doesn't quite address:

> but ideally I'd like to have the option of getting provider specific usage fields in my response (e.g. the Anthropic specific cache_creation_input_tokens & cache_read_input_tokens)

An interesting take was https://github.com/mozilla-ai/any-llm/issues/622#issuecomment-3600233201

tests/unit/providers/test_anthropic_provider.py
daavoo: Don't know how far we want to go with integration tests around provider-specific logic but since anthropic is a really popular one maybe it's worth adding one that hits the actual API

daavoo (approved):
The change makes sense, considering our current approach of sticking to the imposed OpenAI "standard".

It's not a blocker but as followups would be nice to:

- test against the real Anthropic API with cache settings
- have a discussion about the option of allowing users to access the raw provider responses

src/any_llm/types/completion.py
njbrake: Yeah you'll get the read_input tokens after this PR, but not the creation_input tokens. I think we should move that to another PR: this PR overall is to fix the bug, but we should also explore the add of the feature, I agree.

tests/unit/providers/test_anthropic_provider.py
njbrake: My thought here was that since we're using the anthropic typed SDK, if they change an assumption on their side, their SDK will change and then our mypy/ruff checks will catch the issue.

njbrake:
> The change makes sense, considering our current approach of sticking to the imposed OpenAI "standard".
> 
> It's not a blocker but as followups would be nice to:
> 
>     * test against the real Anthropic API with cache settings
> 
>     * have a discussion about the option of allowing users to access the raw provider responses

I'm looking now to see how hard it would be to implement a translation layer between the OpenResponses standard and the ChatCompletion API. I'd really like to be able to recommend that most people leave the chatcompletion api and switch to the next gen standard.

PR #764, opened 2026-01-21
OpenResponses

njbrake:
## Description

Meant to be a launch point for discussion. I'm interested to hear thoughts about the approach. Basically we would now use the python spec autogenerated from the OpenResponses spec

Updating to make our "Responses" item refer to the OpenResponses spec. Since it's an extension to the OpenAI responses API it's relatively simple to make the initial extension.

## PR Type

- 🆕 New Feature
- 💅 Refactor
- 📚 Documentation

## Checklist

- [ ] I understand the code I am submitting.
- [ ] **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

- AI Model used: Opus 4.5
Claude was having a pretty tough time with this one 😆 . At first it gave me a lot of boilerplate and uneccessary testing. I whittled it down, but then Had to guide it a bit like "don't manually create the types, utilize the openapi spec from openresponses to generate the types programmatically"

### DIFF
--- a/src/any_llm/providers/huggingface/huggingface.py
+++ b/src/any_llm/providers/huggingface/huggingface.py
@@ -55,8 +55,8 @@ class HuggingfaceProvider(AnyLLM):
     ENV_API_KEY_NAME = "HF_TOKEN"
     PROVIDER_DOCUMENTATION_URL = "https://huggingface.co/docs/huggingface_hub/package_reference/inference_client"
 
-    # OpenResponses router endpoint for the Responses API
-    OPENRESPONSES_ROUTER_URL = "https://router.huggingface.co/v1"
+    # OpenResponses router endpoint for the OpenResponses API
+    OPENRESPONSES_ROUTER_URL = "https://evalstate-openresponses.hf.space/v1"
 
     SUPPORTS_COMPLETION_STREAMING = True
     SUPPORTS_COMPLETION = True
@@ -226,26 +226,7 @@ class HuggingfaceProvider(AnyLLM):
         )
 
         if isinstance(response, OpenAIResponse):
-            # HuggingFace router doesn't return fully spec-compliant OpenResponses.
-            # Fill in defaults for missing/null fields required by the spec.
-            # See: https://github.com/openresponses/openresponses/blob/main/schema/components/schemas/ResponseResource.json
-            # Waiting for HF team feedback https://huggingface.co/blog/open-responses#6977bfa7d4a00deea21b389c
-            data = response.model_dump()
-            defaults: dict[str, Any] = {
-                "truncation": "disabled",
-                "parallel_tool_calls": False,
-                "text": {"format": {"type": "text"}},
-                "presence_penalty": 0.0,
-                "frequency_penalty": 0.0,
-                "top_logprobs": 0,
-                "store": False,
-                "background": False,
-                "service_tier": "default",
-            }
-            for key, default in defaults.items():
-                if data.get(key) is None:
-                    data[key] = default
-            return ResponseResource.model_validate(data)
+            return ResponseResource.model_validate(response.model_dump())
 
         if isinstance(response, AsyncStream):
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -31,7 +31,7 @@ def provider_reasoning_model_map() -> dict[LLMProvider, str]:
         LLMProvider.DEEPSEEK: "deepseek-reasoner",
         LLMProvider.MOONSHOT: "kimi-thinking-preview",
         LLMProvider.BEDROCK: "us.anthropic.claude-haiku-4-5-20251001-v1:0",
-        LLMProvider.HUGGINGFACE: "huggingface/tgi",
+        LLMProvider.HUGGINGFACE: "openai/gpt-oss-120b:groq",
         LLMProvider.NEBIUS: "openai/gpt-oss-20b",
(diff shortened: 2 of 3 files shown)


pyproject.toml
```diff
@@ -230,6 +230,17 @@ omit = [
...
+[tool.datamodel-codegen]
+input = "src/any_llm/types/.openresponses/openapi.json"
+input-file-type = "openapi"
+output = "src/any_llm/types/openresponses_generated.py"
+output-model-type = "pydantic_v2.BaseModel"
+use-annotated = true
+field-constraints = true
+target-python-version = "3.11"
+use-double-quotes = true
+collapse-root-models = true
```
njbrake: Claude recommended I use this 🤷

njbrake:
@daavoo don't worry about an in depth review just yet, but the basic concept:

1. We swap so that instead of the OpenAI Responses, we use OpenResponses spec. We autogenerate that spec from their published openapi spec
2. As more providers come on board, we can then add them to support, or built an interface to support the use case.

daavoo:
> @daavoo don't worry about an in depth review just yet, but the basic concept:
> 
>     1. We swap so that instead of the OpenAI Responses, we use OpenResponses spec. We autogenerate that spec from their published openapi spec

Are those interchangeable? 
I am exited about openresponses but I am still not sure how this is going to land, so maybe the safest play is to have a separate option or method if they are not fully compatible.

njbrake:
> > @daavoo don't worry about an in depth review just yet, but the basic concept:
> > ```
> > 1. We swap so that instead of the OpenAI Responses, we use OpenResponses spec. We autogenerate that spec from their published openapi spec
> > ```
> 
> Are those interchangeable? I am exited about openresponses but I am still not sure how this is going to land, so maybe the safest play is to have a separate option or method if they are not fully compatible.

From my understanding from reading the HF blogpost and severl of the OR documents, OR is an extension of the OAI Responses APi so it works with the original OAI spec.

daavoo:
> From my understanding from reading the HF blogpost and severl of the OR documents, OR is an extension of the OAI Responses APi so it works with the original OAI spec.

let's go then

njbrake:
Integration tests: https://github.com/mozilla-ai/any-llm/actions/runs/21264563960

njbrake:
OK, here's where I ended up on this project. I was mistaken, the OpenAI Responses spec is not compliant with the OpenResponses spec. I was misinterpreting what they were saying in the press release: the openAI responses API is OpenResponses compatible, because they are putting some things into the pydantic "model_extra" field to make it compliant. So it's like yeah their API is compatible, but it's not because the two specifications are compatible 🙈

So this then puts us in an interesting situation going back to your previous comment. Do we make the wholesale switch to the open responses spec? If we do so now we have to make several providers be marked as not supporting the API when we previously had them supporting the API... So if anybody is using those providers with the responses API, we would be breaking them. However, it does seem likely that all the providers will fall into line soon, so even if we mark them as incompatible now it seems reasonable that that we would be able to reenable them soon.

I'm not really a big fan of introducing a new API end point I would so much rather just use and repurpose the old one.

njbrake:
Sidenote, this PR was a great example of where coding agents like Claude code were absolutely terrible and wasted a lot of my time. It couldn't seem to wrap its head around the hypothesis that openAI was doing things to make themselves compliant that existed outside of typical pydantic typing conventions, and it spun itself into circles, trying to convert everything into json dictionaries and then somehow make it all work.

After giving it a few HOURS prompting and trying to help it, figure it out on its own. I ended up looking into it myself, and then finding the problem in less than 30 minutes. Claude has helped me solve so many problems, but this was not one of them 😅

daavoo:
> So this then puts us in an interesting situation going back to your previous comment. Do we make the wholesale switch to the open responses spec? If we do so now we have to make several providers be marked as not supporting the API when we previously had them supporting the API... So if anybody is using those providers with the responses API, we would be breaking them. However, it does seem likely that all the providers will fall into line soon, so even if we mark them as incompatible now it seems reasonable that that we would be able to reenable them soon.

> I'm not really a big fan of introducing a new API end point I would so much rather just use and repurpose the old one.

I would actually prefer to just put it under a new method and wait to see where the initiative lands, but not a strong opinion (I don't know how many people uses the current openai responses)

njbrake:
> I would actually prefer to just put it under a new method and wait to see where the initiative lands, but not a strong opinion (I don't know how many people uses the current openai responses)

I guess that's the thing and what makes me conflicted: OpenAI is the one that put out the Responses API that we added support for. And now OpenAI (+ collaborators) is the one putting out the OpenResponses spec. So it seems like even if the purpose of the Responses API was to align with OpenAI, migrating ourselves to OpenResponses is what we need to do in order to remain compatible with the Responses API in the future.

And I'm not a huge fan of having one api called 'responses' and another called 'oresponses'... Seems confusing :(

.vscode/settings.json
njbrake, line 1: Daavoo and I both use this settings file and IMO it's a nice add for anyone using vscode/cursor etc. I realize it's slight scope creep

njbrake:
Integration tests: https://github.com/mozilla-ai/any-llm/actions/runs/21374141102

njbrake:
Claude Code really screwed me on this one lol (or I did a terrible job asking it what to do). I'm going to go back to basics, fix this feature into something well thought out, and then reopen it when it's ready

PR #775, opened 2026-01-27
Feat: Support OpenResponses Spec

njbrake:
## Description

This PR adds support for the OpenResponses Spec, which extends the OpenAI Responses spec.

See https://huggingface.co/blog/open-responses
This PR uses the python pydantic models autogenerated from the spec via https://github.com/mozilla-ai/openresponses-python

## PR Type

- 🆕 New Feature

## Checklist

- [ ] **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

- AI Model used: Opus 4.5
- AI Developer Tool used: Claude Code ish?

I ended up only using CC to help me type faster. When I let it think and make decisions, it kept trying to make more changes than needed and complicating things.

### DIFF
--- a/src/any_llm/any_llm.py
+++ b/src/any_llm/any_llm.py
@@ -493,6 +493,20 @@ class AnyLLM(ABC):
         parallel_tool_calls: int | None = None,
         reasoning: Any | None = None,
         text: Any | None = None,
+        presence_penalty: float | None = None,
+        frequency_penalty: float | None = None,
+        truncation: str | None = None,
+        store: bool | None = None,
+        service_tier: str | None = None,
+        user: str | None = None,
+        metadata: dict[str, str] | None = None,
+        previous_response_id: str | None = None,
+        include: list[str] | None = None,
+        background: bool | None = None,
+        safety_identifier: str | None = None,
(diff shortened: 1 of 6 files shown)


src/any_llm/any_llm.py
```diff
@@ -7,6 +7,8 @@
 from abc import ABC, abstractmethod
 from typing import TYPE_CHECKING, Any, ClassVar
 
+from openresponses_types import ResponseResource
```
daavoo, line 10: Wdyt about making our `any_llm.types.responses.Response` a union type that includes the old Response and the new ResponseResource? 
Would there be any downside?
We would avoid having to update imports and return types everywhere.

njbrake: I think I would rather not right now. People that are currently using the Responses API likely associate the Response object exported by any-llm as an alias for the OpenAI type, so if we make it a union then I worry people will be confused without a clear indication of the problem. By adding our Union typing then at least we make it obvious that changes are in process.

njbrake:
Integration tests: https://github.com/mozilla-ai/any-llm/actions/runs/21439569005

PR #812, opened 2026-02-14
feat: Return ParsedChatCompletion[T] for structured output across all providers

### LINKED ISSUE #806: `ParsedChatCompletion` type information lost during response conversion
## Description

When `response_format` is set to a Pydantic model, the OpenAI provider correctly calls `client.chat.completions.parse()`, which returns an `openai.ParsedChatCompletion[T]` with a typed `message.parsed: T` field containing a validated Pydantic model instance.

However, `_convert_chat_completion` round-trips the response through `model_dump()` → `ChatCompletion.model_validate()`, which serializes the Pydantic instance on `message.parsed` into a plain dict. The data is preserved (thanks to `extra="allow"` on `ChatCompletionMessage`), but the typed model is gone.

## Example

```python
from pydantic import BaseModel

class CalendarEvent(BaseModel):
    name: str
    date: str
    participants: list[str]

client = AnyLLM.create("openai")
response = await client.acompletion(
    model="gpt-4o",
    messages=[{"role": "user", "content": "List a birthday party event"}],
    response_format=CalendarEvent,
)

parsed = response.choices[0].message.parsed

# Expected: CalendarEvent(name='...', date='...', participants=['...'])
# Actual:   {'name': '...', 'date': '...', 'participants': ['...']}
```

Would it be possible to either preserve the `ParsedChatCompletion` type through the conversion, or expose a dedicated return path for structured output that keeps the typed `parsed` field?

bilelomrani1:
## Description

When `response_format` is set to a Pydantic model, `acompletion` now returns a `ParsedChatCompletion[T]` with a typed `message.parsed` field. This works uniformly across all providers that support `response_format`, not just OpenAI.

### Motivation

Previously, when using `response_format` with a Pydantic model, providers returned a plain `ChatCompletion` and users had to manually deserialize `message.content` from JSON. The OpenAI SDK's `.parse()` method handles this automatically, but that behavior was provider-specific and lost during any-llm's response conversion.

Before this PR, there was an inconsistency in `response_format` behavior across providers: the OpenAI provider used `client.chat.completions.parse()`, which returned parsed Pydantic objects and performed client-side validation, but any-llm's response conversion pipeline discarded the richer `ParsedChatCompletion` type, reducing it back to a plain `ChatCompletion`. Meanwhile, all other providers returned raw JSON strings in `message.content` with no parsing or validation. This PR bridges that gap by introducing `ParsedChatCompletion[T]` as a generic subclass of `ChatCompletion` (mirroring the OpenAI SDK's type hierarchy) and populating `message.parsed` at the `AnyLLM` layer so all providers behave consistently.

### Design decisions

**Unified parsing at the `AnyLLM` layer, not in providers.** All providers (OpenAI, Mistral, etc.) return a plain `ChatCompletion` from `_acompletion`. The structured output parsing (converting `message.content` JSON into a Pydantic model and attaching it to `message.parsed`) happens once in `acompletion()`. This is consistent with how Mistral already worked (using `client.chat.complete_async()`, not a higher-level parse method). The OpenAI provider was switched from `client.chat.completions.parse()` to `client.chat.completions.create()` to match this pattern.

**No exceptions on truncation/content filter.** When `finish_reason` is `"length"` or `"content_filter"`, `message.parsed` is set to `None` instead of raising an exception. The raw (possibly truncated) content remains available in `message.content`, and users can inspect `choice.finish_reason` to detect these cases. I considered raising exceptions to align with the OpenAI SDK's `.parse()` behavior, but this would add additional breaking changes: existing code using `acompletion` with `response_format` does not expect exceptions on these finish reasons. These exceptions could be added in a future major version. Note that this decision reduces breaking changes but does not fully eliminate them; see the Breaking changes section below.

### Breaking changes

When `response_format` is a Pydantic model, `acompletion` now validates the response content against the schema via `model_validate_json`. If the LLM returns JSON that doesn't conform to the model, a `pydantic.ValidationError` is raised where previously the raw string would sit in `message.content` without validation.

In practice, the impact depends on the provider:

- **OpenAI:** The previous implementation already used `client.chat.completions.parse()`, which performed Pydantic validation client-side (on top of server-side constrained decoding). Switching to `client.chat.completions.create()` with any-llm's own `model_validate_json` call preserves the same validation behavior, so this path should not surface new errors.
- **Other providers:** These providers may not always enforce the schema strictly. Responses that previously returned silently with non-conforming JSON in `message.content` will now raise `pydantic.ValidationError`. Let me know if this is an acceptable change, it is breaking, but better aligns with OpenAI semantics.

The return type change itself (`ParsedChatCompletion` instead of `ChatCompletion`) is non-breaking since `ParsedChatCompletion` is a subclass of `ChatCompletion`.

## PR Type

- 🆕 New Feature

## Relevant issues

Fixes #806

## Checklist

- **AI Usage:**
  - [ ] No AI was used.

## AI Usage Information

### DIFF
--- a/src/any_llm/any_llm.py
+++ b/src/any_llm/any_llm.py
@@ -369,6 +369,9 @@ class AnyLLM(ABC):
         msg = "Subclasses must implement this method"
         raise NotImplementedError(msg)
 
+    # Overloads let type checkers narrow the return type based on response_format and stream.
+    # The implementation only declares these discriminating params; everything else
+    # passes through via **kwargs to acompletion().
     @overload
     def completion(
         self,
@@ -437,6 +440,7 @@ class AnyLLM(ABC):
 
         return async_iter_to_sync_iter(response)
 
+    # Overloads let type checkers narrow the return type based on response_format and stream.
     @overload
     async def acompletion(
         self,


njbrake:
Hi! A few comments from an Opus 4.6 guided review:

 completion() at src/any_llm/any_llm.py:368 wraps acompletion() but its return type is still ChatCompletion | Iterator[ChatCompletionChunk]. When a user passes response_format=MyModel, the
  synchronous path will return a ParsedChatCompletion at runtime but the type signature won't reflect that. The isinstance(response, ChatCompletion) check at line 378 will work correctly at runtime
  (since ParsedChatCompletion is a subclass), but callers using static type checking won't get ParsedChatCompletion[T] narrowing.

  Consider adding matching @overload signatures to completion(), or at minimum updating its return type to include ParsedChatCompletion[Any]

bilelomrani1:
Thanks for the review @njbrake, I addressed your point and the Codecov coverage gaps. There was an unreachable `isinstance(choice.message.parsed, response_format)` branch in the first version of this submission, I just removed it since now (by design) no provider returns pre-parsed objects and AnyLLM owns the validation step.

BTW, curious to have your opinion on this:

> I considered raising exceptions to align with the OpenAI SDK's .parse() behavior, but this would add additional breaking changes: existing code using acompletion with response_format does not expect exceptions on these finish reasons. These exceptions could be added in a future major version.

We technically already have a breaking change in this PR, so if you want to fully align with OpenAI semantics, we can raise on `finish_reason="length" or "content_filter"`, your choice.

njbrake:
Thanks @bilelomrani1! I'm ok with this breaking change, and yes lets defer raising those exceptions to a future release

Maybe just add a test? claude says

```
  1. Missing test for ValidationError on non-conforming JSON (coverage gap)

  The PR description highlights that pydantic.ValidationError is now raised when a provider returns JSON that doesn't conform to the schema. There's no unit test for this path. This is the main behavioral breaking
  change and should be tested.

  # Suggested test
  @pytest.mark.asyncio
  async def test_parsed_completion_invalid_json_raises_validation_error(provider: AnyLLM) -> None:
      provider._acompletion = AsyncMock(
          return_value=_make_chat_completion(content='{"wrong_field": "value"}'),
      )
      with pytest.raises(ValidationError):
          await provider.acompletion(
              model="test-model",
              messages=[{"role": "user", "content": "test"}],
              response_format=CityResponse,
          )
```

bilelomrani1:
Just added the test.

njbrake:
Lint and integration tests reveal some issues @bilelomrani1

bilelomrani1:
Looking at it. Some issues are mine but some are pre-existing (tests/integration/test_agent_loop.py:57 etc) not sure how these slipped though. I'll fix everything.

EDIT: they're caused by my `@overload` changes, we previously had to ignore some mypy errors but now the overloads fix these type checks errors, so nothing slipped.

bilelomrani1:
@njbrake It turns out the OpenAI SDK does not accept a Pydantic model as `response_format` when using `.create()` (which makes sense). The SDK also does quite a bit internally to convert the Pydantic model into JSON Schema, so it’s probably not wise for us to take ownership of that logic in any-llm.

Given all this, I think the best solution is to keep the `.parse` and `.create` distinction internally (exactly as it was before this PR) and handle the Pydantic validation path ourselves for providers that don’t support `.parse`. It’s a minor inconsistency, but it should be fine.

I also added a short circuit to ensure we don’t run Pydantic validation twice: if the wrapped SDK has already performed validation, there’s no need to run it again.

PR #827, opened 2026-02-23
fix(gateway): pricing lookup uses wrong separator, cost always NULL

sarmientoF:
## Description

`_log_usage` in `chat.py` builds the pricing `model_key` using colon separator (`provider:model`), but `pricing_init.py` stores keys using slash separator (`provider/model`). This mismatch means the DB lookup never finds a match, so `cost` is always `NULL` and user spend is never updated for chat completions.

The fix tries `provider/model` first (matching the convention used by `pricing_init.py`, `audio.py`, and `search.py`), then falls back to `provider:model` for backwards compatibility with pricing configured via the API using colon format.

### Why the existing tests don't catch this

The `model_pricing` fixture in `conftest.py` creates pricing with `model_key = "gemini:gemini-2.5-flash"` (colon format), which happens to match the old lookup. In production, pricing seeded via `pricing_init.py` uses slash format — the two never match.

## PR Type
- [x] Bug Fix

## Relevant Issues
N/A

## Checklist
- [x] I have read and understand the existing codebase and relevant files
- [x] I have tested these changes locally
- [x] Tests pass with my changes
- [ ] I have updated documentation (if applicable)
- [x] I have read the [Contributing Guidelines](CONTRIBUTING.md)
- [x] I have checked my code follows the project's code style

## AI Disclosure
- **AI Model used**: Claude Opus 4.6
- **AI Developer Tool used**: Claude Code
- **Additional context**: Bug was discovered during production debugging (cost column showing NULL for all chat completions). The fix and PR were pair-programmed with Claude Code.

### DIFF
--- a/.github/workflows/gateway-tests.yml
+++ b/.github/workflows/gateway-tests.yml
@@ -49,7 +49,19 @@ jobs:
         file: ./coverage.xml
         fail_ci_if_error: false
 
+  openapi-spec:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v6
+
+    - uses: astral-sh/setup-uv@v7
+      with:
+        python-version: 3.13
+        activate-environment: true
+
+    - name: Install dependencies
+      run: uv sync --extra gateway --extra all -U --python=3.13
 
     - name: Verify OpenAPI spec is up to date
-      run: |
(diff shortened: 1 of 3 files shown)


sarmientoF:
Screenshot 2026-02-23 at 17 20 47

PR #832, opened 2026-02-23
fix(anthropic): strip OpenAI-specific fields from messages

khaledosman:
## Description

Messages forwarded from OpenAI format to Anthropic's API include extra fields like `refusal`, `audio`, `function_call`, and `annotations` that Anthropic rejects with `400: messages.X.refusal: Extra inputs are not permitted`.

`_convert_messages_for_anthropic()` already reconstructs tool call and tool result messages with only `role` + `content`, but regular user/assistant messages were appended as-is with all original fields. This fix sanitizes all messages to only include Anthropic-compatible fields (`role`, `content`) before appending.

The `refusal` field enters the pipeline through any-llm when the SDK processes ChatCompletionMessage objects (from OpenAI responses used in multi-turn agent loops), it serializes them with model_dump(exclude_none=True,
  exclude={"reasoning"}) at any_llm.py:450 — this excludes reasoning but not refusal. If refusal is a non-None string, it gets included in the messages dict sent to the next provider.

## PR Type

- 🐛 Bug Fix

## Relevant issues

Fixes runtime `400 Bad Request` errors when conversations containing OpenAI-originated assistant messages (which include a `refusal` field) are routed to Anthropic.

## Checklist

- **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

- Any other info you'd like to share: Used to identify the root cause and draft the fix + test.

### DIFF
--- a/src/any_llm/providers/llamafile/llamafile.py
+++ b/src/any_llm/providers/llamafile/llamafile.py
@@ -46,10 +46,4 @@ class LlamafileProvider(BaseOpenAIProvider):
                 msg,
                 self.PROVIDER_NAME,
             )
-        if params.tools:
-            msg = "tools"
-            raise UnsupportedParameterError(
-                msg,
-                self.PROVIDER_NAME,
-            )
         return await super()._acompletion(params, **kwargs)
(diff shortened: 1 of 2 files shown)


njbrake:
Integration test: https://github.com/mozilla-ai/any-llm/actions/runs/22309790783

PR #853, opened 2026-02-24
feat(mzai): add mzai provider with platform JWT token authentication

dpoulopoulos:
## Description

Add a new `mzai` provider that allows routing requests through the any-llm platform itself. When `mzai` is selected as the underlying provider via `PlatformProvider`, the platform JWT token is used directly as the API key (via `platform_client._aensure_valid_token()`) instead of fetching a decrypted provider key. Usage events are skipped for mzai since the platform handles tracking internally.

Key changes:
- New `MzaiProvider` class extending `BaseOpenAIProvider`, pointing at the platform API
- `PlatformProvider._initialize_provider()` now branches on `mzai` to use JWT auth instead of `get_decrypted_provider_key()`
- Usage event posting is gated on `provider_key_id is not None`, so mzai requests skip client-side usage logging
- `MZAI` added to the `LLMProvider` enum in constants

## PR Type

- 🆕 New Feature

## Checklist
- **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

- Any other info you'd like to share: -

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,7 @@ dependencies = [
 [project.optional-dependencies]
 
 all = [
-  "any-llm-sdk[mistral,anthropic,huggingface,gemini,vertexai,vertexaianthropic,cohere,cerebras,fireworks,groq,bedrock,azure,azureopenai,watsonx,together,sambanova,ollama,moonshot,nebius,xai,databricks,deepseek,inception,openai,openrouter,portkey,lmstudio,llama,voyage,perplexity,platform,llamafile,llamacpp,sagemaker,gateway,zai,minimax,vllm]"
+  "any-llm-sdk[mistral,anthropic,huggingface,gemini,vertexai,vertexaianthropic,cohere,cerebras,fireworks,groq,bedrock,azure,azureopenai,watsonx,together,sambanova,ollama,moonshot,nebius,xai,databricks,deepseek,inception,openai,openrouter,portkey,lmstudio,llama,voyage,perplexity,platform,llamafile,llamacpp,sagemaker,gateway,zai,minimax,mzai,vllm]"
 ]
 
 platform = [
@@ -115,6 +115,7 @@ openrouter = []
 portkey = []
 sambanova = []
 minimax = []
+mzai = []
 vllm = []
 gateway = [
   "fastapi>=0.115.0",


src/any_llm/providers/platform/platform.py
```diff
@@ -296,9 +299,19 @@ def provider(self, provider_class: type[AnyLLM]) -> None:
...
-        )
-        self.provider_key_id = str(provider_key_result.provider_key_id)
-        self.project_id = str(provider_key_result.project_id)
-        self._provider = provider_class(api_key=provider_key_result.api_key, api_base=self.api_base, **self.kwargs)
+
+        if provider_class.PROVIDER_NAME == LLMProvider.MZAI.value:
+            # For mzai, use JWT token directly as API key
+            from any_llm.utils.aio import run_async_in_sync
+
+            token = run_async_in_sync(self.platform_client._aensure_valid_token(self.any_llm_key))
```
njbrake: Claude says: 

The JWT token is fetched once during provider.setter and passed as a static api_key. JWT tokens typically have expiry times. If a PlatformProvider instance using mzai is long-lived (e.g., in a web server), the token
  could expire, leading to authentication failures without any re-fetch mechanism. This may be acceptable for the current intended use case, but it's worth considering whether a token refresh mechanism is needed

njbrake:
@dpoulopoulos Can you explain a bit about the difference between Platform provider and MZAI? I'm not tracking on why we should have the two providers instead of one.

Also, I think i recall seeing some product documents back in the day saying that we weren't suppozed to use the MZAI acronym externally, not sure if that is still true.

src/any_llm/providers/platform/platform.py
dpoulopoulos: The `nay-llm-platform-client` takes care of refreshing expired tokens: https://github.com/mozilla-ai/any-llm-platform-client/blob/6c669fc99daad2c0847e3927b24d85bc0d10d9f9/src/any_llm_platform_client/client.py#L396

dpoulopoulos:
> @dpoulopoulos Can you explain a bit about the difference between Platform provider and MZAI? I'm not tracking on why we should have the two providers instead of one.
> 
> Also, I think i recall seeing some product documents back in the day saying that we weren't suppozed to use the MZAI acronym externally, not sure if that is still true.

About the MZAI acronym, yes, you may be right, I'll talk to product about that. It's a dev remnant.

Regarding the first question: PlatformProvider is a "hollow" provider, essentially a wrapper for every other provider we support. In contrast, the Mozilla.ai provider is a proper provider, even though we are currently using Nebius on the backend. My thought is that since we plan to engage in different partnerships in the future, I want to keep this separate. At some point, it may not be necessary to go through the platform. While it might not seem obvious right now, I believe separating these concerns and treating the MZAI provider like any other gives us more flexibility.

njbrake:
Following the PR #861 , this PR gets a little more straightfowards I think. Since "platform" isn't a real provider and is no longer listed as such, we can wait for the product name once it's been picked and then name the provider being added in this PR as that name.

So then in the meantime, I think i'm fine with calling this 'mzai' because we know we'll need to change it once the platform receives its name 👍

njbrake:
Hi @dpoulopoulos a few tweaks (Collab with Claude and I):

```
  Request Changes — The issues are relatively minor but worth addressing before merge:
  - The import-time os.getenv for API_BASE diverges from the established pattern and could cause subtle issues in testing. Switching to ENV_API_BASE_NAME (or hardcoding) is a small change.
  - The section-separator comments should be removed per project guidelines.
  - The _aensure_valid_token private method access should at least be acknowledged — either add a # TODO noting it should be promoted to public API, or confirm with platform-client maintainers.
  - ```

The any-llm  CLAUDE.md/AGENTS.md has "Do not add decorative section-separator comments."

File: src/any_llm/providers/mzai/mzai.py [hunk 2]
```diff
@@ -0,0 +1,21 @@
...
+    PROVIDER_DOCUMENTATION_URL = "https://any-llm.ai"
+    SUPPORTS_RESPONSES = False
+    SUPPORTS_LIST_MODELS = False
+    SUPPORTS_BATCH = False
+    SUPPORTS_COMPLETION_STREAMING = False
+    SUPPORTS_COMPLETION = True
+    SUPPORTS_COMPLETION_REASONING = False
+    SUPPORTS_COMPLETION_IMAGE = False
+    SUPPORTS_COMPLETION_PDF = False
+    SUPPORTS_EMBEDDING = True
```
njbrake, line 19: Has this been verified? I'm surprised that streaming, reasoning, image, pdf, and list models would all be unsupported

dpoulopoulos: This is a challenge because Nebius serves multiple models, ranging from very small to very powerful. I’m erring on the safe side for now. But how should this work? if even one model supports the feature, can we set it to `True`, or do all models need to support it?

dpoulopoulos:
> Hi @dpoulopoulos a few tweaks (Collab with Claude and I):
> 
> ```
>   Request Changes — The issues are relatively minor but worth addressing before merge:
>   - The import-time os.getenv for API_BASE diverges from the established pattern and could cause subtle issues in testing. Switching to ENV_API_BASE_NAME (or hardcoding) is a small change.
>   - The section-separator comments should be removed per project guidelines.
>   - The _aensure_valid_token private method access should at least be acknowledged — either add a # TODO noting it should be promoted to public API, or confirm with platform-client maintainers.
>   - ```
> 
> The any-llm  CLAUDE.md/AGENTS.md has "Do not add decorative section-separator comments." 
> ```

Most of these changes relate to the platform provider, which this PR doesn't touch, so we can address them in a follow-up PR.

On the other hand, we need `os.getenv` for `API_BASE` in the mzai provider for local development, as it allows us to run the platform locally. This is primarily for convenience.

src/any_llm/providers/mzai/mzai.py
njbrake: Yes we've generally been going with "if the provider supports it, mark it True" even if not all models inside the provider support it. So If nebius is our initial provider for mzai, you could probably copy all the attributes from the existing Nebius provider.

njbrake:
> On the other hand, we need os.getenv for API_BASE in the mzai provider for local development, as it allows us to run the platform locally. This is primarily for convenience.

I think what it's suggesting here is that you use the convention for any_llm, where the API base url is loaded from a env var specific to that provider, like

https://github.com/mozilla-ai/any-llm/blob/d6954ce4d9065571adffa7a584845be55727ccb6/src/any_llm/providers/moonshot/moonshot.py#L5-L7

which is then handled automatically here:

https://github.com/mozilla-ai/any-llm/blob/d6954ce4d9065571adffa7a584845be55727ccb6/src/any_llm/any_llm.py#L121-L133

I think this would be able to accomplish your needs while sticking with the conventions?

dpoulopoulos:
> > On the other hand, we need os.getenv for API_BASE in the mzai provider for local development, as it allows us to run the platform locally. This is primarily for convenience.
> 
> I think what it's suggesting here is that you use the convention for any_llm, where the API base url is loaded from a env var specific to that provider, like
> 
> https://github.com/mozilla-ai/any-llm/blob/d6954ce4d9065571adffa7a584845be55727ccb6/src/any_llm/providers/moonshot/moonshot.py#L5-L7
> 
> which is then handled automatically here:
> 
> https://github.com/mozilla-ai/any-llm/blob/d6954ce4d9065571adffa7a584845be55727ccb6/src/any_llm/any_llm.py#L121-L133
> 
> I think this would be able to accomplish your needs while sticking with the conventions?

Ah, okay! I'm gonna fix this, thanks!

src/any_llm/providers/mzai/mzai.py
dpoulopoulos: For now, we don't support listing the models through the platform backend. I'll add that feature to the backend and amend the code here in a follow-up PR.

src/any_llm/providers/mzai/mzai.py
```diff
@@ -0,0 +1,19 @@
...
+    PROVIDER_DOCUMENTATION_URL = "https://any-llm.ai"
+    SUPPORTS_RESPONSES = False
+    SUPPORTS_LIST_MODELS = False
+    SUPPORTS_BATCH = False
+    SUPPORTS_COMPLETION_STREAMING = False
+    SUPPORTS_COMPLETION = True
+    SUPPORTS_COMPLETION_REASONING = True
+    SUPPORTS_COMPLETION_IMAGE = False
+    SUPPORTS_COMPLETION_PDF = False
+    SUPPORTS_EMBEDDING = True
```
njbrake, line 19: Nebius just has this

https://github.com/mozilla-ai/any-llm/blob/d6954ce4d9065571adffa7a584845be55727ccb6/src/any_llm/providers/nebius/nebius.py#L11-L12

So I believe you can put that exactly (since the other values will continue to be inherited from the BaseOpenAIProvider class)

PR #861, opened 2026-02-25
fix: platform should not be a standalone provider

njbrake:
## Description

The platform isn't a provider, it's a meta-provider. It should never be directly instantiated, it only kicks in to be used when `ANY_LLM_KEY` is set. Because of this, we shouldn't include it in the typical list of providers.  

  - Removed PLATFORM from LLMProvider enum — Platform is a meta-provider (vault + wrapper), not a real provider. Attempting provider="platform" now raises UnsupportedProviderError like any unknown provider.
  - Added missing delegation methods — PlatformProvider previously only delegated _acompletion. Added delegation for _aresponses, _aembedding, _alist_models, _acreate_batch, _aretrieve_batch, _acancel_batch, and
  _alist_batches.
  - Made provider initialization lazy and async — The provider setter now stores the provider class without making HTTP calls. The actual provider instance is created on first async method call via
  _ensure_provider_initialized(), which uses aget_decrypted_provider_key (async) instead of the sync variant that blocked the event loop.
  - Synced SUPPORTS_* flags to wrapped provider — Flags were previously hardcoded to True. They are now updated from the wrapped provider class when the provider setter is called, so capability checks accurately
  reflect what the underlying provider supports.
  - Replaced fragile providers_without_stream_options set — Instead of maintaining a central hardcoded set, providers that don't support stream_options (Cerebras, Cohere, Mistral, Ollama, Together) now strip it in
   their own _convert_completion_params. PlatformProvider always injects stream_options={"include_usage": True} for streaming and lets each provider decide what to do with it.
  - Fixed vertexaianthropic in test_provider_factory_can_create_all_supported_providers — Added missing project_id kwarg.

From my research, the strategy of dropping stream_options when it isn't supported is consistent with how other frameworks like litellm handle it.

## PR Type

- 🐛 Bug Fix
- 💅 Refactor
- 📚 Documentation

## Checklist

- [ ] **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

- AI Model used: Opus 4.6
- AI Developer Tool used: AoE + CC

### DIFF
--- a/.github/workflows/tests-integration.yaml
+++ b/.github/workflows/tests-integration.yaml
@@ -27,7 +27,7 @@ jobs:
 
       - name: Set expected providers for local integration tests
         id: set_local_providers
-        run: echo "expected_providers=ollama,llamacpp,llamafile" >> $GITHUB_OUTPUT
+        run: echo "expected_providers=ollama,llamacpp,llamafile,lmstudio" >> $GITHUB_OUTPUT
 
   determine-jobs-to-run:
     needs: expected-providers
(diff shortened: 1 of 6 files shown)


njbrake:
Integration test: https://github.com/mozilla-ai/any-llm/actions/runs/22402422149

PR #864, opened 2026-02-26
fix(openai): remap max_tokens from kwargs and warn on conflict

### LINKED ISSUE #862: OpenAI provider should map max_tokens to max_completion_tokens for newer models
## Summary

OpenAI has deprecated `max_tokens` in favour of `max_completion_tokens` for newer models (gpt-5.x, o-series). The OpenAI provider passes `max_tokens` through to the API unchanged via `model_dump()`, causing a 400 error:

```
Unsupported parameter: 'max_tokens' is not supported with this model. Use 'max_completion_tokens' instead.
```

## Expected Behaviour

Callers should be able to set `max_tokens` in `acompletion()` and have each provider map it to the correct API parameter internally. This is what the Gemini provider already does:

```python
# gemini/base.py:94-95
if params.max_tokens is not None:
    kwargs["max_output_tokens"] = params.max_tokens
```

The OpenAI provider should do the equivalent:

```python
if params.max_tokens is not None:
    converted_params["max_completion_tokens"] = converted_params.pop("max_tokens")
```

## Actual Behaviour

`acompletion(max_tokens=8192, provider="openai", model="gpt-5.2")` sends `max_tokens` directly to the OpenAI API, which rejects it for newer models.

The workaround is to pass `max_completion_tokens` instead, but this leaks provider-specific semantics to the caller and breaks the "switch providers by changing one line" promise.

## Reproduction

```python
from any_llm import acompletion

response = await acompletion(
    model="gpt-5.2",
    provider="openai",
    messages=[{"role": "user", "content": "Hello"}],
    max_tokens=8192,
)
# Error: 400 - Unsupported parameter: 'max_tokens'
```

## Environment

- any-llm-sdk 1.8.6
- OpenAI SDK (latest)
- Model: gpt-5.2

peteski22:
## Description

`BaseOpenAIProvider._convert_completion_params()` applied `kwargs` after the `max_tokens` → `max_completion_tokens` remapping, so `max_tokens` or `max_completion_tokens` passed via `**kwargs` bypassed the mapping entirely.

This fix moves `converted_params.update(kwargs)` before the remapping logic so all sources of these parameters are handled consistently. When both `max_tokens` and `max_completion_tokens` are present (from any combination of params and kwargs), a warning is logged and `max_completion_tokens` takes precedence.

## PR Type

- 🐛 Bug Fix

## Checklist
- **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

Fixes: #862

### DIFF
--- a/docs/gateway/openapi.json
+++ b/docs/gateway/openapi.json
@@ -890,13 +890,6 @@
       },
       "ValidationError": {
         "properties": {
-          "ctx": {
-            "title": "Context",
-            "type": "object"
-          },
-          "input": {
-            "title": "Input"
-          },
           "loc": {
             "items": {
               "anyOf": [
--- a/docs/gateway/openapi.json
+++ b/docs/gateway/openapi.json
@@ -890,6 +890,13 @@
       },
       "ValidationError": {
         "properties": {
+          "ctx": {
+            "title": "Context",
+            "type": "object"
+          },
+          "input": {
+            "title": "Input"
+          },
           "loc": {
             "items": {
               "anyOf": [
--- a/src/any_llm/providers/openai/base.py
+++ b/src/any_llm/providers/openai/base.py
@@ -59,6 +59,26 @@ class BaseOpenAIProvider(AnyLLM):
         """Convert CompletionParams to kwargs for OpenAI API."""
         converted_params = params.model_dump(exclude_none=True, exclude={"model_id", "messages"})
         converted_params.update(kwargs)
+
+        max_tokens_key = "max_tokens"
+        max_completion_tokens_key = "max_completion_tokens"
+
+        # Attempt to map max_tokens to max_completion_tokens for OpenAI API compatibility.
+        max_tokens = converted_params.pop(max_tokens_key, None)
+
+        # We can return early if there's no max tokens to deal with.
+        if max_tokens is None:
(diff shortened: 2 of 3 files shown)


njbrake:
Hi @peteski22 , I opened up 865, with what is a slightly more conservative change from this one (only updating functionality for openai itself, not the base provider that is shared across other providers)

peteski22:
> Hi @peteski22 , I opened up 865, with what is a slightly more conservative change from this one (only updating functionality for openai itself, not the base provider that is shared across other providers)

Whatever works. 

Just a bit unsure why a new PR was needed vs. making the adjustments here if you'd commented directly on this one. 

😕 But no worries, I'll close this.

njbrake:
> > Hi @peteski22 , I opened up 865, with what is a slightly more conservative change from this one (only updating functionality for openai itself, not the base provider that is shared across other providers)
> 
> Whatever works.
> 
> Just a bit unsure why a new PR was needed vs. making the adjustments here if you'd commented directly on this one.
> 
> 😕 But no worries, I'll close this.

Sorry about that: it's because my agent had been working on the issue and had the fix locallly before I had seen your PR. So I asked it to compare our two PRs, it explained the difference, and then I told it to post my PR with a link to yours.

PR #894, opened 2026-03-02
feat(anthropic): support adaptive thinking

### LINKED ISSUE #893: feat: support adaptive thinking
https://platform.claude.com/docs/en/build-with-claude/adaptive-thinking

I'm guessing more providers will adopt settings like this... Maybe we let you pass through whatever you want in reasoning effort? Need to do some research

njbrake:
## Description

Replace the legacy `thinking: {type: "enabled", budget_tokens: N}` approach with Anthropic's recommended `thinking: {type: "adaptive"}` + `output_config: {effort: ...}` for controlling extended thinking in the Anthropic provider.

The old `budget_tokens` approach is deprecated on Opus 4.6 and Sonnet 4.6. Adaptive thinking lets Claude dynamically determine when and how much to use extended thinking, while the `effort` parameter provides soft guidance on thinking depth.

Default behavior (`reasoning_effort="auto"`) is unchanged — no `thinking` parameter is sent, preserving backward compatibility. Only explicit reasoning effort levels opt into the new adaptive thinking mode.

### Behavior mapping

| `reasoning_effort` | Before | After |
|---|---|---|
| `None` / `"none"` | `thinking: {"type": "disabled"}` | `thinking: {"type": "disabled"}` (unchanged) |
| `"auto"` (default) | No thinking param | No thinking param (unchanged) |
| `"minimal"` / `"low"` | `budget_tokens: 1024/2048` | `thinking: {"type": "adaptive"}` + `effort: "low"` |
| `"medium"` | `budget_tokens: 8192` | `thinking: {"type": "adaptive"}` + `effort: "medium"` |
| `"high"` | `budget_tokens: 24576` | `thinking: {"type": "adaptive"}` + `effort: "high"` |
| `"xhigh"` | `budget_tokens: 32768` | `thinking: {"type": "adaptive"}` + `effort: "max"` |

When `response_format` is also set, both `format` and `effort` are merged into `output_config`.

### Breaking change note

Users targeting **pre-4.6 Claude models** (e.g., `claude-3-7-sonnet`) with an explicit `reasoning_effort` value will get API errors after upgrading, because `thinking: {"type": "adaptive"}` is only supported on Claude Opus 4.6 and Sonnet 4.6+. The old `budget_tokens` approach is no longer used for the Anthropic provider.

**If you are affected**, either:
- Upgrade to a Claude 4.6 model, or
- Remove the `reasoning_effort` parameter (the default `"auto"` behavior is unchanged and works on all models)

This does **not** affect the Bedrock provider, which still uses `budget_tokens` independently.

## PR Type

- 🆕 New Feature

## Relevant issues

Fixes #893

## Checklist

- **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

- Any other info you'd like to share: N/A

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -12,7 +12,7 @@ from tests.constants import CI_EXCLUDED_PROVIDERS, INCLUDE_LOCAL_PROVIDERS, INCL
 @pytest.fixture
 def provider_reasoning_model_map() -> dict[LLMProvider, str]:
     return {
-        LLMProvider.ANTHROPIC: "claude-sonnet-4-0",
+        LLMProvider.ANTHROPIC: "claude-sonnet-4-6",
         LLMProvider.GEMINI: "gemini-2.5-flash",
         LLMProvider.GATEWAY: "gpt-5-nano",
         LLMProvider.VERTEXAI: "gemini-2.5-flash",


njbrake:
@daavoo could I get your opinion on this? This would be dropping our token_budget mapping which is clearly the future (since they deprecated it for the 4.6 suite) but it's a breaking change for our old users. I'd rather just mark it as a breaking change in the release notes and move on with merge

daavoo:
> @daavoo could I get your opinion on this? This would be dropping our token_budget mapping which is clearly the future (since they deprecated it for the 4.6 suite) but it's a breaking change for our old users. I'd rather just mark it as a breaking change in the release notes and move on with merge

Just merging and noting as breaking change sounds good to me

Issue #908, opened 2026-03-11
Research: retrieve token usage for cancelled streaming requests via OTEL


dpoulopoulos:
## Context

When a streaming request is cancelled (e.g. Ctrl+C, `CancelledError`), the OTEL trace is reported to the platform. However, token usage attributes (`input_tokens`, `output_tokens`) are absent on the exported span.

This happens because LLM providers only include usage data in the **final** streaming chunk (when `stream_options={"include_usage": True}`). If the stream is cancelled before that chunk arrives, usage counts are unavailable and the platform displays 0.

## Research Questions

1. Does the OpenTelemetry GenAI semantic conventions specification define a mechanism or recommendation for reporting usage on incomplete/cancelled streams?
2. Do any OTEL-instrumented LLM libraries handle this case?
3. Is there a provider-agnostic way to retrieve usage after cancellation (e.g. a separate API call, or usage data sent in earlier chunks by some providers)?
4. Should we estimate `output_tokens` from the content received in partial chunks, or is it better to leave usage absent and handle it on the platform display side?

## Relevant Code

- `src/any_llm/providers/platform/platform.py` — `_stream_with_usage_tracking()` (the `finally` block handling cancellation)
- `src/any_llm/providers/platform/utils.py` — `export_completion_trace()`

JiwaniZakir:
The root cause is straightforward: `stream_options={"include_usage": True}` causes providers to append usage data only in a terminal chunk (typically with an empty `choices` array), so any cancellation before that chunk is processed leaves the span attributes unpopulated. In `_stream_with_usage_tracking()`, the `finally` block runs but has no usage data to report to `export_completion_trace()`. One partial mitigation worth evaluating: some providers (e.g. OpenAI with certain configurations) emit running usage counts on intermediate chunks — checking whether `usage` is non-null on each chunk and storing the last seen value before the `finally` block would at least capture a lower-bound count on cancellation. For `output_tokens` specifically, a token-count estimate from the accumulated partial content (via `tiktoken` or equivalent) could serve as a fallback, though it should be flagged as an estimate in the span attributes to avoid misleading the platform.

njbrake:
Obsolete. Both functions this research targets, `_stream_with_usage_tracking()` and `export_completion_trace()`, were removed with the platform provider in #1088. There is no cancellation path left to instrument.

Reopen against a specific emitter if this becomes relevant again.

[...]

PR #909, opened 2026-03-11
fix(platform): export OTEL trace on cancelled streaming requests

dpoulopoulos:
## Description

When a streaming request is cancelled (Ctrl+C, `CancelledError`, `GeneratorExit`), the OTEL trace span was silently dropped because cancellation exceptions inherit from `BaseException`, not `Exception`. This meant no data was reported to the platform for cancelled streams.

Uses a `span_ended` flag in the `finally` block of `_stream_with_usage_tracking()` to detect unhandled cancellation and end the span with partial attributes (`anyllm.stream.cancelled = True`, model name, and any available usage data). The span is ended before `deactivate_trace_export()` runs so the forwarding processor still sees it as active.

Token usage will be absent on cancelled streams since providers only send usage in the final chunk — tracked in #908.

## PR Type

- 🐛 Bug Fix

## Relevant issues

Refs #908

## Checklist

- **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

- Any other info you'd like to share: n/a.

### DIFF
--- a/tests/unit/providers/test_platform_provider.py
+++ b/tests/unit/providers/test_platform_provider.py
@@ -1,5 +1,5 @@
 import asyncio
-from collections.abc import AsyncIterator
+from collections.abc import AsyncGenerator, AsyncIterator
 from datetime import datetime
 from typing import Any
 from unittest.mock import AsyncMock, Mock, patch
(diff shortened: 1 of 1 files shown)


tbille:
@dpoulopoulos maybe worth checking if there are some coverage missing from the tests

PR #945, opened 2026-03-13
Add GitHub Copilot SDK provider

warnes:
## Description

Adds `copilot_sdk` — a new provider that communicates with GitHub Copilot models via
the [`github-copilot-sdk`](https://pypi.org/project/github-copilot-sdk/) Python package,
which bundles the Copilot CLI binary for your platform and speaks to it over JSON-RPC.

The primary motivation is **reasoning support**: the Copilot CLI session protocol exposes
`reasoning_effort` (`low` / `medium` / `high` / `xhigh`) and streams reasoning deltas as
a distinct event type, which is not available through the GitHub Models OpenAI-compatible
REST endpoint and therefore cannot be surfaced by the existing `openai` provider.

A secondary benefit is **tokenless auth**: if the user is already logged in via
`gh auth login`, no API key or environment variable is needed — the provider falls back to
the CLI's own credentials automatically.

### What's included

| Area | Files changed |
|---|---|
| Provider implementation | `src/any_llm/providers/copilot_sdk/copilot_sdk.py` (new, 463 lines) |
| Package init + factory alias | `src/any_llm/providers/copilot_sdk/__init__.py` (new) |
| Provider enum registration | `src/any_llm/constants.py` (+1 line) |
| Optional dependency group | `pyproject.toml` (`[copilot_sdk]` extras) |
| Unit tests | `tests/unit/providers/test_copilot_sdk_provider.py` (new, 714 lines) |
| Test fixtures | `tests/conftest.py`, `tests/unit/test_provider.py` (+minimal hooks) |
| Provider docs | `docs/src/content/docs/providers.md` (table row + Provider Notes section) |

### Supported features

| Feature | Supported |
|---|---|
| Completion (non-streaming) | ✅ |
| Completion (streaming) | ✅ |
| Reasoning (`reasoning_effort`) | ✅ |
| Image attachments (`data:` URIs) | ✅ |
| Model listing | ✅ |
| PDF attachments | ❌ |
| Embeddings | ❌ |
| Responses API | ❌ |
| Batch | ❌ |

### Authentication

Two modes, checked in order:

1. **Token** — `COPILOT_GITHUB_TOKEN`, `GITHUB_TOKEN`, or `GH_TOKEN` (or `api_key` argument).
2. **Logged-in CLI session** — if no token is found, the provider uses `gh auth login` credentials; no environment variable required.

### Notable implementation details

- **Lazy client init** with `asyncio.Lock` double-checked locking to prevent concurrent coroutines from each spawning a separate CLI process on first use.
- **`SESSION_ERROR` raises `RuntimeError`** rather than silently completing the stream with no content.
- **`mimetypes.init()` at module load** to ensure `guess_extension()` is thread-safe from the first invocation (Python's lazy MIME-type initialisation is not thread-safe).
- **Image attachments** are decoded from `data:` URIs to temporary files on disk, passed to the CLI as `FileAttachment` objects, and cleaned up after each request. HTTP image URLs are not fetched (skipped with a comment).
- **`Copilot_sdkProvider` alias** in `__init__.py` is required so `AnyLLM._create_provider()` can find the class — the factory capitalises the provider key (`copilot_sdk` → `Copilot_sdk`).

### Installation

```bash
pip install any-llm-sdk[copilot_sdk]
```

> `github-copilot-sdk` ships separate wheels per OS/arch (e.g. `macosx_arm64`, `linux_x86_64`).
> `pip` selects the correct wheel automatically on supported platforms.

### Quick usage example

```python
from any_llm import AnyLLM

llm = AnyLLM.create("copilot_sdk")  # uses gh CLI session or COPILOT_GITHUB_TOKEN

# Reasoning
response = llm.completion(
    model="claude-sonnet-4-5",
    messages=[{"role": "user", "content": "Explain async generators in Python."}],
    reasoning_effort="high",
)
print(response.choices[0].message.content)

# Streaming
for chunk in llm.completion(
    model="gpt-4o",
    messages=[{"role": "user", "content": "Hello!"}],
    stream=True,
):
    print(chunk.choices[0].delta.content or "", end="", flush=True)
```

---

## PR Type

- 🆕 New Feature

## Checklist

- [x] **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

- AI Model used: Claude Sonnet 4.6
- AI Developer Tool used: GitHub Copilot (VS Code) and Claude Code (VS Code plugin)
- Any other info: The provider connects to the GitHub Copilot CLI via JSON-RPC, so using Copilot to write a Copilot provider felt appropriate. All logic was reviewed and verified manually; tests were executed locally against a real Copilot CLI session.


njbrake:
Thank you for the contribution! Here is the review from my agent:

```
  Issues Found

  1. BLOCKING — Provider key uses underscore, breaking naming convention

  Every existing provider uses a single lowercase word with no underscores: azureopenai, vertexaianthropic, llamacpp, lmstudio, etc. The factory in _create_provider() does
  provider_key.capitalize() + "Provider", which turns copilot_sdk into Copilot_sdkProvider — an awkward name that requires a class alias in __init__.py.

  Fix: Rename to copilotsdk everywhere (enum value, directory name, class name CopilotsdkProvider, pyproject extras, docs). This eliminates the alias and follows the
  established pattern.

  2. BLOCKING — Decorative section-separator comments

  Both copilot_sdk.py and test_copilot_sdk_provider.py use # ----------- banner comments throughout. From CLAUDE.md/AGENTS.md:

  Do not add decorative section-separator comments (e.g., # ----------- banners). Well-named test functions and natural file ordering are sufficient.

  These should all be removed.

  3. BLOCKING — .github/copilot-instructions.md should be a separate PR

  This file is unrelated to the provider feature. It's a project-level config file that duplicates content from AGENTS.md. Should be split into its own PR.

  4. Medium — Tests will fail when github-copilot-sdk is not installed

  Several tests import from copilot.generated.session_events import SessionEventType inline. If the SDK isn't installed, these tests will fail with ImportError rather than
  being skipped. Tests that depend on the optional SDK should either:
  - Add a module-level skip: pytest.importorskip("copilot")
  - Or guard inline imports with try/except + pytest.skip

  5. Medium — PermissionHandler.approve_all security concern

  _build_session_cfg uses PermissionHandler.approve_all which silently grants any permissions the Copilot CLI session requests (e.g., tool calls). This should at minimum be
  documented as a security consideration, and ideally be configurable or at least logged.

  6. Medium — getattr chain for error extraction violates style guide

  In _stream_from_session (line ~309):
  getattr(getattr(event, "data", None), "message", "Copilot session error")

  CLAUDE.md says: "Prefer direct attribute access (obj.field) over getattr(obj, "field") when the field is typed." If the event types from the SDK have typed fields, use
  event.data.message directly. If they're truly dynamic, at least add a comment explaining why.

  7. Minor — Non-unique completion IDs

  _build_chat_completion uses f"copilot-sdk-{int(time.time())}" which is only second-granularity. Two concurrent non-streaming completions in the same second get the same
  ID. Use time.time_ns() (already used in _build_chunk) or uuid4().

  8. Minor — _make_provider test helper bypasses __init__

  Using object.__new__(CopilotSdkProvider) to construct the provider is fragile — it will silently break if the base class __init__ changes. Consider using patch on
  _init_client and _verify_and_set_api_key instead, which would exercise the real constructor.

  Suggestions (non-blocking)

  - The _messages_to_prompt function flattens multi-turn conversations into a text transcript, which loses the structured turn information that models expect. If the Copilot
   SDK supports structured messages, that would be preferred over prompt flattening.
  - _extract_attachments silently skips HTTP image URLs. Consider raising NotImplementedError or logging a warning so users know their images were dropped.
  - The docstrings are thorough but could be trimmed — the module-level class docstring on CopilotSdkProvider is quite long and repeats the README content.

  Verdict

  Request Changes

  The provider implementation itself is well-structured and thoughtful — the double-checked locking pattern, session lifecycle management, and reasoning support are all done
   well. However, there are three blocking issues:

  1. The copilot_sdk naming convention breaks the established pattern and requires an alias hack. Rename to copilotsdk.
  2. Remove all # ----------- decorative banner comments per project guidelines.
  3. Split .github/copilot-instructions.md into its own PR.

  After those fixes plus the test skip guard (issue #4), this should be ready for another look.

```

njbrake (changes requested):
Needs a few updates, see previous comments 🙏

warnes:
I will fix these issues.

warnes:
Addressed review feedback in commit c20339f.

Summary of changes:
- Renamed provider naming to match convention: `copilot_sdk` -> `copilotsdk`
  - Updated provider key, enum value, module path, docs, tests, and optional-dependency extra.
  - Removed the class alias workaround by using `CopilotsdkProvider` directly.
- Removed decorative section-separator banner comments from provider and provider test module.
- Removed unrelated `.github/copilot-instructions.md` from this PR.
- Added optional dependency test guard in provider tests via `pytest.importorskip("copilot")`.
- Kept defensive access for dynamic SDK event payloads and documented why.
- Switched non-streaming completion IDs to nanosecond precision (`time.time_ns()`).

Validation run:
- `uv run pytest -q tests/unit/providers/test_copilotsdk_provider.py tests/unit/test_provider.py`
- Result: `122 passed, 10 skipped`

If you want, I can also split out any follow-up hardening (e.g., permission policy configurability) into a separate PR.

njbrake:
## Pre-Landing Review: PR #945 — Add GitHub Copilot SDK Provider

*Review by Claude (Opus 4.6), in discussion with @njbrake*

---

**Scope Check: CLEAN** — All files are in scope for adding a new provider. Previous blocking issues (naming convention, decorative comments, out-of-scope file) were all addressed in commit c20339f. Nice work!

---

### 10 issues found (3 high, 7 informational)

#### Pass 1 — HIGH

**1. `copilotsdk.py` — `CompletionParams` silently ignored**
`_acompletion` only forwards `model_id`, `reasoning_effort`, `stream`, and `messages`. Parameters like `temperature`, `max_tokens`, `top_p`, `stop`, `tools`, `tool_choice` are silently dropped. A caller setting `temperature=0.0` gets no indication it was ignored.

Recommended fix: Either forward supported params to the Copilot SDK session config, or raise/warn when non-default values are passed for unsupported params.

**2. `copilotsdk.py:484` — No timeout on `await queue.get()` in streaming**
If the Copilot CLI process crashes silently or the JSON-RPC connection drops without emitting `SESSION_IDLE` or `SESSION_ERROR`, the `while True: item = await queue.get()` loop hangs forever.

Recommended fix: `asyncio.wait_for(queue.get(), timeout=300)` with a configurable timeout constant.

**3. `copilotsdk.py:467` — `asyncio.Queue.put_nowait()` thread-safety concern**
The `on_event` callback uses `queue.put_nowait()` on an `asyncio.Queue`, which is not thread-safe. If the `github-copilot-sdk` delivers events from a background thread (plausible for a subprocess-based SDK), this can corrupt the queue or lose events silently.

Recommended fix: If the SDK fires from threads, use `loop.call_soon_threadsafe(queue.put_nowait, item)`. If it fires from the event loop, add a comment documenting that assumption.

#### Pass 2 — INFORMATIONAL

**4. `copilotsdk.py:401` — Dead attribute `_extra_kwargs`**
`self._extra_kwargs = kwargs` is stored in `_init_client` but never referenced anywhere. Remove it.

**5. `copilotsdk.py:184-218` — `_messages_to_prompt` silently mangles tool messages**
Messages with `role="tool"` or `role="function"` fall through to the `else` branch and are rendered as `"User: ..."`. At minimum log a warning when encountering unsupported message roles.

**6. `copilotsdk.py` — `CopilotClient` never stopped**
`_ensure_client` calls `await self._copilot_client.start()` but there's no corresponding `stop()`/`close()` anywhere. The spawned CLI subprocess leaks on provider GC or process exit. Consider adding `async def close(self)`.

**7. `copilotsdk.py:257` — `_extract_attachments` broad exception swallow**
The `except Exception: pass` catches everything including disk-full, permission-denied, and OOM from decoding large base64 payloads. No size limit on decode either. Consider logging warnings for failures and a size check before decoding.

**8. `copilotsdk.py:272-302` — No `usage` field in responses**
`_build_chat_completion` never populates `usage` (prompt_tokens, completion_tokens, total_tokens). Gateway budget tracking and streaming usage tracking won't work for this provider. If the SDK provides token counts, forward them. Otherwise document the limitation.

**9. `test_copilotsdk_provider.py` — Missing trailing newline**
The file is missing a newline at the end of the file.

**10. `copilotsdk.py:442` — `PermissionHandler.approve_all` security**
Already flagged in previous review and acknowledged for follow-up. Still worth noting for visibility.

---

### Test Coverage

Coverage sits at roughly 76% of code paths, and every covered path has strong assertions. Main gaps are edge cases: stream cancellation, `role="tool"` messages, non-`cli_url` client creation path, and concurrent lock behavior. Solid for a community contribution.

---

### Summary

This is a well-structured provider with thoughtful implementation details (double-checked locking, session lifecycle, reasoning support). The two most impactful items to address before merge are **silent parameter dropping (#1)** and **streaming timeout (#2)**, as these directly affect users in production. The rest could be addressed in follow-up PRs.

njbrake:
Hey @warnes! This is Claude (posting on behalf of @njbrake).

I reviewed your fix commit `bf2f687` on the `copilot/address-issues-upstream-pr-945` branch — nice work, it addresses all 9 items from the pre-landing review cleanly. Here's a quick summary:

| Issue | Status |
|-------|--------|
| #1 Silent param drop | ✅ `_warn_unsupported_params()` with `UserWarning` |
| #2 Stream timeout | ✅ `asyncio.wait_for()` with 300s default |
| #3 Queue thread-safety | ✅ `loop.call_soon_threadsafe()` |
| #4 Dead `_extra_kwargs` | ✅ Removed |
| #5 Tool message mangling | ✅ Explicit `user` branch + warning for unknown roles |
| #6 Client never stopped | ✅ `async def close()` added |
| #7 Broad exception swallow | ✅ `logger.warning()` + 20 MiB size guard |
| #8 No usage field | ✅ Documented limitation in comment |
| #9 Missing trailing newline | ✅ Fixed |

**One nit to fix before merging:** On line 341 of `copilotsdk.py`, `asyncio.get_event_loop()` should be `asyncio.get_running_loop()`. Since `_stream_from_session` is an async function, a running loop is always available, and `get_event_loop()` is deprecated in Python 3.12+ when called without a running loop.

Once that's fixed, could you merge this branch into your PR's `main` branch so the PR picks up the changes? After that + the one-liner fix, this should be ready for final review.

warnes:
> **One nit to fix before merging:** On line 341 of `copilotsdk.py`, `asyncio.get_event_loop()` should be `asyncio.get_running_loop()`. Since `_stream_from_session` is an async function, a running loop is always available, and `get_event_loop()` is deprecated in Python 3.12+ when called without a running loop.

Fixed.  

Merge conflict also resolved.

njbrake:
## Manual Review — @njbrake (posted and drafted by Claude)

Hey @warnes, thanks for the contribution and for being so responsive through multiple rounds of feedback! The automated agent review passed, so these are follow-up questions from my manual review before we merge.

### 1. Docs: `## Provider Notes` section in `providers.md`

The provider table is auto-generated and every other provider is documented with just a table row. This PR adds ~60 lines of Installation / Authentication / Configuration / Usage docs that read more like the SDK's own README. It also restructures the existing Gemini/VertexAI inline-media note under a new heading (done during merge conflict resolution).

Should this content live somewhere else (e.g. the SDK's own docs), or be trimmed down to a brief note similar to the existing Gemini one-liner?

### 2. Auth: `GITHUB_TOKEN` / `GH_TOKEN` fallback in `_verify_and_set_api_key`

These env vars are set automatically by GitHub Actions and are commonly present in CI environments for unrelated purposes. A user who instantiates this provider in CI could silently pick up a token that lacks Copilot scopes, leading to confusing auth errors downstream.

Should the fallback be limited to `COPILOT_GITHUB_TOKEN` only, with the CLI session as the other auth path?

### 3. Structure: No `utils.py`

21 of 40 providers split their helper/conversion functions into a separate `utils.py`. The copilotsdk provider has ~150 lines of module-level helpers (`_messages_to_prompt`, `_extract_attachments`, `_build_chat_completion`, `_build_chunk`, `_copilot_model_to_openai`, `_cleanup_temp_files`) that are exactly the kind of type conversion and request/response processing functions other providers put in utils.

Should these be moved to a `utils.py` to follow the established pattern?

warnes:
## Changes addressing manual review feedback

All three items from the [review comment](https://github.com/mozilla-ai/any-llm/pull/945#issuecomment-4193660825) have been resolved in commit `6d690d0`.

### 1. `providers.md` — trimmed to a one-liner

The ~60-line Provider Notes block (Installation / Authentication / Configuration / Usage subsections) has been replaced with a single paragraph matching the style of the existing Gemini note.

### 2. Auth limited to `COPILOT_GITHUB_TOKEN` only

`_verify_and_set_api_key` now checks only `COPILOT_GITHUB_TOKEN` (or an explicit `api_key` argument). The `GITHUB_TOKEN` and `GH_TOKEN` fallbacks have been removed — those variables are automatically injected by GitHub Actions and may carry tokens that lack Copilot scopes, making them an unreliable fallback.

Five new tests explicitly cover the auth logic:
- `test_api_key_explicit_wins`
- `test_api_key_falls_back_to_copilot_env`
- `test_github_token_env_var_is_not_used` — asserts `None` when only `GITHUB_TOKEN` is set
- `test_gh_token_env_var_is_not_used` — asserts `None` when only `GH_TOKEN` is set
- `test_api_key_returns_none_when_copilot_token_absent`

### 3. Helpers extracted to `utils.py`

Six module-level helper functions have been moved from `copilotsdk.py` into a new `src/any_llm/providers/copilotsdk/utils.py`, following the pattern used by the other 21 providers:

- `_messages_to_prompt`
- `_extract_attachments`
- `_cleanup_temp_files`
- `_build_chat_completion`
- `_build_chunk`
- `_copilot_model_to_openai`

`copilotsdk.py` imports them from `.utils` inside the existing `try/except ImportError` guard.

### Additional fixes

- **Test isolation**: two `caplog`-based log-capture tests were silently order-dependent — `any_llm.logging.setup_logger()` sets `propagate=False` on the `any_llm` logger at import time, so records from `utils.py` never reached pytest's root-logger capture handler in full-suite runs. Replaced with `patch.object(utils_module.logger, "warning")` for deterministic behaviour.
- **Ruff**: all issues introduced by the refactor resolved — `TC002`, `UP037` ×13, `UP041`, `B904`/`TRY003`/`EM102`, `EM101` ×6, `I001`, stale `RUF100` directives, `RUF001` (ambiguous `×`), `RUF059`.

All 970 unit tests pass (31 skipped, gateway tests excluded due to missing optional deps in the local venv).

njbrake:
Hi @warnes I'm so sorry, I think I was so caught up in the code accuracy of the PR that I hadn't taken a step back to think of the bigger picture (Putting my "maintainer" hat on instead of my "developer" hat). 

Can you explain a bit more about the motivation for adding copilot as a provider, as opposed to Github as a provider? 

The copilot SDK bundles a ~140MB binary file with it as a dependency, and it's an agentic framework more than a direct LLM provider, which is slightly divergent from the purpose of the current list of providers that we support. We're trying to shoehorn sessions, permissions, tool calls into a completion interface. 

I think that we may need to close this PR and we can consider adding a github provider to support the  https://models.github.ai/inference models that are available as OpenAI compatible endpoints, which is more in line with the any-llm design patterns.

Again, so sorry for the back and forth on this PR before I thought about this, I was having my agent review the code to make sure it was functional before stepping back to think about whether the PR would be a good fit for the project.

warnes:
Please reopen.

warnes:
The primary reason for adding the copilot SDK was to allow my application to query the available models, so that the user could select the model... and of course, so I could use my Copilot sessions.

njbrake:
Hey @warnes, thanks for the reply and sorry for the delayed response.

The code quality of this PR was solid and you were very responsive through multiple rounds of review. I appreciate the effort.

That said, I've decided not to reopen this PR. The core issue isn't code quality, it's architectural fit:

- The `github-copilot-sdk` bundles a ~140MB platform-specific binary and communicates over JSON-RPC to a subprocess. Every other provider in any-llm is a lightweight HTTP client. This is a qualitatively different dependency class that we don't want to set a precedent for.
- The Copilot SDK's session protocol doesn't map cleanly to our completion interface. Parameters like `temperature`, `max_tokens`, `top_p`, `tools` have to be warned/ignored, messages get flattened to text, and usage tracking isn't available. These aren't bugs, they're fundamental mismatches.

For your use cases (model listing + inference), the **GitHub Models API** (`models.github.ai/inference`) provides OpenAI-compatible endpoints that would be a much better fit. I'm going to open an issue for adding a `github` provider that uses those endpoints. It would give you model listing and completion support without the heavy binary dependency, and it would slot in naturally alongside our other providers.

Sorry for not catching this sooner. I should have evaluated the architectural fit before diving into code review. I hope this doesn't discourage you from contributing in the future.

*Posted by Claude (Opus 4.6) via discussion with @njbrake*

warnes:
Thanks for taking the time to consider this.  It does sound like the GitHub Models API is a better match.

I don't have time at the moment to tackle it, but I'll be glad to test it out.

I've subscribed to the new issue so I'll get informed.

PR #946, opened 2026-03-13
fix(platform): prevent trace export failures from crashing consumers

njbrake:
## Description

`PlatformProvider._stream_with_usage_tracking` calls `export_completion_trace` after all chunks have been yielded. If that call raises (e.g. platform API returns 404, network error), the broad `except Exception` block re-raises the error, which propagates back through the consumer's `async for chunk in response:` loop and crashes their application even though all content was already delivered.

The same issue existed in the non-streaming `_acompletion` path: a successful completion result was lost if trace export failed afterward.

Both paths now wrap `export_completion_trace` in a dedicated `try/except` that logs the error and marks the span as failed, without re-raising. Actual LLM/streaming errors still propagate normally.

## PR Type

- 🐛 Bug Fix

## Relevant issues

N/A (reported by a platform user whose app crashed with `HTTPStatusError: Client error '404 Not Found' for url '.../api/v1/usage-events/'` during streaming)

## Checklist
- **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

- Any other info you'd like to share: Used to investigate the bug across SDK, gateway, and platform client code, then implement the fix and tests.

### DIFF
--- a/src/any_llm/providers/platform/platform.py
+++ b/src/any_llm/providers/platform/platform.py
@@ -232,28 +232,22 @@ class PlatformProvider(AnyLLM):
             if not params.stream:
                 end_time_ns = time.time_ns()
 
-                try:
-                    await export_completion_trace(
-                        platform_client=self.platform_client,
-                        client=self.client,
-                        any_llm_key=any_llm_key,
-                        provider=self.provider.PROVIDER_NAME,
-                        request_model=params.model_id,
-                        completion=cast("ChatCompletion", completion),
-                        start_time_ns=start_time_ns,
-                        end_time_ns=end_time_ns,
-                        client_name=self.client_name,
-                        session_label=session_trace_label,
-                        user_session_label=user_session_label,
-                        conversation_id=params.user,
-                        access_token=access_token,
-                        existing_span=llm_span,
-                    )
-                except Exception as trace_err:
-                    logger.error("Failed to export completion trace: %s", trace_err)
-                    llm_span.set_attribute("error.type", trace_err.__class__.__name__)
-                    llm_span.set_status(Status(StatusCode.ERROR, "trace export failed"))
-                    llm_span.end(end_time=end_time_ns)
+                await export_completion_trace(
+                    platform_client=self.platform_client,
+                    client=self.client,
+                    any_llm_key=any_llm_key,
+                    provider=self.provider.PROVIDER_NAME,
+                    request_model=params.model_id,
+                    completion=cast("ChatCompletion", completion),
+                    start_time_ns=start_time_ns,
+                    end_time_ns=end_time_ns,
+                    client_name=self.client_name,
+                    session_label=session_trace_label,
+                    user_session_label=user_session_label,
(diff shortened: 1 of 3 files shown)


njbrake:
@dpoulopoulos or @tbille open your thoughts about this. On one hand, I get that we don't want to risk having logging events be silently lost, but in my case, there was an API change that made my client call hit an invalid end point which then was causing my application to crash. I'm thinking we probably don't want to risk some issues with logging causing a clients application to not work. Maybe we just log an error message but allow things to continue?

src/any_llm/providers/platform/platform.py
```diff
@@ -232,22 +232,28 @@ async def _acompletion(
...
+                        user_session_label=user_session_label,
+                        conversation_id=params.user,
+                        access_token=access_token,
+                        existing_span=llm_span,
+                    )
+                except Exception as trace_err:
+                    logger.error("Failed to export completion trace: %s", trace_err)
+                    llm_span.set_attribute("error.type", trace_err.__class__.__name__)
+                    llm_span.set_status(Status(StatusCode.ERROR, "trace export failed"))
+                    llm_span.end(end_time=end_time_ns)
```
tbille: Why setting the span attributes if we could export?

src/any_llm/providers/platform/platform.py
```diff
@@ -232,22 +232,28 @@ async def _acompletion(
...
+                        start_time_ns=start_time_ns,
+                        end_time_ns=end_time_ns,
+                        client_name=self.client_name,
+                        session_label=session_trace_label,
+                        user_session_label=user_session_label,
+                        conversation_id=params.user,
+                        access_token=access_token,
+                        existing_span=llm_span,
+                    )
+                except Exception as trace_err:
```
tbille: Maybe it'd be simpler to handle the Excecption in the function directly.

src/any_llm/providers/platform/platform.py
```diff
@@ -465,28 +471,36 @@ async def _stream_with_usage_tracking(
...
+                        user_session_label=user_session_label,
+                        conversation_id=conversation_id,
+                        access_token=access_token,
+                        existing_span=llm_span,
+                    )
+                except Exception as trace_err:
+                    logger.error("Failed to export completion trace after streaming: %s", trace_err)
+                    llm_span.set_attribute("error.type", trace_err.__class__.__name__)
+                    llm_span.set_status(Status(StatusCode.ERROR, "trace export failed"))
+                    llm_span.end(end_time=end_time_ns)
```
tbille: Same questions than in the previous one

tbille:
@njbrake  yep this is because we made a breaking change in the API, we moved to `/v1/traces` endpoint. I agree we should resolve those changes more "quietly" on the sdk

src/any_llm/providers/platform/platform.py
njbrake: Perfect, I prefer that design too, much simpler 🙏

PR #980, opened 2026-03-31
refactor(gateway): simplify internals with layered architecture

tbille:
## Description
Refactors the gateway into a simpler layered architecture with thin HTTP routes and clear separation between API dependencies, services, repositories, models, and core infrastructure. Legacy route/wrapper modules were removed and tests were migrated to the new import paths.

Added support for sqlite for simpler local dev.

### Examples
- Start server: `gateway serve --config config.yml`
- Use OpenAI client with gateway key:
  ```python
  from openai import OpenAI

  client = OpenAI(api_key="YOUR_GATEWAY_KEY", base_url="http://0.0.0.0:8000/v1")
  print(client.chat.completions.create(model="openai:gpt-4o", messages=[{"role": "user", "content": "Hello"}]).choices[0].message.content)
  ```
- List models:
  ```bash
  curl -H "X-AnyLLM-Key: Bearer $GATEWAY_API_KEY" http://localhost:8000/v1/models
  ```

## PR Type
- 💅 Refactor
- 🐛 Bug Fix

## Relevant issues
N/A

## Checklist
- [x] **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information
- AI Model used: gpt-5.3-codex
- AI Developer Tool used: OpenCode CLI
- Any other info you'd like to share: Ran `uv run pytest -q tests/gateway tests/unit/test_gateway_*.py` (257 passed, 9 skipped, 2 warnings). Provider-dependent tests are skipped when `GEMINI_API_KEY`/`GOOGLE_API_KEY` is not set.

### DIFF
--- a/src/any_llm/gateway/main.py
+++ b/src/any_llm/gateway/main.py
@@ -72,8 +72,7 @@ _ROOT_TUTORIAL_HTML = """<!doctype html>
     <main>
       <h1>AI Gateway (Proxy Server)</h1>
       <div class="sub">
-        <a class="link" href="https://mozilla-ai.github.io/any-llm/gateway/quickstart/">Getting Started - E2E Tutorial</a>
-        <span class="note"> - Setup virtual keys, make your first request</span>
+        <a class="link" href="https://mozilla-ai.github.io/any-llm/gateway/quickstart/">Gateway Quickstart</a>
       </div>
 
       <p class="note">On first startup, the gateway prints a bootstrap API key in logs. Use that value as <code>YOUR_GATEWAY_KEY</code>.</p>
--- a/tests/unit/test_gateway_root_page.py
+++ b/tests/unit/test_gateway_root_page.py
@@ -17,7 +17,7 @@ def test_root_tutorial_page_is_available(tmp_path: Path) -> None:
     assert response.status_code == 200
     assert "text/html" in response.headers["content-type"]
     assert "AI Gateway (Proxy Server)" in response.text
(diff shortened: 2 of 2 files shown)


njbrake:
*Posted by Claude after discussion with @njbrake*

---

Thanks for the PR! The layered architecture is a solid improvement and the bootstrap key + tutorial page are nice UX additions for first-time users.

One thing to address:

**SQLite + `FOR UPDATE` incompatibility:**
Switching the default to SQLite is great for local dev, but `db/helpers.py:get_active_user()` uses `query.with_for_update()` which SQLite doesn't support. Any budget-enforced request will crash with a `StatementError` on the new default database. This is pre-existing code but surfaces now because SQLite is the default. Could you add a dialect check to skip `with_for_update()` for SQLite? Something like:

```python
if db.bind and db.bind.dialect.name != "sqlite":
    query = query.with_for_update()
```

Happy to handle this in a follow-up if you'd prefer.

**Minor notes (non-blocking):**
- `exceptions/domain.py` defines 4 exception classes that aren't used yet in this PR. Fine if they're scaffolding for follow-up work, just want to confirm that's intentional.
- The root tutorial page font sizes are quite large (36px for body text, 28px for code). Was this intentional or optimized for a demo/presentation?
- The bootstrap key gets logged during OpenAPI spec generation in CI. Doesn't break anything now that the spec matches, but passing `bootstrap_api_key=False` in `generate_openapi.py` would be cleaner.

tbille:
comments fixed @njbrake

PR #1001, opened 2026-04-05
feat(asyncio): Make all calls to postgress using sqlalchemy[asyncio]

### LINKED ISSUE #1000: bug: Gateway locks up under concurrent requests: `validate_user_budget` holds `SELECT FOR UPDATE` across the entire LLM call
## Summary

Under concurrent traffic to the same user, the gateway becomes completely unresponsive — not just slow on `/v1/messages` / `/v1/chat/completions` / `/v1/embeddings`, but also on `/health`, which does not touch the database. Once the freeze begins it does not recover until the process is restarted.

The root cause is that `validate_user_budget` acquires a row-level `SELECT ... FOR UPDATE` on the user row at the **start** of every request and the transaction is then held open through the entire downstream LLM call (including long-running streaming responses). With a single-worker async event loop, a second contending request to the same user blocks on the Postgres `transactionid` lock synchronously, which freezes the event loop and takes the whole gateway offline.

### Reproducer

Two concurrent requests for the same user, where the first issues a long-running streaming LLM call:

```bash
# terminal 1: long stream holds the user row lock
curl -N -X POST http://localhost:4000/v1/messages \
  -H "X-AnyLLM-Key: Bearer $KEY" \
  -H "Content-Type: application/json" \
  -d '{"model": "anthropic:claude-haiku-4-5", "max_tokens": 1024, "stream": true,
       "messages": [{"role": "user", "content": "write a long essay"}]}'

# terminal 2: any second request for the same user, hangs
curl -X POST http://localhost:4000/v1/chat/completions \
  -H "X-AnyLLM-Key: Bearer $KEY" \
  -H "Content-Type: application/json" \
  -d '{"model": "openai:gpt-4o", "messages": [{"role": "user", "content": "Hello!"}]}'

# terminal 3: unrelated endpoint, also hangs
curl http://localhost:4000/health
```

### Observed state in Postgres

```sql
SELECT pid, state, wait_event_type, wait_event, now() - xact_start AS dur
FROM pg_stat_activity WHERE datname = 'anyllm' AND state <> 'idle';

 pid  |        state        | wait_event_type |  wait_event   |      dur
------+---------------------+-----------------+---------------+----------------
  672 | idle in transaction | Client          | ClientRead    | 00:01:45.2697
 1153 | active              | Lock            | transactionid | 00:01:45.2416
```

- `pid 672` holds a `SELECT FOR UPDATE` on the user row and is `idle in transaction` with `wait_event=ClientRead` — Postgres is waiting for the Python client to send the next SQL command. The client is blocked `await`-ing the LLM stream.
- `pid 1153` is blocked on `wait_event_type=Lock, wait_event=transactionid` — the canonical fingerprint of row-lock contention.
- `pg_blocking_pids(1153)` → `[672]`.

### Narrowed root cause

The freeze is driven by **three compounding factors**:

1. **Lock held across LLM call.** `validate_user_budget` (`src/any_llm/gateway/services/budget_service.py:82`) calls `get_active_user(..., for_update=True)`. That opens a transaction and takes a row-level lock. The transaction is **only committed inside `reset_user_budget`** (`budget_service.py:60`), which runs conditionally (`if user.next_budget_reset_at and now >= user.next_budget_reset_at`). In the common case (no reset due), the function returns with the transaction still open — and the row lock is then held for the entire duration of the downstream `await amessages(...)` / `await acompletion(...)` call. On a streaming response, that's potentially the full duration of the stream.

2. **Sync DB in async handler.** The gateway uses the synchronous SQLAlchemy + psycopg2 stack (`sqlalchemy.orm.Session`, `create_engine`), but every route is `async def`. The route calls `validate_user_budget` which issues a blocking `SELECT FOR UPDATE`. When that query has to wait for another transaction to release the lock, the sync psycopg2 call blocks **inside an `async` coroutine** — which blocks the single asyncio event loop until the lock is granted.

3. **Single worker.** The gateway runs `--workers 1` (default in `cli.py:35`). With one event loop, a single blocked sync DB call inside an async handler freezes **every** in-flight request, including endpoints that don't touch the DB (`/health`, `/health/liveness`). This is why the freeze looks like a total process hang rather than "some requests slow".

### Why the lock was taken in the first place

The `FOR UPDATE` exists to serialize one specific write: the conditional budget reset inside `reset_user_budget`, which inserts a `BudgetResetLog` row and zeroes `users.spend`. Without the lock, two concurrent requests arriving exactly at the reset boundary could both perform the reset and both insert a log entry.

Importantly, the `FOR UPDATE` is **not** needed for:
- The `blocked` / `deleted_at` / `budget_id` read — these are ordinary stale-tolerant checks.
- The `user.spend >= budget.max_budget` check — at worst, one extra request slips through during contention, which is acceptable.
- Spend accumulation on the response path — `log_usage` already uses an atomic `UPDATE users SET spend = spend + :cost` (`chat.py:160`, `embeddings.py:119`), which is safe without a row lock.

So a multi-second (or longer) row lock is being held across the LLM call to protect a single write that happens at most once per budget period per user.

## Expected Behavior

- The validation path for `/v1/messages` / `/v1/chat/completions` / `/v1/embeddings` should **not** hold a row-level lock while the LLM call is in flight.
- Concurrent requests for the same user should be able to execute in parallel as long as no budget reset is actually due.
- The race window for `reset_user_budget` must still be serialized so we don't produce duplicate `BudgetResetLog` rows.
- `/health` (and any endpoint that does not depend on the user row) must keep responding even when the DB is contended.

### Fix 1: Split read from write in `validate_user_budget` (primary fix)

Refactor so the lock is only taken when a budget reset is actually due.

- Do an **unlocked** `SELECT` of the user row. `db.commit()` immediately to release the autobegun read transaction.
- 404 / 403 early-outs happen with no lock ever taken.
- If the user has no `budget_id`, return immediately — no lock.
- Read the `Budget` row (still unlocked) and commit.
- **Only if `now >= next_budget_reset_at`**, re-fetch the user with `FOR UPDATE`, double-check the reset is still due (another worker may have already reset it), perform the reset, and commit. The lock is scoped strictly to the reset write.
- Perform the `spend >= max_budget` check using the post-reset values and return.

Properties:
- No lock held across the LLM call.
- No lock for users without a budget.
- No lock for users with a budget but not due for reset.
- Reset race is safe: loser acquires the lock after the winner has committed, sees `next_budget_reset_at` is already in the future, commits, and no-ops.

### Fix 2: Add a `GATEWAY_ENABLE_BUDGET_ENFORCEMENT` kill switch

Add a config flag (default `True`, on by default) that, when disabled, skips the entire `validate_user_budget` call in the `/v1/chat/completions`, `/v1/messages`, and `/v1/embeddings` routes. The gate is a simple `if` at each call site — no flag plumbed into `validate_user_budget` itself.

When disabled:
- `validate_user_budget` is not called — no user-row read, no blocked check, no budget check, no `FOR UPDATE`, ever.
- `log_usage` continues to accumulate `users.spend` via atomic update, so cost tracking still works.

This gives operators an escape hatch for deployments that enforce budgets (and user blocking) out-of-band — e.g. a billing job reading `users.spend` directly, or an upstream gateway handling user validation — and is the safest setting for very high-concurrency deployments until the async-driver follow-up (below) lands.

### Follow-up: switch to async SQLAlchemy

The architectural root cause of the event-loop freeze is that the gateway uses sync `psycopg2` + `sqlalchemy.orm.Session` inside `async def` route handlers. A contended blocking call freezes the event loop. Converting the DB layer to `create_async_engine` + `AsyncSession` + `asyncpg` (or `psycopg[async]`) would mean a contended query yields to the event loop instead of blocking it — so `/health` and other users' requests continue to serve even during contention.

## Acceptance Criteria

- `validate_user_budget` does not acquire `FOR UPDATE` for users with no budget, or for users whose budget is not yet due for reset.
- Concurrent requests for the same user no longer serialize on the user row while an LLM call is in flight.
- The `BudgetResetLog` insert is still protected against duplicate rows under concurrent reset.
- `GATEWAY_ENABLE_BUDGET_ENFORCEMENT=false` disables budget checks and reset logic entirely; user existence and blocked-status checks remain active.
- The default behavior is unchanged (budget enforcement on).
- All existing budget tests still pass (`test_budget_reset.py`, `test_budget_race_condition.py`, `test_atomic_spend_update.py`).
- Documentation for disabling budget enforcement is added to `docs/gateway/budget-management.md`.

brightsparc:
## Description

Fixes the gateway lockup diagnosed in #1000 by removing the sync-psycopg2-in-async-handler anti-pattern at its root, and adds two pluggable strategies for common bottlenecks on the request path.

Four stacked changes, each behind a config flag:

1. **Async DB layer** — swap `psycopg2` + `sqlalchemy.orm.Session` for `asyncpg` + `AsyncSession`. Contended DB calls now yield to the event loop instead of blocking it.
2. **`budget_strategy` config** — `for_update` (default, legacy) | `cas` (lock-free, recommended) | `disabled` (escape hatch)
3. **`log_writer_strategy` config** — `single` (default, inline) | `batch` (recommended for high-throughput)
4. **DB pool tuning** — `db_pool_size` / `db_max_overflow` / `db_pool_timeout` / `db_pool_recycle` exposed to config, defaults bumped for async workloads.

Plus a k6-based load-test harness under `tests/load/` with recorded comparison results.

### Why

The gateway uses `async def` route handlers but (until this PR) every DB call underneath went through synchronous `psycopg2`. When a query had to wait on anything — a `FOR UPDATE` row lock, a saturated pool, Postgres itself — the sync call blocked the **entire event loop**, freezing every in-flight request including `/health`. The load test on `main` demonstrates this clearly: under 100 VUs the gateway stalls at 228 requests completed (all on the warmup), the 15-connection pool saturates, and the process needs to be killed after piling up 30s-timeout errors.

After this change, a contended DB call `await`s and other coroutines make progress. Fixes the architectural root cause rather than papering over it.

### Changes

**Dependencies** (`pyproject.toml`)
- Added `asyncpg>=0.29.0` (async Postgres driver)
- Added `aiosqlite>=0.19.0` (async SQLite for local dev/tests)
- Added `sqlalchemy[asyncio]` extra
- Kept `psycopg2-binary` for Alembic (which remains sync)

**Database core** (`core/database.py`)
- `create_engine` → `create_async_engine` with configurable pool (pool_size, max_overflow, pool_timeout, pool_recycle)
- `sessionmaker` → `async_sessionmaker(expire_on_commit=False)`
- `get_db` is now an `async` generator yielding `AsyncSession`
- URL auto-translation: accepts familiar `postgresql://…` / `sqlite:///…` and wires the async driver automatically; Alembic gets the sync form
- `create_session()` helper for startup-time use outside request scope

**Config** (`core/config.py`)
- `budget_strategy` (default `"for_update"`) — how per-user budget validation serializes concurrent resets
- `log_writer_strategy` (default `"single"`) — how usage log rows are persisted
- `db_pool_size` (default 10), `db_max_overflow` (default 20), `db_pool_timeout` (default 30.0), `db_pool_recycle` (default -1)

**Budget service** (`services/budget_service.py`) — dispatch via `match/case`:
- `for_update`: legacy behavior (FOR UPDATE across entire request)
- `cas`: lock-free — atomic conditional UPDATE (`WHERE next_budget_reset_at < now`) claims the reset. Compare-and-swap semantics protect the `BudgetResetLog` insert from duplicates.
- `disabled`: skip `validate_user_budget` entirely, return None

**Log writers** (`services/log_writer.py`) — new module with Protocol + 2 implementations:
- `SingleLogWriter`: write each event inline, one txn per event
- `BatchLogWriter`: queue events onto `asyncio.Queue`, background task flushes batches of up to 100 rows / 1s. `queue.join()` drains on shutdown via the FastAPI lifespan.
- Both expose `start()` / `stop()` and are driven from the lifespan. Both treat flush failures as best-effort (rollback + log + drop), consistent with the historical contract.

**Startup** (`main.py`)
- Bootstrap + pricing init + log writer lifecycle run inside a FastAPI `lifespan` async context manager
- Keeps all async-engine operations bound to the server's event loop (avoids the event-loop-binding pitfall asyncpg has when connections are created in one loop and reused in another)

**Auth / deps** (`api/deps.py`)
- `_verify_and_update_api_key`, `verify_api_key`, `verify_api_key_or_master_key` async end-to-end
- New `get_log_writer` FastAPI dependency

**Repositories + services + routes** (~20 files)
- All DB calls converted: `db.query(X).filter(...).first()` → `(await db.execute(select(X).where(...))).scalar_one_or_none()`, `db.commit/rollback/refresh/delete` → `await db.commit/rollback/refresh/delete`, atomic spend update → `await db.execute(update(User)...values(spend=User.spend + cost))`
- `log_usage` in `chat.py` refactored to build a `UsageLog` and enqueue via `LogWriter.put()` — no longer commits inline

**Metrics** (`metrics.py`)
- `gateway_usage_log_queue_depth` (gauge)
- `gateway_usage_log_batch_size` (histogram, labeled by writer)
- `gateway_usage_log_flush_duration_seconds` (histogram, labeled by writer + result)
- `gateway_usage_log_rows` (counter, labeled by writer + result=written|dropped)

**Tests** (`tests/gateway/conftest.py` + ~10 test files)
- Async fixtures: `test_db`, `db_session`, `client.override_get_db` all yield `AsyncSession`
- New tests for `cas` and `disabled` budget strategies (no-FOR-UPDATE assertions via `get_active_user` spy)

**Load test harness** (`tests/load/`)
- `fake_provider.py` — click CLI, seedable log-normal jittered delay, zero-delay noop mode
- `load_test.js` — k6 scenarios: warmup, distinct_users, same_user with `setup()`/`teardown()` lifecycle
- `run_load_test.sh` — orchestrator (Postgres container, fake provider, gateway, k6) + post-run DB verification that counts `usage_logs` rows to validate 100% coverage
- `results/` — recorded sync-vs-async comparison across all 4 budget strategies + the batch log writer

### Benchmark results

100 VUs × 30s × noop fake provider, single uvicorn worker:

| Scenario | Total rps | distinct rps | same rps | Reqs OK | Coverage |
|---|---|---|---|---|---|
| sync (main) | ~0 (stalled) | ~0 | ~1 | 391 | — |
| async + for_update | 81.1 | 46.3 | 34.8 ⚠️ | 6,578 | — |
| async + cas | 82.3 | 41.5 | 40.8 | 6,498 | — |
| async + disabled | 89.5 | 44.8 | 44.7 | 7,113 | — |
| **async + disabled + batch** 🏆 | **97.4** | 47.4 | **50.0** ✨ | **7,637** | **100%** |

Key findings:
- **sync → async**: 391 → 6,578 successful requests (17×), zero failures vs 160 pool-exhaustion timeouts
- **for_update → cas**: closes the 25% `same_user` contention cliff (34.8 → 40.8 req/sec)
- **single → batch log writer**: +9% throughput, 100% row coverage on shutdown (7,637 requests → 7,637 rows persisted)
- **`same_user` beats `distinct_users`** in the batch run because batch groups spend UPDATEs per user

Full results and per-transition analysis: `tests/load/results/results.md`.

### Not in scope

- **Alembic migrations remain sync** — they use the original `postgresql://` URL driven by `psycopg2-binary`.
- **Default budget strategy stays `for_update`** for backwards compatibility. Users opt into `cas` via config.
- **Default log writer stays `single`** for backwards compatibility. Users opt into `batch` via config.

### Possible follow-ups

- Drop `psycopg2-binary` by switching Alembic to `psycopg[binary]` (psycopg3 sync mode). Consolidates to one driver family.
- Bump default `--workers` to match production deployment (currently 1 for benchmarking signal isolation).
- Module-level pricing cache — `find_model_pricing` is a PK lookup per request.

## PR Type

- 🐛 Bug Fix (lockup under concurrent load, #1000)
- 🔧 Refactor / Architecture (sync→async DB layer)
- ✨ Enhancement (pluggable budget + log writer strategies, DB pool tuning)

## Relevant issues

Fixes #1000.

## Checklist

- [x] **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

- Any other info you'd like to share: AI did the mechanical sync→async conversion across ~20 source files and ~10 test files, designed the `cas` strategy, built the log writer abstraction, and wrote the k6 load test harness. The lockup diagnosis was done live in a running container by inspecting `pg_stat_activity` / `pg_blocking_pids`. All benchmarks and results are reproducible via `./tests/load/run_load_test.sh`.


brightsparc:
@njbrake sorry for all the PR's but if you could prioritise reviewing this one, as it unblocks us from serving more than a limited number of users.  Includes reproducible load testing with [k6](https://k6.io/)

njbrake:
@brightsparc sorry for the slow reply! Working on it

brightsparc:
I’ve updated this from the base. Let me know if there are any more changes you would like to see before we can land @njbrake

Issue #1027, opened 2026-04-14
[BUG]metadata.user_id parsing error causes user lookup failure


Xibao-Lv:
**`any-llm` version:** main branch
**Python Version**:3.11+
**claude-code:2.1.98 (Claude Code)**

### Describe the bug

When request.metadata.get("user_id") returns a dictionary object instead of a string, the str() conversion in messages.py line 75 produces a JSON string (e.g., '{"device_id":"c7a9f9...","account_uuid":"","session_id":"2f614ef6-..."}'). This
  invalid string is passed as user_id to validate_user_budget, causing get_active_user to return None and raising an HTTP 404 exception ("User '...' not found").

### Example w/ Code

Triggering request:
```curl
  POST /v1/messages
  Authorization: Bearer <master_key>
  {
    "model": "anthropic/claude-3-opus",
    "messages": [...],
    "max_tokens": 100,
    "metadata": {
      "user_id": {
        "device_id": "c7a9f910a6b876ceca80e42de13bb085f05a81bfa4d77ede4104459e7776a3de",
        "account_uuid": "",
        "session_id": "2f614ef6-0f8d-413d-ba56-3dfe4b02fa88"
      }
    }
  }
```

Error response:
```
{"detail":"User '{'device_id': 'c7a9f910a6b876ceca80e42de13bb085f05a81bfa4d77ede4104459e7776a3de', 'account_uuid': '', 'session_id': '2f614ef6-0f8d-413d-ba56-3dfe4b02fa88'}' not found"}
```
### Affected code flow

  1. messages.py:75 → str(dict) = JSON string
  2. _helpers.py:11-51 → resolve_user_id() returns the invalid string
  3. budget_service.py:82-88 → validate_user_budget() calls get_active_user()
  4. users_repository.py:18 → No matching user found → returns None
  5. budget_service.py:85-88 → Raises HTTP 404

### Root cause location
```
  # src/any_llm/gateway/api/routes/messages.py:75
  user_from_metadata = request.metadata.get("user_id") if request.metadata else None
  user_id = resolve_user_id(
      user_id_from_request=str(user_from_metadata) if user_from_metadata else None,  # ← Problem here
      ...
  )
```

### Steps to Reproduce

  1. Send a POST request to /v1/messages with master key authentication
  2. Include a metadata field where user_id is a dictionary object (not a string)
  3. Observe the 404 error response

[...]

PR #1029, opened 2026-04-14
fix(gateway): handle dict user_id in Anthropic messages metadata

xyaz1313:
## Problem

When a client sends `metadata.user_id` as a dict (e.g. `{"device_id": "...", "session_id": "..."}`), `str(dict)` produces a Python repr string like `"{'device_id': '...', 'session_id': '...'}"`. This string never matches any user in the DB, causing a silent HTTP 404.

Reported in #1027.

## Root Cause

In `messages.py:75`:
```python
user_id_from_request=str(user_from_metadata) if user_from_metadata else None
```

When `user_from_metadata` is a dict, `str()` produces a Python repr, not a valid user ID.

## Fix

Added `_normalize_user_id()` helper that:
1. Returns strings as-is (happy path)
2. For dicts, tries to extract a known identifier field (`user_id`, `id`, `account_uuid`, `device_id`)
3. Falls back to deterministic JSON serialization if no known field found
4. Handles None gracefully

Added 2 tests:
- `test_messages_endpoint_dict_user_id_in_metadata` — dict with no known fields (extracts device_id)
- `test_messages_endpoint_dict_user_id_extracts_known_field` — dict with a `user_id` key

## AI Disclosure

This PR was authored with AI assistance (Hermes Agent).

### DIFF
--- a/src/any_llm/gateway/api/routes/messages.py
+++ b/src/any_llm/gateway/api/routes/messages.py
@@ -1,3 +1,4 @@
+import json
 from typing import Annotated, Any
 
 from fastapi import APIRouter, Depends, HTTPException, Request, Response, status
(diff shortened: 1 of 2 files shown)


tbille:
hi @xyaz1313 
Thank you for the contribution.

We are currently migrating the code of the gateway into its own repository. Could you make a pull request against: https://github.com/mozilla-ai/gateway

PR #1028, opened 2026-04-14
feat(gateway): rename auth header to AnyLLM-Key (RFC 6648)

peteski22:
## Description

Rename the gateway's Bearer-authentication header from `X-AnyLLM-Key` to `AnyLLM-Key`, so the name follows [RFC 6648](https://datatracker.ietf.org/doc/html/rfc6648) (no `X-` prefix). The gateway has no live deployments yet, so there is no backward-compatibility path — this is a simple rename.

This mirrors the same rename landing in [mozilla-ai/gateway#45](https://github.com/mozilla-ai/gateway/pull/45); both repos need to flip in step so the SDK's gateway client and the gateway itself agree on the header name.

### What changes

- `src/any_llm/gateway/core/config.py`: `API_KEY_HEADER = "AnyLLM-Key"`, comment cites RFC 6648.
- `src/any_llm/gateway/api/deps.py`: `_extract_bearer_token` walks a precedence tuple (`AnyLLM-Key` > `Authorization`, then `x-api-key` as the Anthropic-compatible fallback).
- `src/any_llm/gateway/main.py`: `Vary` hint and CORS `allow_headers` list use the new name via `API_KEY_HEADER`.
- `src/any_llm/providers/gateway/gateway.py`: `GATEWAY_HEADER_NAME` is now `"AnyLLM-Key"`, so the SDK's gateway client sends the canonical header by default.
- `docs/`: every curl example, quickstart snippet, and authentication reference now teaches the canonical header.
- `tests/gateway/test_client_args.py`, `test_key_management.py`, `test_provider_kwargs_override.py`: hard-coded header literals replaced with `API_KEY_HEADER` so the tests track the rename.
- `tests/unit/test_extract_bearer_token.py` (new): direct unit tests of `_extract_bearer_token` covering every precedence pair among `AnyLLM-Key`, `Authorization`, and `x-api-key`, malformed `Bearer` prefix on each bearer-format header, missing credentials, and empty-string fall-through.
- `tests/gateway/test_auth_header_precedence.py` (new): end-to-end integration tests that exercise the middleware stack: virtual-API-key auth on `/v1/models`, the `Vary` caching hint, and the CORS allow-list.

### Commits

1. `feat(gateway): rename auth header from X-AnyLLM-Key to AnyLLM-Key`
2. `test(gateway): cover AnyLLM-Key precedence and bearer-token extraction`

## PR Type

- 🆕 New Feature

## Relevant issues

Fixes #1023

## Checklist

[...]

PR #1031, opened 2026-04-14
fix(mistral): handle Optional choice.message from v2 SDK

peteski22:
## Description

The mistralai v2 SDK (adopted in #1020) types `choice.message` as `AssistantMessage | None`, but `_create_mistral_completion_from_response` dereferenced `message_data.content` and `message_data.tool_calls` without a None guard. This produced five `union-attr` errors under mypy strict mode:

```
src/any_llm/providers/mistral/utils.py:174: error: Item "None" of "AssistantMessage | None" has no attribute "content"  [union-attr]
src/any_llm/providers/mistral/utils.py:175: error: ... [union-attr]
src/any_llm/providers/mistral/utils.py:181: error: ... [union-attr]
src/any_llm/providers/mistral/utils.py:183: error: ... [union-attr]
src/any_llm/providers/mistral/utils.py:201: error: ... [union-attr]
```

The lint job on `main` has been intermittently failing since #1020 merged — whichever CI run hit the resolver state where mistralai's tightened typing was visible would fail. The most recent failure is on PR #1028.

### Fix

Add an explicit `ValueError` guard at the top of the loop so the converter fails loudly rather than silently emitting a truncated `ChatCompletion`. A real chat-completion choice always carries a message; if the SDK ever hands us `None`, we want to know rather than drop the choice silently.

Includes a unit test that covers the new guard path.

## PR Type

- 🐛 Bug Fix

## Relevant issues

Regression introduced by #1020 (`fix(mistral): update imports for mistralai v2 SDK`).

## Checklist

- [x] **AI Usage:**

## AI Usage Information

- AI Model used: Claude Opus 4.6 (1M context)

### DIFF
--- a/src/any_llm/providers/mistral/utils.py
+++ b/src/any_llm/providers/mistral/utils.py
@@ -170,6 +170,12 @@ def _create_mistral_completion_from_response(
 
     for i, choice_data in enumerate(response_data.choices):
         message_data = choice_data.message
+        # mistralai v2 types choice.message as Optional, but every chat-completion choice
+        # carries a message in practice. Fail loudly if the SDK ever hands us None so we
+        # do not silently emit a truncated ChatCompletion.
+        if message_data is None:
+            msg = "Mistral chat completion returned a choice without a message"
+            raise ValueError(msg)
 
         if message_data.content:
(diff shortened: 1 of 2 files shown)


njbrake (approved):
Looks good to me! 🚀

PR #1036, opened 2026-04-14
feat: add audio transcription and speech API support

tbille:
## Description
Add audio transcription (speech-to-text) and speech generation (text-to-speech) support to `any_llm` following the existing three-layer provider pattern. This enables providers with audio capabilities (e.g., OpenAI Whisper, OpenAI TTS) to handle audio via a unified API.

**New public API:**
- `atranscription(model, file, ...)` / `transcription(...)` — async/sync audio transcription
- `aspeech(model, input, voice, ...)` / `speech(...)` — async/sync text-to-speech (returns raw bytes)

**Changes:**
- `types/audio.py` — New `AudioTranscriptionParams` and `AudioSpeechParams` pydantic models + re-exported `Transcription`/`TranscriptionVerbose` from OpenAI types
- `any_llm.py` — Added `SUPPORTS_AUDIO_TRANSCRIPTION` and `SUPPORTS_AUDIO_SPEECH` flags (default `False`) and three-layer methods for each
- `api.py` — Top-level `transcription()`, `atranscription()`, `speech()`, `aspeech()` convenience functions with standard provider dispatch
- `providers/openai/base.py` — `BaseOpenAIProvider` implementation using `self.client.audio.transcriptions.create()` and `self.client.audio.speech.create()`
- `providers/openai/openai.py` — Explicit `SUPPORTS_AUDIO_TRANSCRIPTION = True`, `SUPPORTS_AUDIO_SPEECH = True`
- `providers/azureopenai/azureopenai.py` — Same flags
- `providers/gateway/gateway.py` — Same flags
- `types/provider.py` — Added `audio_transcription` and `audio_speech` fields to `ProviderMetadata`
- `__init__.py` — Exported new functions

## PR Type

- 🆕 New Feature

## Relevant issues

N/A

## Checklist
- **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

- AI Model used: Claude Opus 4
- AI Developer Tool used: OpenCode
- Any other info you'd like to share: Used for implementation scaffolding and test generation


src/any_llm/types/audio.py
```diff
@@ -0,0 +1,47 @@
+"""Audio transcription and speech types for any_llm."""
+
+from typing import IO, Any, Literal
+
+from openai.types.audio import Transcription as OpenAITranscription
+from openai.types.audio import TranscriptionVerbose as OpenAITranscriptionVerbose
+from pydantic import BaseModel, ConfigDict
+
+
+class AudioTranscriptionParams(BaseModel):
```
njbrake, line 10: Are there any existing types that we should import here instead of redefinining from a pydantic basemodel? Since the output formats re-export OpenAI types it's not clear why we don't do the same here.

njbrake (dismissed):
One comment about redefining a type but other than that, LGTM 🚀

src/any_llm/types/audio.py
tbille: little check with my friend Claude:

> The OpenAI SDK does export TranscriptionCreateParams and SpeechCreateParams, but those are TypedDicts with OpenAI-specific fields (like stream). For input parameter types, the codebase convention is to define custom Pydantic BaseModels (see CompletionParams and ResponsesParams) rather than re-exporting SDK types. The re-export pattern is only used for output/response types, which is what we do here with Transcription and TranscriptionVerbose.

dpoulopoulos (dismissed):

dpoulopoulos (approved):

Issue #1048, opened 2026-04-23
[DOC] Broken link: llms.txt returns 404


Hanaasagi:
- [x] I have already searched for related issues.

The documentation link for `llms.txt` and `llms-full.txt` appears to be broken.

https://github.com/mozilla-ai/any-llm/blob/e944c46e36388f913a64dee5ad0f04c3078486fb/docs/index.md?plain=1#L72-L73

Hanaasagi:
I see this has already been fixed, thank you.

PR #1050, opened 2026-04-23
Feat/gitbook docs

angpt:
## Description
- Migrate documentation pipeline from Astro/Starlight to GitBook-
- Stop tracking generated docs in docs/ — they are now CI/build artifacts
- Update all docs links from mozilla-ai.github.io/any-llm to docs.mozilla.ai/any-llm
- Point llms.txt reference to the unified docs.mozilla.ai/llms.txt

## PR Type
- 📚 Documentation
- 🚦 Infrastructure

## Relevant issues
https://github.com/mozilla-ai/any-llm/issues/1048

## Checklist

- [X] I understand the code I am submitting.
- [X] I have run this code locally and verified it fixes the issue.
- [X] I have read and followed the [contribution guidelines](https://github.com/mozilla-ai/any-llm/blob/main/CONTRIBUTING.md)
- [X] **AI Usage:**
    - [ ] No AI was used.
    - [X] AI was used for drafting/refactoring.

## AI Usage Information

- AI Model used:
- AI Developer Tool used:

### DIFF
--- a/.github/workflows/gateway-tests.yml
+++ b/.github/workflows/gateway-tests.yml
@@ -49,20 +49,3 @@ jobs:
       with:
         file: ./coverage.xml
         fail_ci_if_error: false
-
-  openapi-spec:
-    runs-on: ubuntu-latest
-
-    steps:
-    - uses: actions/checkout@v6
-
-    - uses: astral-sh/setup-uv@v7
-      with:
-        python-version: 3.13
-        activate-environment: true
-
-    - name: Install dependencies
-      run: uv sync --extra gateway --extra all -U --python=3.13
-
-    - name: Verify OpenAPI spec is valid
-      run: python scripts/generate_openapi.py
--- a/.github/workflows/gateway-tests.yml
+++ b/.github/workflows/gateway-tests.yml
@@ -64,5 +64,5 @@ jobs:
     - name: Install dependencies
       run: uv sync --extra gateway --extra all -U --python=3.13
 
-    - name: Verify OpenAPI spec is up to date
-      run: python scripts/generate_openapi.py --check
+    - name: Verify OpenAPI spec is valid
+      run: python scripts/generate_openapi.py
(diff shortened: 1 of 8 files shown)


.github/workflows/gateway-tests.yml
```diff
@@ -64,5 +64,5 @@ jobs:
     - name: Install dependencies
       run: uv sync --extra gateway --extra all -U --python=3.13
 
-    - name: Verify OpenAPI spec is up to date
-      run: python scripts/generate_openapi.py --check
```
njbrake, line 68: Why this change, woudln't this make it so that mismatched openapi specs no longer throw errors?

njbrake (changes requested):
It's not clear to me why it is ok to remove the --check, I think this would make us not know if the openapi spec is no longer compliant?

angpt:
The spec is now freshly generated from the code on every CI run, so it can't drift - which makes the `--check` unnecessary.

njbrake:
What should be cleaned up:

  1. The openapi-spec job in .github/workflows/gateway-tests.yml (lines 53-68) should be removed. It's checking a file that no longer exists in SCM, so it's either silently broken or will fail.
  2. The --check mode in generate_openapi.py is now dead code (no committed file to check against). You could remove it, or keep it if you foresee needing snapshot comparison in the future.
  3. The help text on line 109 still says docs/public/openapi.json even though the default was changed to docs/openapi.json.

  Want me to clean these up?

✻ Cogitated for 1m 18s

❯ what help text?

● Looking at the script again:

  parser.add_argument(
      "--output",
      type=Path,
      default=Path(__file__).parent.parent / "docs" / "openapi.json",
      help="Output path for OpenAPI spec (default: docs/public/openapi.json)",
  )

  The default was changed to docs/openapi.json in the GitBook commit, but the help string still says docs/public/openapi.json. Minor mismatch, but it's there on line 109.

angpt:
@njbrake Your comments have been addressed in the latest commit.

njbrake (approved):

PR #1071, opened 2026-05-07
fix: use NullHandler instead of setup_logger at import time

dashitongzhi:
## Description
Replace unconditional `setup_logger()` at module level with `NullHandler` per Python library logging best practices.

## Problem

`setup_logger()` was called unconditionally at module level in `logging.py`, which installed a `RichHandler` on the `any_llm` logger whenever the library was imported. This violates Python library logging best practice - libraries should use `NullHandler` so that importing them does not reconfigure logging for the entire application.

Per [Python docs](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library):
> It is strongly advised that you do not add any handlers other than NullHandler to your library loggers.

## Fix

Replace `setup_logger()` at module level with `logger.addHandler(logging.NullHandler())`. The `setup_logger()` function is preserved so users can still call it explicitly to enable RichHandler output.

## PR Type
- 🐛 Bug Fix

## Testing

Added `tests/unit/test_logging.py` with 3 tests verifying:
- Logger has NullHandler at import time
- No RichHandler is installed at import time
- Logger name matches the package name

## Checklist
- [x] I understand the code I am submitting.
- [x] I have added unit tests that prove my fix/feature works
- [x] I have run this code locally and verified it fixes the issue.
- [x] New and existing tests pass locally
- [x] Documentation was updated where necessary
- [x] I have read and followed the [contribution guidelines](https://github.com/mozilla-ai/any-llm/blob/main/CONTRIBUTING.md)
- [ ] **AI Usage:**
    - [x] No AI was used.
    - [ ] AI was used for drafting/refactoring.
    - [ ] This is fully AI-generated.

## AI Usage Information
- AI Model used: N/A
- AI Developer Tool used: N/A
- Any other info you like to share: N/A

- [ ] I am an AI Agent filling out this form (check box if true)

PR #1074, opened 2026-05-15
fix: stop reconfiguring consumer logging on import (Fixes #1024)

### LINKED ISSUE #1024: Library reconfigures consumer's logging at import time
`any_llm` and `any_llm.gateway` violate Python's [library logging guidance](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library) by configuring handlers at module-import time. A consumer that does `import any_llm` has their logging silently reshaped before they can object.

**Where it happens:**

- `src/any_llm/logging.py:41` — `setup_logger()` runs on import.
- `src/any_llm/gateway/log_config.py:41` — same pattern for the `gateway` logger.

**What `setup_logger()` does that a library should not:**

1. Attaches a `RichHandler` directly to the named logger (`any_llm` / `gateway`).
2. Sets `logger.propagate = False`, which actively prevents the consumer from capturing these logs through the root logger or any ancestor handler — even if they want to.
3. Removes any handlers already attached to the logger (`for handler in logger.handlers[:]: logger.removeHandler(handler)`), wiping consumer configuration if the consumer happened to set it up first.
4. Forces a `rich` dependency on every consumer, including those who never use the gateway and don't want pretty terminal output (CI logs, JSON log shippers, etc.).

**Concrete consumer impact:**

- A FastAPI app that mounts the gateway router and configures structured JSON logging via `structlog` or `python-json-logger` will have plain `RichHandler` text output for `gateway.*` and `any_llm.*` loggers, and cannot route them through its own pipeline because `propagate=False`.
- A consumer that called `logging.getLogger("any_llm").addHandler(...)` before importing further any-llm submodules can have their handler stripped by a later import path that triggers `setup_logger()` re-execution.
- Test suites that capture logs via `caplog` (which relies on propagation to the root logger) will not see any-llm or gateway logs.

**Proposed fix:**

1. Replace the module-level `setup_logger()` calls with `logger.addHandler(logging.NullHandler())`. This is the standard library pattern: silent by default, no handler, no formatter, no propagation override.
2. Keep `setup_logger()` exported as an opt-in convenience for application entry points (the gateway CLI is a legitimate caller — `gateway/cli.py:54`).
3. In `setup_logger()`, stop setting `propagate = False` unconditionally. The default should be `propagate = True`; let callers opt out if they truly need to.
4. In `setup_logger()`, stop unconditionally removing existing handlers. Either document it as "resets the handler chain" or guard with an explicit `reset: bool = False` argument.
5. Move `rich` from a hard dependency to an optional extra (`any-llm-sdk[rich]` or similar). `setup_logger()` can import it lazily and raise a helpful error if the extra isn't installed.

**Backward compatibility:**

Anyone relying on `import any_llm` producing pretty log output will see plain logs instead (i.e., none, until they configure handlers themselves). This is a one-line fix on their side: `from any_llm.logging import setup_logger; setup_logger()`. Worth calling out in release notes but not a true API break — the public function is unchanged.

KardeniaPoyu:
## Description
Fixes #1024 by implementing Python logging best practices for libraries.

### `logging.py` & `gateway/log_config.py`
1. Replace module-level `setup_logger()` with `NullHandler` pattern
2. Change `propagate` default from `False` to `True`
3. Add `reset` parameter for opt-in handler removal
4. Lazy-import `RichHandler` with helpful ImportError

### `pyproject.toml`
5. Move `rich` from hard dependency to optional extra (`pip install any-llm-sdk[rich]`)

## PR Type
- [x] Bug Fix

## Relevant issues
Fixes #1024

## Checklist
- [x] **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information
- AI Model used: DeepSeek V4 (via OpenClaw)
- AI Developer Tool used: OpenClaw
- Any other info you'd like to share: This PR was created by an AI agent following the project's contributing guidelines.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,7 +13,6 @@ dependencies = [
   "openai>=1.99.3",
   "openresponses-types>=2.3.0.post1",
   "anthropic>=0.83.0",
-  "rich",
   "httpx",
   "typing_extensions>=4.5.0",
 ]
@@ -28,6 +27,10 @@ platform = [
   "any-llm-platform-client>=0.3.0",
   "opentelemetry-sdk>=1.30.0",
   "opentelemetry-exporter-otlp-proto-http>=1.30.0",
+]
+
+rich = [
+  "rich",
 ]
 
 perplexity = []
@@ -126,6 +129,7 @@ vllm = []
 # Deprecated: the gateway is moving to a standalone package.
 # See https://github.com/mozilla-ai/gateway
 gateway = [
+  "rich",
   "fastapi>=0.115.0",
   "uvicorn[standard]>=0.30.0",
   "sqlalchemy>=2.0.0",
--- a/src/any_llm/gateway/log_config.py
+++ b/src/any_llm/gateway/log_config.py
@@ -1,8 +1,6 @@
 import logging
 from typing import Any
 
-from rich.logging import RichHandler
-
 logger = logging.getLogger("gateway")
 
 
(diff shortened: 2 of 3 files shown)


tbille:
hi @KardeniaPoyu thank you for your contribution. This bug was just fixed by https://github.com/mozilla-ai/any-llm/pull/1071 and we would rather keep the logging with the standard libraries for this sdk.
Closing this pull request.

PR #1080, opened 2026-05-19
docs: redirect GitHub Pages docs to docs.mozilla.ai

angpt:
Retire GitHub pages, and redirect all users to GitBook

## Description
Redirect users from GitHub pages to docs.mozilla.ai

## PR Type

- 📚 Documentation
- 🚦 Infrastructure

## Checklist

- [X] I understand the code I am submitting.
- [X] I have run this code locally and verified it fixes the issue.
- [X] I have read and followed the [contribution guidelines](https://github.com/mozilla-ai/any-llm/blob/main/CONTRIBUTING.md)
- [X] **AI Usage:**
    - [ ] No AI was used.
    - X AI was used for drafting/refactoring.

## AI Usage Information

- AI Model used:
- AI Developer Tool used:

### DIFF
--- a/.github/workflows/pages-redirect.yaml
+++ b/.github/workflows/pages-redirect.yaml
@@ -1,0 +1,85 @@
+name: GitHub Pages Redirects
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - 'docs/**'
+      - 'src/**'
+      - 'scripts/generate_provider_table.py'
+      - 'scripts/generate_openapi.py'
+      - 'scripts/generate_api_docs.py'
+      - 'scripts/generate_cookbooks.py'
+      - 'scripts/convert_to_gitbook.py'
+      - 'scripts/generate_github_pages_redirects.py'
+      - '.github/workflows/pages-redirect.yaml'
+      - 'pyproject.toml'
+  pull_request:
+    paths:
+      - 'docs/**'
+      - 'src/**'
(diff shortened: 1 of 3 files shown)


njbrake:
Looks like this will work: I wonder if there is a more straightforward way via DNS isntead of needing to deploy a shell site? As I understand you want that mozilla-ai.github.io/any-llm site url to point to gitbooks instead of being a site, I wonder if that is possible via dns settings.

tbille:
@njbrake @angpt instead we could keep the pages live but set the canonical URL in the HTML headers be the ones from docs.mozilla.ai. This way we let SEO do a bit its magic and once we see no more traffic we remove the pages.

angpt:
@njbrake Since we don't use a Mozilla owned domain with GH pages, we won't be able to change what the old docs url will resolve to. Not sure if you were thinking of a different way. 
@tbille Canonical tags would help with SEO, but redirects are the stronger signal for a moved site and also improve the actual user experience by sending bookmarked or linked old GitHub Pages URLs straight to docs.mozilla.ai. [Google Docs](https://developers.google.com/search/docs/crawling-indexing/301-redirects)

tbille (approved):

tbille:
@angpt yes but, in this case we are doing a redirect from JS which is not as effective as a redirect from the backend (with a 301 status code).
Anyway, I think it's ok to merge it as it is.

PR #1133, opened 2026-06-12
feat(messages): support structured output via output_format

njbrake:
_Note: this PR was implemented by Claude (Opus 4.8) via Claude Code, working with @njbrake. The reasoning and design decisions are his; the code and prose are Claude's. Design analysis by @avalyset in #1073 shaped the approach (credited as co-author)._

## Description

`messages()` / `amessages()` did not accept a structured-output option, so callers could not get typed JSON the way `completion()` and `responses()` already allow (the gap reported in #1073).

This adds structured output to the Messages API using Anthropic's own primitive, `messages.parse(output_format=...)`, keeping the any-llm Messages API a faithful mirror of the Anthropic Messages API. No Anthropic types are extended: when `output_format` is given, the call returns Anthropic's own `ParsedMessage`, with the typed object on its `parsed_output` property and on each text block, identical to what the Anthropic SDK itself returns.

- `MessagesParams` gains `output_format` (a Pydantic model or dataclass type).
- The **Anthropic** provider calls native `client.messages.parse`, which drives the GA `output_config` primitive (schema-constrained generation), and returns the SDK's `ParsedMessage` unchanged.
- **Other providers** route `output_format` through the existing Messages to Completions bridge as the completion `response_format`; a new `build_parsed_message` helper assembles the same `ParsedMessage` shape from the response text, so the result is indistinguishable across providers.
- Streaming combined with `output_format` raises `ValueError`, matching the behavior of `responses()`.

Validated against the pinned `anthropic` SDK (0.105.2): `output_config` is GA with no beta header, `messages.parse` exists on both sync and async clients, `ParsedMessage.parsed_output` is a property over the text blocks, and runtime-parametrized `ParsedMessage[T]` / `ParsedTextBlock[T]` construction validates for both Pydantic models and dataclasses.

Raw `output_config` dict passthrough for non-Pydantic JSON schemas is intentionally left as a follow-up.

## PR Type

- 🆕 New Feature

## Relevant issues

Fixes #1073

## Checklist

- [x] I understand the code I am submitting.
- [x] I have added unit tests that prove my fix/feature works
- [x] I have run this code locally and verified it fixes the issue.
- [x] New and existing tests pass locally
- [x] Documentation was updated where necessary
- [x] I have read and followed the [contribution guidelines](https://github.com/mozilla-ai/any-llm/blob/main/CONTRIBUTING.md)
- [x] **AI Usage:**
    - [ ] No AI was used.
    - [ ] AI was used for drafting/refactoring.
[...]

Issue #1073, opened 2026-05-15
Enhancement: Supporting structured output for messages API


AbdealiLoKo:
Hi, thanks a bunch of any-llm-sdk, it has really simplified a lot of boiler plate code I have

I was using the Messages API - https://mozilla-ai.github.io/any-llm/api/messages/
and was struggling to find a way to get my output as a stuctured json.

And then saw that Completions API has that feature - https://mozilla-ai.github.io/any-llm/api/completion/#structured-output

My current understanding (from minimal googling) is that OpenAI Completions API is the older format and OpenAI Chat API or Anthropic Messages API is preferred going forward.

I was curious to understand whether I was using the wrong APIs or is it just a feature that is not implemented currently ?
Would love to get some clarity so I can take the full benefit of any-llm-sdk features. Thanks

avalyset:
I'd like to take this. Reading through `api.py` and `messages_compat.py`, the gap is clear: `completion()` and `responses()` both accept `response_format` and thread it through (with `responses()` returning `output_parsed` for Pydantic/dataclass types), but `messages()`/`amessages()` have no such parameter and `MessagesParams` has no field for it.

Before writing it I want to confirm direction on three design points, since the Anthropic message shape makes this non-trivial:

1. Where parsed output lands. `MessageResponse` extends `AnthropicMessage` (text/tool_use/thinking content blocks) and has no `output_parsed` slot like `ParsedResponse`. Should structured output surface as a new parsed variant, or be mapped into a content block?
2. Underlying mechanism. The Anthropic Messages API has no native `response_format`; structured output is normally done via tool-use/JSON-schema. Do we add `response_format` to the OpenAI-bridge path (`messages_params_to_completion_params` → completion → `chat_completion_to_message_response`), or implement it per-provider on `llm.messages`?
3. Streaming consistency. `chat_completion_chunk_to_message_stream_events` currently only emits text/thinking/tool_use blocks. If parsed output goes via the completion bridge, how should it map back consistently for the streaming case?

My instinct is the OpenAI-bridge path with a parsed variant on the response, mirroring how `responses()` already does it — but I'd rather align on the approach than send a PR that picks the wrong one. Happy to implement once there's a steer.

itsayellow:
Isn't `output_config.format` in Anthropic's `messages.create()` what you're looking for?
https://platform.claude.com/docs/en/api/python/messages/create#create.output_config.format

[...]

PR #1134, opened 2026-06-12
feat(messages): accept raw output_config dict for non-Pydantic structured output

avalyset:
## Description

@njbrake offered this one there — "the raw `output_config` dict passthrough for non-Pydantic JSON schemas is a clean, self-contained follow-up and it is yours if you want it" — so I'm picking it up.

#1133 added structured output to the Messages API via `output_format`, but only as a Pydantic `BaseModel`/dataclass **type**. This extends `output_format` to also accept a raw Anthropic `output_config` **dict**, so callers with a non-Pydantic JSON schema (e.g. loaded from a file) can use schema-constrained generation without defining a Python type.

It keeps #1133's approach of mirroring the Anthropic Messages API:
- **Anthropic provider**: a dict is passed straight to native `messages.create(output_config=...)` (the same GA primitive `messages.parse` drives for the typed path); a type still uses `messages.parse`.
- **Other providers**: the dict's schema is bridged to the completion `response_format` as an OpenAI `json_schema`, the same path types already take.
- **Return shape**: both produce Anthropic's `ParsedMessage`. For a type, `parsed_output` holds the typed object; for a raw dict, it holds the parsed JSON (a plain `dict`/`list`), since there is no Python type to validate into.
- `output_format` with `stream=True` raises `ValueError` for the dict case too, matching the typed path.

No SDK floor bump beyond what #1133 already assumes — `output_config` is the same GA primitive. Validated against the pinned `anthropic` (0.104.1): `messages.create(output_config=...)` accepts the dict and the `OutputConfigParam`/`JSONOutputFormatParam` shape matches.

## PR Type

- 🆕 New Feature

## Relevant issues

Follows #1073 / #1133.

## Checklist

- [x] **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

- Any other info you'd like to share: Architecture and all design decisions are @avalyset's, carried over from the #1073 analysis — the cross-provider raw-dict passthrough returning ParsedMessage, the bridge semantics, and the streaming behavior. Claude implemented the chosen design and wrote the tests.

## Summary by CodeRabbit

* **New Features**
  * Structured-output now accepts raw JSON-schema-style `output_config` dictionaries in addition to type-based definitions, with matching parsing behaviour.

* **Bug Fixes**
  * Improved handling of structured parsing vs non-parsed text paths to ensure the correct output is produced for each `output_format` form.

* **Documentation**
  * Updated public API docs to describe both supported structured-output inputs and reiterated that structured parsing is not supported with streaming.

* **Tests**
  * Added coverage for the raw `output_config` pass-through path, JSON decoding/parsing outcomes, translation to response formats, and the expected error when used with streaming.

### DIFF
--- a/src/any_llm/any_llm.py
+++ b/src/any_llm/any_llm.py
@@ -691,7 +691,7 @@ class AnyLLM(ABC):
         metadata: dict[str, Any] | None = None,
         thinking: dict[str, Any] | None = None,
         cache_control: dict[str, Any] | None = None,
-        output_format: type | None = None,
+        output_format: type | dict[str, Any] | None = None,
         **kwargs: Any,
     ) -> MessageResponse | ParsedMessage[Any] | AsyncIterator[MessageStreamEvent]:
         """Create a message using the Anthropic Messages API asynchronously.
@@ -714,10 +714,11 @@ class AnyLLM(ABC):
             metadata: Request metadata.
             thinking: Thinking/reasoning configuration.
             cache_control: Cache control configuration for prompt caching.
-            output_format: Structured-output type, mirroring Anthropic's
-                ``messages.parse(output_format=...)``. A Pydantic ``BaseModel`` or dataclass
-                makes the call return Anthropic's ``ParsedMessage``, with the typed object on
-                its ``parsed_output`` property. Not supported with ``stream=True``.
+            output_format: Structured output, mirroring Anthropic's ``messages.parse``/
+                ``output_config``. Either a Pydantic ``BaseModel``/dataclass **type** (typed
+                ``parsed_output``) or a raw Anthropic ``output_config`` **dict** for non-Pydantic
+                JSON schemas (``parsed_output`` holds the parsed JSON). The call returns
+                Anthropic's ``ParsedMessage``. Not supported with ``stream=True``.
             **kwargs: Additional provider-specific arguments.
 
         Returns:
@@ -728,7 +729,7 @@ class AnyLLM(ABC):
             ValueError: If `output_format` is combined with `stream=True`.
 
         """
-        if is_structured_output_type(output_format) and stream:
+        if output_format is not None and stream:
             msg = "stream is not supported for output_format"
             raise ValueError(msg)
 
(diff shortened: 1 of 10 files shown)


avalyset:
Thanks @njbrake — CodeRabbit's three points are addressed in `091407b`.

On the malformed-JSON case: the typed path from #1133 lets `parse_json_content` raise raw on a parse failure — a pydantic `ValidationError` — so the raw-dict path leaves `json.loads` to raise `json.JSONDecodeError` the same way, no wrapper. The new test pins that, so both structured-output paths behave the same when parsing fails.

The other two are mechanical: `_parse` now returns `object` instead of `Any`, and there's a regression test for a title-less `output_config` schema falling back to the default name.

njbrake (approved):

Approving. This is a clean, self-contained follow-up to #1133 that extends `output_format` to accept a raw Anthropic `output_config` dict alongside the existing Pydantic/dataclass type path.

What I checked:
- **Routing**: Anthropic provider sends a type to `messages.parse(output_format=...)` and a dict to `messages.create(output_config=...)`, then converts to a `MessageResponse`. Other providers bridge the dict's schema to an OpenAI `json_schema` `response_format`, which `CompletionParams.response_format` already accepts. Sound.
- **Stream guard and post-processing**: broadening the checks from `is_structured_output_type(...)` to `output_format is not None` is correct and introduces no regression to the typed path (the Anthropic typed path returns a `ParsedMessage`, so the rebuild is skipped; the dict and bridged paths return a `MessageResponse`, so it runs).
- **Error semantics**: the raw-dict path lets `json.JSONDecodeError` propagate, mirroring the typed path letting pydantic's `ValidationError` propagate. Consistent, and pinned by a test.
- Tests pass, ruff and mypy are clean.

I pushed one small cleanup to the branch (6150006): `json` is a core dependency, so its import was moved to module scope in `structured_output.py`, matching the convention already applied to the test file. The inline `anthropic.types` imports correctly stay inline.

One non-blocking follow-up to consider later: `_output_config_to_response_format` derives the `json_schema` name from the schema's `title`. OpenAI requires names matching `^[a-zA-Z0-9_-]{1,64}$`, so a title with spaces would be rejected on the bridged path, whereas the typed path uses `__name__` and Anthropic's native path passes the dict through untouched. Edge case and user-controlled, so fine for this PR.

PR #1151, opened 2026-06-27
fix(messages): report Anthropic streaming usage from the trailing usage-only chunk

Syrunekai:
## Description

When bridging an OpenAI-compatible provider to the Anthropic Messages API, streaming via `amessages` reported zero token usage. OpenAI-compatible providers emit token counts in a final usage-only chunk that arrives *after* the `finish_reason` chunk, but the bridge emitted the closing `message_delta`/`message_stop` on the `finish_reason` chunk — before that usage was available — so `message_delta` always reported `input_tokens=0`/`output_tokens=0` (and no cache).

The fix defers the closing `message_delta`/`message_stop` to the stream wrapper's post-loop flush so usage is complete by the time they're emitted, and maps `prompt_tokens_details.cached_tokens` → `cache_read_input_tokens`. Since the converter no longer emits those closing events, it also removes the now-dead `emitted_stop` bookkeeping in the wrapper and narrows the converter's return type to the events it actually produces.

**Verified** against multiple OpenAI-compatible providers (including OpenRouter and Anthropic's OpenAI-compatible endpoint), and cross-checked against the native Anthropic provider as a reference for usage/cache placement — native Anthropic reports cache in `message_start`, while the bridge necessarily reports it in `message_delta` because OpenAI-compatible usage arrives last (both are valid per the SDK). New unit tests added; full unit suite + `pre-commit` (ruff + mypy strict) pass locally.

## PR Type

- 🐛 Bug Fix

## Relevant issues

None found.

## Checklist

- [x] **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

- Any other info you'd like to share: AI-authored, but human-directed and reviewed change-by-change. On your "discuss with the human, not the AI" policy — fully respected: the strings here are pulled by a human. For any review discussion that needs a real conversation, my puppeteer @0xSylice will reply personally (not paste my answers back).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved streaming event ordering so final stop reasons are emitted reliably, including when a stream ends with trailing usage-only chunks.
  * Streaming usage is now flushed correctly on normal completion and mid-stream failures, and cached input tokens are reported when available.
  * Standardised stop-reason handling across streaming conversions, with consistent defaults when no finish reason is provided.
* **Tests**
  * Expanded unit and compatibility tests for cached-token scenarios (present/zero/missing), stop-reason semantics, and robustness during exceptions and early closure.

### DIFF
--- a/src/any_llm/any_llm.py
+++ b/src/any_llm/any_llm.py
@@ -785,16 +785,11 @@ class AnyLLM(ABC):
 
         async def convert_stream() -> AsyncIterator[MessageStreamEvent]:
             state = StreamingState()
-            emitted_stop = False
             async for chunk in result:
-                events = chat_completion_chunk_to_message_stream_events(chunk, state)
-                for event in events:
-                    if isinstance(event, MessageStopEvent):
-                        emitted_stop = True
+                for event in chat_completion_chunk_to_message_stream_events(chunk, state):
                     yield event
-            # Some providers don't send a final chunk with finish_reason,
-            # so ensure the stream always ends with message_stop.
-            if state.started and not emitted_stop:
+            # Emit the closing events after the full stream is consumed so trailing-chunk usage is included.
+            if state.started:
                 if state.current_block_type is not None:
                     yield ContentBlockStopEvent(
                         type="content_block_stop",
@@ -802,10 +797,11 @@ class AnyLLM(ABC):
                     )
                 yield MessageDeltaEvent(
                     type="message_delta",
-                    delta=MessageDelta(stop_reason="end_turn"),
+                    delta=MessageDelta(stop_reason=state.stop_reason or "end_turn"),
                     usage=MessageDeltaUsage(
                         output_tokens=state.output_tokens,
                         input_tokens=state.input_tokens,
+                        cache_read_input_tokens=state.cache_read_input_tokens or None,
                     ),
                 )
                 yield MessageStopEvent(type="message_stop")
--- a/src/any_llm/any_llm.py
+++ b/src/any_llm/any_llm.py
@@ -804,8 +804,10 @@ class AnyLLM(ABC):
                         yield event
             except Exception:
                 # Flush the usage accumulated so far before re-raising, so a mid-stream failure still reports tokens.
+                # Report stop_reason=None so a finish_reason seen before the failure is never mistaken for a
+                # successful completion; message_stop stays reserved for the clean-completion path below.
                 if state.started:
-                    yield usage_delta(state.stop_reason)
+                    yield usage_delta(None)
                 raise
             # Emit the closing events after the full stream is consumed so trailing-chunk usage is included.
             if state.started:
--- a/src/any_llm/any_llm.py
+++ b/src/any_llm/any_llm.py
@@ -37,6 +37,7 @@ from any_llm.types.messages import (
     MessageStopEvent,
     MessageStreamEvent,
     ParsedMessage,
+    StopReason,
 )
 from any_llm.types.provider import ProviderMetadata
 from any_llm.types.responses import (
@@ -785,9 +786,27 @@ class AnyLLM(ABC):
 
         async def convert_stream() -> AsyncIterator[MessageStreamEvent]:
             state = StreamingState()
-            async for chunk in result:
-                for event in chat_completion_chunk_to_message_stream_events(chunk, state):
-                    yield event
+
+            def usage_delta(stop_reason: StopReason | None) -> MessageDeltaEvent:
+                return MessageDeltaEvent(
+                    type="message_delta",
+                    delta=MessageDelta(stop_reason=stop_reason),
+                    usage=MessageDeltaUsage(
+                        output_tokens=state.output_tokens,
+                        input_tokens=state.input_tokens,
+                        cache_read_input_tokens=state.cache_read_input_tokens or None,
+                    ),
+                )
+
+            try:
+                async for chunk in result:
+                    for event in chat_completion_chunk_to_message_stream_events(chunk, state):
+                        yield event
+            except Exception:
+                # Flush the usage accumulated so far before re-raising, so a mid-stream failure still reports tokens.
+                if state.started:
+                    yield usage_delta(state.stop_reason)
+                raise
             # Emit the closing events after the full stream is consumed so trailing-chunk usage is included.
             if state.started:
                 if state.current_block_type is not None:
@@ -795,15 +814,7 @@ class AnyLLM(ABC):
                         type="content_block_stop",
                         index=state.current_block_index,
                     )
-                yield MessageDeltaEvent(
-                    type="message_delta",
-                    delta=MessageDelta(stop_reason=state.stop_reason or "end_turn"),
-                    usage=MessageDeltaUsage(
-                        output_tokens=state.output_tokens,
-                        input_tokens=state.input_tokens,
-                        cache_read_input_tokens=state.cache_read_input_tokens or None,
-                    ),
-                )
+                yield usage_delta(state.stop_reason or "end_turn")
                 yield MessageStopEvent(type="message_stop")
 
         return convert_stream()
--- a/tests/unit/test_messages.py
+++ b/tests/unit/test_messages.py
@@ -504,6 +504,62 @@ async def test_default_amessages_streaming_flushes_usage_when_stream_fails() -> 
 
 
 @pytest.mark.asyncio
+async def test_default_amessages_streaming_failure_after_finish_reason_reports_no_stop_reason() -> None:
+    """A stream that emits finish_reason and then fails must not report a completion stop_reason."""
+
+    async def mock_stream() -> AsyncIterator[ChatCompletionChunk]:
+        yield ChatCompletionChunk(
+            id="chunk-1",
+            model="gpt-4",
+            created=0,
+            object="chat.completion.chunk",
+            choices=[ChunkChoice(index=0, delta=ChoiceDelta(content="Hi"), finish_reason=None)],
+            usage=CompletionUsage(prompt_tokens=100, completion_tokens=10, total_tokens=110),
+        )
+        yield ChatCompletionChunk(
+            id="chunk-2",
+            model="gpt-4",
+            created=0,
+            object="chat.completion.chunk",
+            choices=[ChunkChoice(index=0, delta=ChoiceDelta(), finish_reason="stop")],
+        )
+        # The trailing usage-only chunk never arrives; the provider stream drops instead.
+        msg = "provider stream dropped"
+        raise RuntimeError(msg)
+
+    mock_provider = Mock()
+    mock_provider._acompletion = AsyncMock(return_value=mock_stream())
+
+    params = MessagesParams(
+        model="gpt-4",
+        messages=[{"role": "user", "content": "Hello"}],
+        max_tokens=100,
+        stream=True,
+    )
+    result = await AnyLLM._amessages(mock_provider, params)
+    assert not isinstance(result, (MessageResponse, ParsedMessage))
+
+    seen: list[Any] = []
+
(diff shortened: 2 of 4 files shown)


Syrunekai:
Thanks @coderabbitai — checked each against the repo's config and conventions:

**1. Split the converter (PLR0912):** the repo's `[tool.ruff.lint]` config ignores `PLR`/`PLR0912`, so the project's own linter doesn't flag this and pre-commit is green. I'd prefer not to restructure the converter inside a focused bug-fix PR — glad to do it as a separate change if the maintainers want it.

**2. `is not None` for usage values:** the `prompt_tokens`/`completion_tokens` checks are pre-existing and outside this change. For the `cached_tokens` line I added, truthy vs `is not None` produces the same result for the single-usage-chunk case providers actually emit (and the wrapper does `state.cache_read_input_tokens or None`, so a `0` collapses to an omitted field either way); the zero case is covered by a new unit test. Left as-is to keep the diff surgical — happy to switch if you'd prefer the explicit `None` check.

**3. Imports to top / `-> Any` (ANN401):** the repo's ruff config ignores `ANN`, and this test matches the file's established pattern — inline imports throughout, and an identical `mock_stream() -> Any` in the sibling test (`test_default_amessages_streaming`, ~line 356). Kept it consistent with the file.

Happy to make any of these if a maintainer sees them as blocking.

0xSylice:
Human in the loop, reporting in.

This is a fairly simple fix. When a response is streamed out through the Anthropic messages format, the token usage numbers (input, output, and how many were cached) only land right at the very end of the stream.

The code was closing out and sending the usage report a bit too early, before those final numbers had actually arrived, so it always returned as zero.

The fix just waits until the stream is completely finished before reporting, so the real values (cached count included) are actually there. That's pretty much it!

- Sylice

src/any_llm/any_llm.py
```diff
@@ -785,27 +785,23 @@ async def _amessages(
...
                     type="message_delta",
-                    delta=MessageDelta(stop_reason="end_turn"),
+                    delta=MessageDelta(stop_reason=state.stop_reason or "end_turn"),
                     usage=MessageDeltaUsage(
                         output_tokens=state.output_tokens,
                         input_tokens=state.input_tokens,
+                        cache_read_input_tokens=state.cache_read_input_tokens or None,
                     ),
                 )
                 yield MessageStopEvent(type="message_stop")
```
tbille: **Usage is lost when the stream fails mid-iteration.** These closing events are the only ones carrying usage, and they are now reached only if `async for` completes normally. If `result` raises during iteration, this block is skipped and the caller gets zero usage despite `state` already holding accumulated tokens.

Wrap the loop so the closing events are flushed on failure too, while still handling early consumer-close (`GeneratorExit`, where yielding is illegal) and avoiding double emission:

```python
async def convert_stream() -> AsyncIterator[MessageStreamEvent]:
    state = StreamingState()
    emitted_closing = False

    def closing_events() -> list[MessageStreamEvent]:
        events: list[MessageStreamEvent] = []
        if not state.started:
            return events
        if state.current_block_type is not None:
            events.append(ContentBlockStopEvent(type="content_block_stop", index=state.current_block_index))
        events.append(
            MessageDeltaEvent(
                type="message_delta",
                delta=MessageDelta(stop_reason=state.stop_reason or "end_turn"),
                usage=MessageDeltaUsage(
                    output_tokens=state.output_tokens,
                    input_tokens=state.input_tokens,
                    cache_read_input_tokens=state.cache_read_input_tokens or None,
                ),
            )
        )
        events.append(MessageStopEvent(type="message_stop"))
        return events

    try:
        async for chunk in result:
            for event in chat_completion_chunk_to_message_stream_events(chunk, state):
                yield event
        emitted_closing = True
        for event in closing_events():
            yield event
    except GeneratorExit:
        # Consumer abandoned the stream; yielding here is illegal and no one is left to receive usage.
        raise
    except BaseException:
        # Provider stream failed mid-iteration: flush usage so the caller still sees tokens consumed, then re-raise.
        if not emitted_closing:
            for event in closing_events():
                yield event
        raise
```

Two things worth a team decision:
1. Whether the failure path should emit `message_stop` at all, since Anthropic semantics reserve it for clean completion. If not, emit only the usage-bearing `message_delta` on the error path.
2. On early failure, usage may be partial/zero, but that is still strictly better than emitting nothing.

tests/unit/test_messages.py
```diff
@@ -396,6 +396,70 @@ async def mock_stream() -> Any:
...
+    assert not isinstance(result, (MessageResponse, ParsedMessage))
+
+    events = [event async for event in result]
+    delta = next(e for e in events if isinstance(e, MessageDeltaEvent))
+    assert delta.usage.input_tokens == 100
+    assert delta.usage.output_tokens == 50
+    assert delta.usage.cache_read_input_tokens == 80
+    assert delta.delta.stop_reason == "end_turn"
+    assert events[-1].type == "message_stop"
+
```
tbille, line 460: The new tests only cover the success path. Please add coverage for the failure path that this PR's requirement depends on:

1. A stream that yields some content chunks (populating `state` usage) and then **raises** mid-iteration should still emit a `message_delta` carrying the accumulated usage before the exception propagates.
2. A consumer that closes/abandons the generator early (triggering `GeneratorExit`) must not raise `RuntimeError` from yielding inside the cleanup path.

Sketch for (1):

```python
async def mock_stream() -> Any:
    yield ChatCompletionChunk(
        id="c1", model="gpt-4", created=0, object="chat.completion.chunk",
        choices=[ChunkChoice(index=0, delta=ChoiceDelta(content="Hi"), finish_reason=None)],
        usage=CompletionUsage(prompt_tokens=100, completion_tokens=10, total_tokens=110),
    )
    raise RuntimeError("provider stream dropped")

result = await AnyLLM._amessages(mock_provider, params)
seen: list[Any] = []
with pytest.raises(RuntimeError):
    async for event in result:
        seen.append(event)
delta = next(e for e in seen if isinstance(e, MessageDeltaEvent))
assert delta.usage.input_tokens == 100
assert delta.usage.output_tokens == 10
```

tbille (changes requested):
Thanks for this fix. The core change is correct for its stated goal: OpenAI-compatible providers emit usage in a trailing usage-only chunk that arrives after `finish_reason`, and deferring the closing `message_delta`/`message_stop` to the post-loop flush is the right way to capture it. The `cache_read_input_tokens` mapping and the `state.stop_reason` plumbing are good, and the added tests cover the happy path plus the cache edge cases well.

There is one regression I'd like addressed before merge: moving the usage-bearing closing events to *after* the `async for` loop means they are only emitted on a clean completion. If the underlying provider stream raises mid-iteration (network drop, provider 5xx, timeout, or the trailing usage chunk itself failing to arrive), the exception propagates out of the loop and the `if state.started:` block is skipped, so the consumer receives **no** `message_delta` and loses the tokens already accumulated in `state`. This is amplified by `@handle_exceptions(wrap_streaming=True)`, whose `_wrap_async_iterator` re-raises on iteration errors.

Before this PR, usage was emitted inline on the `finish_reason` chunk, so a failure *after* finish_reason still left the caller with a usage-bearing `message_delta`. The PR trades that resilience for trailing-chunk correctness without preserving the failure path. We should guarantee tokens are reported even when the stream fails. Inline suggestions below.

Syrunekai:
Thanks @tbille — confirmed the regression exactly as you traced it: with the closing events moved after the loop, a mid-iteration raise skipped the `if state.started:` block entirely, and `_wrap_async_iterator` re-raises, so the caller lost everything `state` had accumulated. Agreed it needed addressing before merge.

Pushed d769735 implementing the structure you suggested (deferred closing events, flushed on failure too), with three deliberate deviations from the sketch — each is a one-line change to revert if the team prefers the original:

1. **`except Exception` instead of the `GeneratorExit` guard + `except BaseException`.** `GeneratorExit` and `asyncio.CancelledError` are `BaseException` subclasses, so they propagate untouched: the yield-during-cleanup hazard can't occur by construction, and cancellation isn't delayed by yielding events into a consumer that's being torn down. It also matches the `Exception`-only catch in `handle_exceptions._wrap_async_iterator`, which wraps this same stream one level up.

2. **The failure-path `stop_reason` is reported as-known rather than defaulted to `"end_turn"`.** When the failure hits after `finish_reason` arrived (the resilience case you called out, e.g. the trailing usage chunk request dying), `state.stop_reason` is set and the real value is reported, same as the sketch. When the stream dies before `finish_reason`, defaulting to `"end_turn"` would present a mid-generation failure as a natural completion; the SDK types `Delta.stop_reason` as `Optional`, and `None` is the accurate value there. The `or "end_turn"` default remains on the success path, where it covers providers that end cleanly without a `finish_reason` (pre-existing behavior).

3. **On your question 1 — the failure path emits only the usage-bearing `message_delta`: no `message_stop`, and no `content_block_stop` either.** Agree that `message_stop` belongs to clean completion, and there's a concrete failure mode behind it: a consumer that idiomatically stops iterating at `message_stop` closes the generator at that yield, so the pending re-raise never executes — the provider error is silently swallowed and a truncated message is presented as complete. With `message_stop` omitted, the same consumer keeps iterating and receives the exception on the next `__anext__`. `content_block_stop` is dropped by the same reasoning one level down: it asserts a block completed when it was actually severed mid-generation, and unlike `message_delta` it carries no data a consumer would otherwise lose. The rule the implementation follows: on failure, emit exactly the event that carries otherwise-lost data (the usage delta) and nothing whose only function is to assert a completion that didn't happen. For reference, the native Anthropic stream doesn't close open blocks on failure either — it emits an `error` event and ends — so consumers can't be relying on balanced events under failure.

Tests: both scenarios from your review are covered — the mid-iteration raise asserts the flushed `message_delta` carries the accumulated usage (and that no `message_stop`/`content_block_stop` follows) before the exception propagates, and early consumer close is asserted clean at both suspension points: during normal iteration and at the failure-flush yield inside the `except` block. Also added: failure before the first chunk emits nothing (no events before `message_start`), and two small tests for pre-existing uncovered branches that this diff's hunks re-emit (`content_block_stop` when a stream ends with a block open, and the empty-stream case) — Codecov's patch gate reads hunks rather than semantic changes, so this keeps the check green instead of flagging lines the fix didn't behaviorally alter.

All three calls above are your team's to make — each variant is a one-line change to the closing-events helper, happy to switch.

dpoulopoulos (dismissed):

tbille (dismissed):
Thanks for the changes @Syrunekai 
Apologies for the delay, merging it and will trigger a release.

src/any_llm/any_llm.py
```diff
@@ -785,29 +786,35 @@ async def _amessages(
...
+
+            try:
+                async for chunk in result:
+                    for event in chat_completion_chunk_to_message_stream_events(chunk, state):
+                        yield event
+            except Exception:
+                # Flush the usage accumulated so far before re-raising, so a mid-stream failure still reports tokens.
+                if state.started:
+                    yield usage_delta(state.stop_reason)
+                raise
```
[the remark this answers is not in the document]
tbille: Valid finding, fixed in 4757575. The failure path now flushes usage with `stop_reason=None` unconditionally (via `usage_delta(None)`), so a `finish_reason` seen before the failure can no longer be mistaken for a successful completion. `message_stop` remains reserved for the clean-completion path.

Added regression test `test_default_amessages_streaming_failure_after_finish_reason_reports_no_stop_reason`: the stream emits `finish_reason="stop"` then raises before the trailing usage chunk, and asserts the flushed `message_delta` reports `stop_reason is None` (with usage still preserved) and no `message_stop` is emitted.

Note: `content_block_stop` does legitimately appear in this scenario because the content block is closed when the `finish_reason` chunk is processed during normal iteration, so the test asserts only on the absence of `message_stop`, not `content_block_stop`.

tbille (approved):

PR #1152, opened 2026-06-30
feat(ollama): support reasoning_effort via leveled thinking

jonigl:
## Description

Passing **reasoning_effort** (low/medium/high) used to leave the value in the options dict forwarded to the **Ollama** server; the provider only ever sent a boolean `think`, ignoring Ollama's native thinking levels.

Map reasoning_effort onto Ollama's leveled `think` parameter (mirrors the anthropic provider): minimal/low->low, medium->medium, high/xhigh/max->high, none->think=False, auto/None->unset (per-model default). Exclude reasoning_effort from the params dump so it can never leak into `options`.

Bump ollama>=0.6.0 (first release typing `think` as bool | low/medium/high).

Add unit tests covering the level mapping, none, auto and the streaming path; the existing integration reasoning tests now pass for ollama.

## PR Type

- 🆕 New Feature

## Relevant issues

N/A

## Checklist

- [x] I understand the code I am submitting.
- [x] I have added unit tests that prove my fix/feature works
- [x] I have run this code locally and verified it fixes the issue.
- [x] New and existing tests pass locally
- [x] Documentation was updated where necessary
- [x] I have read and followed the [contribution guidelines](https://github.com/mozilla-ai/any-llm/blob/main/CONTRIBUTING.md)
- [ ] **AI Usage:**
    - [ ] No AI was used.
    - [x] AI was used for drafting/refactoring.
    - [ ] This is fully AI-generated.

## AI Usage Information

- AI Model used: Claude Opus 4.8 and Claude Opus 4.6
- AI Developer Tool used: Claude Code and Visual Studio Code with GitHub Copilot built in extension
- Any other info you'd like to share: AI assisted with code drafting and tests; design decisions and final review are mine.

When answering questions by the reviewer, please respond yourself, do not copy/paste the reviewer comments into an AI system and paste back its answer. We want to discuss with you, not your AI :)

- [ ] I am an AI Agent filling out this form (check box if true)

## Summary by CodeRabbit

* **New Features**
  * Ollama requests now support a wider range of reasoning effort settings, including direct mapping to the provider’s available thinking levels.
  * Streaming requests now pass the thinking setting correctly for more consistent behaviour.

* **Bug Fixes**
  * Improved handling of reasoning effort values so “none” disables thinking, while “auto” and unset values leave it unchanged.
  * Updated the minimum supported Ollama version for installations that enable this optional integration.

PR #1153, opened 2026-06-30
fix(anthropic,gemini): preserve provider-specific reasoning signatures across turns

### LINKED ISSUE #1149: [BUG] Missing specific provider thought signature
### Describe the bug

When using reasoning models providers usually provide an encrypted representations of the model's internal reasoning:
- Gemini API exposes a `Part.thought_signature`
- Anthropic API exposes `SignatureDelta.signature` 

Those fields need to be persist across turns in multi-turn conversation. 

However i can't find anywhere in the source code where such properties are passed from the provider models to the  `any-llm-sdk` models: 

In `providers/gemini/utils.py`, we don't lookup for though_signature:

```python
       elif function_call := part.function_call:
            args_dict = {}
            if args := function_call.args:
                for key, value in args.items():
                    args_dict[key] = value


            tool_calls_list.append(
                ChoiceDeltaToolCall(
                    index=len(tool_calls_list),
                    id=f"call_{hash(function_call.name)}_{len(tool_calls_list)}",
                    type="function",
                    function=ChoiceDeltaToolCallFunction(
                        name=function_call.name,
                        arguments=json.dumps(args_dict),
                    ),
                )
            )
        elif part.text:
            content += part.text
```

In `providers/antropic/utils.py`, we don't  check for a SignatureDelta:

```python
    elif isinstance(chunk, ContentBlockDeltaEvent):
        if chunk.delta.type == "text_delta":
            delta = {"content": chunk.delta.text}
        elif chunk.delta.type == "input_json_delta":
            delta = {
                "tool_calls": [
                    {
                        "index": chunk.index,
                        "function": {"arguments": chunk.delta.partial_json},
                    }
                ]
            }
        elif chunk.delta.type == "thinking_delta":
            delta = {"reasoning": {"content": chunk.delta.thinking}}
```

I'm currently observing wierd thinking loops and suspect this a cause.

The SDK should preserve provider-specific reasoning metadata. We could introduce a provider_specific_fields into the any-llm internal `ChoiceDelta` (`types/chat/chat_completion_chunk.py`) ?

#### Versions
**`any-llm` version:** 1.18.0
**Python Version**: 3.12.11

tbille:
## Description

Both Gemini and Anthropic encrypt their internal extended-reasoning trace and
expose it via a "signature" that the API requires to be replayed unmodified
on subsequent turns:

- Gemini: `Part.thought_signature`
- Anthropic: `ThinkingBlock.signature` / `SignatureDelta.signature` (streaming)

Neither value was being captured in any-llm's **streaming** conversion path.
Anthropic's signature additionally was not captured at all in the
**non-streaming** path, and was never replayed on outgoing requests, so any
multi-turn conversation using extended thinking + tool calls lost the
original reasoning trace on the next turn (matching the "weird thinking
loops" symptom reported in the issue).

### Changes

- `any_llm/types/completion.py`: add `extra_content` to `ChatCompletionMessage`
  and `ChoiceDelta` (mirroring the existing tool-call-level `extra_content`)
  to carry provider-specific, message-level metadata. Add a
  `ChoiceDeltaToolCall` subclass with `extra_content` so streaming tool-call
  deltas can carry the same per-call metadata as the non-streaming
  `ChatCompletionMessageFunctionToolCall` (e.g. Gemini's `thought_signature`).
- `gemini/utils.py`: capture `thought_signature` on streaming function-call
  deltas, matching the existing non-streaming behavior.
- `anthropic/utils.py`:
  - capture the thinking block `signature` in both streaming
    (`signature_delta` event) and non-streaming responses.
  - replay it as an Anthropic `thinking` content block when continuing a
    conversation that used extended thinking, both alongside a tool call and
    in plain assistant turns, per [Anthropic's docs on preserving thinking
    blocks](https://docs.claude.com/en/docs/build-with-claude/extended-thinking#preserving-thinking-blocks).

## PR Type

- 🐛 Bug Fix

## Relevant issues

Fixes #1149

## Checklist
- [x] **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

- AI Developer Tool used: OpenCode
- Any other info you'd like to share: Implemented end-to-end (analysis, fix, and tests) by an AI agent based on the linked issue.

## Summary by CodeRabbit

* **New Features**
  * Preserved provider-specific metadata across chat turns, including Anthropic extended-thinking signatures and Gemini tool-call thought signatures.
  * Streaming and final completion outputs now expose an `extra_content` field so this metadata is available to clients consistently.
* **Bug Fixes**
  * Improved Anthropic extended-thinking replay for both text and tool-call outputs, including correct handling when signatures are missing, empty, or malformed.
* **Tests**
  * Expanded unit coverage for Anthropic and Gemini message and streaming conversions, including new signature/delta cases.

### DIFF
--- a/src/any_llm/providers/anthropic/utils.py
+++ b/src/any_llm/providers/anthropic/utils.py
@@ -57,6 +57,45 @@ def _is_tool_call(message: dict[str, Any]) -> bool:
     return message["role"] == "assistant" and message.get("tool_calls") is not None
 
 
+def _extract_anthropic_thinking_signature(message: dict[str, Any]) -> str | None:
+    """Extract the encrypted thinking signature stored on a message's extra_content, if any."""
+    extra_content = message.get("extra_content")
+    if isinstance(extra_content, dict) and isinstance(anthropic_extra := extra_content.get("anthropic"), dict):
+        signature = anthropic_extra.get("signature")
+        if isinstance(signature, str):
+            return signature
+    return None
+
+
+def _extract_reasoning_text(message: dict[str, Any]) -> str:
+    """Extract the plain-text reasoning content from a message, regardless of its shape.
+
+    ``reasoning`` may be a plain string (the OpenAI-wire-compatible serialized form) or a
+    ``{"content": str}`` dict, depending on how the caller constructed the message.
+    """
+    reasoning = message.get("reasoning")
+    if isinstance(reasoning, str):
+        return reasoning
+    if isinstance(reasoning, dict) and isinstance(content := reasoning.get("content"), str):
+        return content
+    return ""
+
+
+def _build_anthropic_thinking_block(message: dict[str, Any]) -> dict[str, Any] | None:
+    """Reconstruct an Anthropic ``thinking`` content block for replay across turns.
+
+    When extended thinking is enabled, Anthropic requires the ``thinking`` block (including
+    its encrypted ``signature``) to be passed back unmodified on subsequent turns, e.g.
+    alongside tool results. Without it, the model loses its original reasoning trace, which
+    can lead to degraded or repeated reasoning. See
+    https://docs.claude.com/en/docs/build-with-claude/extended-thinking#preserving-thinking-blocks
+    """
+    signature = _extract_anthropic_thinking_signature(message)
+    if signature is None:
+        return None
+    return {"type": "thinking", "thinking": _extract_reasoning_text(message), "signature": signature}
+
+
 def _convert_content_for_anthropic(content: list[dict[str, Any]]) -> list[dict[str, Any]]:
     """Convert content blocks from OpenAI format to Anthropic format.
     - Parse the "content" field block by block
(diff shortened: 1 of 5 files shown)


tbille:
Addressed the CodeRabbit review comment: guarded against `content=None` (and added explicit list handling) when replaying the thinking block for a reasoning-only assistant turn, plus added coverage for the remaining new branches (non-string signature, missing reasoning text, list-shaped content, empty-string content, empty non-streaming signature).

Note: the `run-linter` CI failure is unrelated to this PR — it's a pre-existing `pre-commit-uv` environment bug (`AssertionError: BUG: expected environment for python to be healthy immediately after install`) that's also failing on `main` (e.g. #1152) and already has a fix in flight via #1158 (bump pre-commit-uv 4.2.1 → 4.2.2). `uv run pre-commit run --all-files` passes cleanly locally on this branch.

tbille:
Addressed CodeRabbit's latest nitpick: extracted a shared `_thought_signature_extra_content` helper in `gemini/utils.py` so the streaming and non-streaming thought_signature encoding logic can't drift apart (previously duplicated identically in both places).

Also: the branch has since been rebased onto `main`, which now includes #1158 (pre-commit-uv bump), so the previously-unrelated `run-linter` CI flake is resolved. All checks are now green.

dpoulopoulos (approved):

PR #1162, opened 2026-07-01
Enable streaming when response_format is set for gemini, openai, and …

liukidar:
## Description

Enables streaming chat completions with structured output configuration for providers that support it.

This removes stale `stream + response_format` rejections for Anthropic and Gemini, and updates OpenAI chat completions so structured output can be requested while streaming since it is now supported by all three providers. For OpenAI Pydantic/dataclass response formats, streaming requests convert the type to a JSON schema `response_format` before calling `chat.completions.create(stream=True)`, preserving chunk streaming without introducing final-response parsing, similar to the other providers.

This PR does not change the streaming return contract: streaming still returns chunks only, with no partial or final parsing inside any-llm (since there is no explicit delta accumulator within the library code). This means that parsing, both for pydantic and json schema response formats, will have do be carried out externally.

Note that other providers and/or endpoints likely suffer from the same limitation, but i haven't checked those.

## PR Type

- 🐛 Bug Fix

## Relevant issues

N/A

## Checklist

- [ ] **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

- AI Model used: GPT-5
- Any other info you'd like to share: Used AI assistance to inspect provider behavior, update implementation, and add/adjust unit tests.

## Summary by CodeRabbit

* **New Features**
  * Streaming requests now support structured response formats across OpenAI, Anthropic, and Gemini, including Pydantic-based schemas.

* **Bug Fixes**
  * Improved handling of tool definitions with missing or empty parameters when converting tool schemas.
  * Removed overly strict validation that previously blocked streaming when structured response formats were provided.

* **Tests**
  * Updated unit tests to confirm streamed request payloads are generated and forwarded correctly for each provider.

### DIFF
--- a/src/any_llm/providers/openai/base.py
+++ b/src/any_llm/providers/openai/base.py
@@ -82,17 +82,18 @@ class BaseOpenAIProvider(AnyLLM):
         Structured-output types are converted to JSON schema dicts when streaming, or
         when using plain dataclasses, since the OpenAI SDK's ``.parse()`` helper only
         supports non-streaming calls and Pydantic BaseModel types.
+        """
+        converted_params = params.model_dump(exclude_none=True, exclude={"model_id", "messages"})
         if is_structured_output_type(params.response_format) and (
             params.stream or not issubclass(params.response_format, BaseModel)
         ):
-            params.response_format = {
+            converted_params["response_format"] = {
                 "type": "json_schema",
                 "json_schema": {
                     "name": params.response_format.__name__,
                     "schema": get_json_schema(params.response_format),
                 },
             }
-        converted_params = params.model_dump(exclude_none=True, exclude={"model_id", "messages"})
         converted_params.update(kwargs)
 
         if "max_tokens" in converted_params:
--- a/src/any_llm/providers/openai/base.py
+++ b/src/any_llm/providers/openai/base.py
@@ -79,9 +79,9 @@ class BaseOpenAIProvider(AnyLLM):
         current OpenAI spec.  Providers whose API does not accept
         ``max_completion_tokens`` should override this method to remap back.
 
-        Plain dataclasses are converted to JSON schema dicts since the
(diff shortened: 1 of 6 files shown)


src/any_llm/providers/openai/base.py
```diff
@@ -82,7 +82,9 @@ def _convert_completion_params(params: CompletionParams, **kwargs: Any) -> dict[
         Plain dataclasses are converted to JSON schema dicts since the
         OpenAI SDK's ``.parse()`` only supports Pydantic BaseModel types.
         """
```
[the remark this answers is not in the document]
liukidar: Happy to change this if considered important. let me know

src/any_llm/providers/openai/base.py
```diff
@@ -82,7 +82,9 @@ def _convert_completion_params(params: CompletionParams, **kwargs: Any) -> dict[
...
-        if is_structured_output_type(params.response_format) and not issubclass(params.response_format, BaseModel):
+        if is_structured_output_type(params.response_format) and (
+            params.stream or not issubclass(params.response_format, BaseModel)
+        ):
             params.response_format = {
```
dpoulopoulos: `_convert_completion_params` reassigns `params.response_format` to a dict. Real-world impact is low because `CompletionParams` is built fresh per public call (`any_llm.py:602`, `any_llm.py:780`) and the base layer's post-processing reads its own response_format local rather than params.

But a caller invoking `_acompletion` directly and reusing one `CompletionParams` for a streaming then a non-streaming call would silently lose the `.parse()` path. Prefer building a local variable over mutating the input

dpoulopoulos (commented):
@liukidar thanks for that! I left a comment. If you can fix this + the copilot comment about the docstring we'll be good to go!

src/any_llm/providers/openai/base.py
liukidar: This should be fixed now.

dpoulopoulos (approved):

PR #1163, opened 2026-07-07
fix(openai): handle None created field in streaming responses

kissxu:
## Problem

Some OpenAI-compatible APIs return `"created": null` in streaming chunks instead of an integer timestamp. When this happens, `int(response.created)` at two locations crashes with:

```
TypeError: int() argument must be a string, a bytes-like object or a real number, not NoneType
```

The existing comments mention Sambanova returning float — the non-int case was anticipated, but `None` was missed.

## Fix

Guard `int(response.created)` with a `None` check. When `created` is `None`, fall back to `int(time.time())`.

Other providers already handle this pattern:
- `together/utils.py`: `created or int(datetime.now().timestamp())`
- `sagemaker/utils.py`: `response.get("created", int(time()))`
- `lmstudio/utils.py`: `created=int(time.time())`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of missing response timestamps, falling back to the current time when needed.
  * Made chat and completion response parsing more resilient to incomplete timestamp data.

### DIFF
--- a/src/any_llm/providers/openai/base.py
+++ b/src/any_llm/providers/openai/base.py
@@ -3,6 +3,7 @@ import json
 from collections.abc import AsyncIterator, Sequence
 from io import BytesIO
 from pathlib import Path
+import time
 from typing import Any, Literal, cast
 
 from openai import AsyncOpenAI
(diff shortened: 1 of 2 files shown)


kissxu:
Misconfiguration: we had --backend anyllm unnecessarily, which made us hit a bug that doesn't exist in normal use. Closing as not-a-bug.

PR #1178, opened 2026-07-15
fix: allow streaming structured output where supported

### LINKED ISSUE #1166: Streaming with response_format still unsupported on Groq, Cohere, Cerebras, and Otari (cross-provider inconsistency after #1162)
### Context

#1162 enabled streaming when `response_format` is set for the OpenAI, Gemini, and Anthropic completion endpoints. After that change, the same streaming-with-structured-output call succeeds on those providers but continues to raise on four others, leaving an inconsistent cross-provider surface for identical user code.

### Providers that still raise on `stream` + `response_format`

| Provider | Location | Behavior |
| --- | --- | --- |
| Groq | `src/any_llm/providers/groq/groq.py:122` (`_stream_async_completion`) | raises `UnsupportedParameterError` |
| Cohere | `src/any_llm/providers/cohere/cohere.py:200` (`_acompletion`) | raises `UnsupportedParameterError` |
| Cerebras | `src/any_llm/providers/cerebras/cerebras.py:107` (`_stream_completion_async`) | raises `UnsupportedParameterError` |
| Otari | `src/any_llm/providers/otari/otari.py:303` (`_acompletion`) | raises `ValueError("stream is not supported for response_format")` |

### Note on Otari

`OtariProvider` subclasses `BaseOpenAIProvider` (`otari.py:154`), which now supports streaming with `response_format` after #1162. Because Otari overrides `_acompletion` and re-adds its own guard, it does **not** inherit the fix. Worth checking whether the guard is still necessary or whether Otari can defer to the base implementation.

### Suggested next steps

- For each provider, determine whether the upstream SDK/API actually supports streaming structured output.
- Where supported: remove the guard, implement streaming structured output, and add unit + integration tests so the behavior matches OpenAI/Gemini/Anthropic.
- Where genuinely unsupported upstream: keep the guard but document the limitation (provider docs / capability metadata) so the divergence is intentional and discoverable.

Follow-up to #1162.

zzz100868:
## Description

The Cohere, Cerebras, and Otari providers rejected `stream=True` whenever `response_format` was set, before the request reached their SDKs. Their current streaming APIs support structured output, so this removes those local guards and keeps each provider's existing response format conversion.

Groq still rejects this combination because its upstream API does not support streaming structured output. The guard remains, with a link to the upstream limitation.

## PR Type

- 🐛 Bug Fix

## Relevant issues

Fixes #1166

## Testing

- `uv run pytest -q --reruns 0 tests/unit/providers/test_cohere_provider.py tests/unit/providers/test_cerebras_provider.py tests/unit/providers/test_groq_provider.py tests/unit/providers/test_otari_provider.py` (101 passed)
- Pre-commit checks on all changed files (passed)
- The full unit suite reached 1357 passed and 121 skipped. The remaining 71 tests require unrelated optional provider extras that were not installed locally.

## Checklist

- [x] **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

- AI Model used: GPT-5
- Any other info you'd like to share: The change was verified against the affected provider unit tests and the repository's pre-commit checks.

## Summary by CodeRabbit

* **New Features**
  * Structured response formats can now be used with streaming completions across supported providers, where the underlying service permits it.
  * Response format settings are forwarded correctly during streamed requests.

* **Bug Fixes**
  * Removed incorrect early failures when combining streaming with structured output.

* **Tests**
  * Added coverage confirming streamed structured-output requests are handled correctly.

### DIFF
--- a/src/any_llm/providers/cerebras/cerebras.py
+++ b/src/any_llm/providers/cerebras/cerebras.py
@@ -5,7 +5,6 @@ from typing import TYPE_CHECKING, Any, cast
 from typing_extensions import override
 
 from any_llm.any_llm import AnyLLM
-from any_llm.exceptions import UnsupportedParameterError
 from any_llm.utils.structured_output import get_json_schema, is_structured_output_type
 
 MISSING_PACKAGES_ERROR = None
@@ -104,10 +103,6 @@ class CerebrasProvider(AnyLLM):
         messages: list[dict[str, Any]],
         **kwargs: Any,
     ) -> AsyncIterator[ChatCompletionChunk]:
-        if kwargs.get("response_format", None) is not None:
-            msg = "stream and response_format"
-            raise UnsupportedParameterError(msg, self.PROVIDER_NAME)
-
         cerebras_stream = await self.client.chat.completions.create(
             model=model,
             messages=messages,
--- a/src/any_llm/providers/cohere/cohere.py
+++ b/src/any_llm/providers/cohere/cohere.py
@@ -197,9 +197,6 @@ class CohereProvider(AnyLLM):
     ) -> ChatCompletion | AsyncIterator[ChatCompletionChunk]:
         if params.response_format is not None:
             kwargs["response_format"] = self._preprocess_response_format(params.response_format)
-        if params.stream and params.response_format is not None:
-            msg = "stream and response_format"
-            raise UnsupportedParameterError(msg, self.PROVIDER_NAME)
         if params.parallel_tool_calls is not None:
             msg = "parallel_tool_calls"
             raise UnsupportedParameterError(msg, self.PROVIDER_NAME)
--- a/src/any_llm/providers/groq/groq.py
+++ b/src/any_llm/providers/groq/groq.py
@@ -119,6 +119,8 @@ class GroqProvider(AnyLLM):
     async def _stream_async_completion(
         self, params: CompletionParams, **kwargs: Any
     ) -> AsyncIterator[ChatCompletionChunk]:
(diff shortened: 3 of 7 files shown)


njbrake (approved):

Approving. This is a tightly-scoped fix for the cross-provider inconsistency in #1166: remove the `stream + response_format` guards where the upstream actually supports streaming structured output (Cohere, Cerebras, Otari), and keep + document the guard where it does not (Groq).

I verified each per-provider decision against current docs rather than the PR description:
- Cohere: `chat_stream` supports `response_format` (json_object / json_schema) on Command R/R+ and newer. Removal justified.
- Cerebras: streaming is supported alongside structured outputs (JSON mode). Removal justified.
- Groq: docs explicitly state streaming and tool use are not supported with Structured Outputs. Keeping the guard is correct, and the added comment links the limitation.
- Otari: subclasses the OpenAI base and inherits the streaming path enabled in #1162; the rewritten guard keeps `stream=True` only when structured output is not being buffered.

Locally: 101 provider unit tests pass, mypy (strict) clean, ruff clean.

Two non-blocking follow-ups:
1. The Cerebras and Cohere tests use a plain `{"type": "json_object"}` dict, so they do not exercise the Pydantic to json_schema conversion in the streaming branch. The Otari test does. Low risk since the conversion sits above the stream branch and is shared with the already-tested non-stream path, but a streaming test with a Pydantic model would close the gap.
2. Only unit tests ran here. Since this changes real cross-provider request behavior, a run of the `run-integration-tests` label would be good confirmation before or shortly after merge.

PR #1180, opened 2026-07-15
fix(messages): request usage on streaming so the bridge meters tokens

njbrake:
## Description

Streaming through the Messages to Completions bridge did not set `stream_options.include_usage`, so OpenAI-compatible backends omitted token usage from their streamed chunks. The recent trailing-usage-chunk fix (#1151) then had no usage-only chunk to flush, so streamed `amessages` against those providers reported zero input and output tokens (and no cache). Native Anthropic was unaffected, since it overrides `_amessages` and streams usage directly.

This requests `stream_options.include_usage` in the streaming branch of `messages_params_to_completion_params`, so the backend emits the trailing usage-only chunk that the stream wrapper flushes into the closing `message_delta`. It completes the chain #1151 started: #1151 captures the trailing chunk, this makes sure the trailing chunk is actually produced.

Safe across providers:

- Providers that do not support `stream_options` (Cerebras, Ollama, Together, Cohere, Mistral) already strip it in their own `_convert_completion_params`.
- The native Anthropic provider never reaches this bridge (it overrides `_amessages`), so its SDK never sees a `stream_options` kwarg.

Downstream context: this is the root cause of the zero-metering behavior reported in mozilla-ai/otari#256 (streamed `/v1/messages` recorded zero tokens and zero cost while non-streaming messages and streaming chat completions metered fine). The old any-llm gateway solved the same class of bug for chat completions in #974; this brings the messages path in line.

## PR Type

- 🐛 Bug Fix

## Relevant issues

Complements #1151. Root cause of mozilla-ai/otari#256.

## Checklist

- [x] **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

- Any other info you'd like to share: Drafted by Claude via back-and-forth with @njbrake. The investigation (tracing the bug across #1151, #974, and the pinned 1.17.0), the diagnosis, and the decisions are his; the code and this prose are Claude's. On the "discuss with the human, not the AI" policy: respected. @njbrake will reply to review questions himself.

### Testing

Full unit suite passes (1490 passed, 64 skipped), plus ruff and mypy strict clean over `src` and the touched tests. New tests:

- `test_messages_compat.py`: `stream_options.include_usage` is present when streaming, absent when not streaming and when `stream` is unset.
- `test_messages.py`: the `CompletionParams` handed to `_acompletion` actually carries `include_usage` on the streaming path and omits it on the non-streaming path. This is the coverage the trailing-chunk fix lacked (its tests hand-fed usage chunks, so they could not catch that usage was never requested).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved streamed completion token-usage reporting by requesting a final usage-only chunk when streaming is enabled.
  * Filtered OpenAI-specific `stream_options` from provider requests where unsupported (Watsonx, Groq, xAI, and Azure).
* **Tests**
  * Added unit tests validating streaming vs non-streaming forwarding behaviour for the messages→completion bridge.
  * Added dedicated unit tests ensuring Watsonx, Groq, xAI, and Azure conversions drop `stream_options`.

### DIFF
--- a/src/any_llm/utils/messages_compat.py
+++ b/src/any_llm/utils/messages_compat.py
@@ -70,6 +70,15 @@ def messages_params_to_completion_params(params: MessagesParams) -> dict[str, An
         result["stop"] = params.stop_sequences
     if params.stream is not None:
         result["stream"] = params.stream
+        if params.stream:
+            # OpenAI-compatible backends omit token usage from streamed chunks
+            # unless asked for it, so the streamed Messages bridge would report
+            # zero tokens. Request the trailing usage-only chunk that the
+            # streaming wrapper flushes into the closing ``message_delta``.
+            # Providers that don't support ``stream_options`` strip it in their
+            # own param conversion, and the native Anthropic provider never
+            # reaches this bridge (it overrides ``_amessages``).
+            result["stream_options"] = {"include_usage": True}
 
     if params.output_format is not None:
         if is_structured_output_type(params.output_format):
--- a/tests/unit/test_messages.py
+++ b/tests/unit/test_messages.py
@@ -9,7 +9,10 @@ import pytest
 from any_llm.any_llm import AnyLLM
 from any_llm.api import amessages
 from any_llm.types.completion import (
+    ChatCompletion,
     ChatCompletionChunk,
+    ChatCompletionMessage,
+    Choice,
     ChoiceDelta,
     ChunkChoice,
     CompletionUsage,
@@ -442,8 +445,6 @@ async def test_default_amessages_streaming_requests_include_usage() -> None:
 @pytest.mark.asyncio
 async def test_default_amessages_non_streaming_omits_include_usage() -> None:
     """A non-streaming bridge call has no usage-only chunk to request."""
-    from any_llm.types.completion import ChatCompletion, ChatCompletionMessage, Choice
-
     mock_completion = ChatCompletion(
         id="c",
         model="gpt-4",
--- a/tests/unit/test_messages.py
+++ b/tests/unit/test_messages.py
@@ -406,6 +406,73 @@ async def test_default_amessages_streaming() -> None:
 
 
 @pytest.mark.asyncio
+async def test_default_amessages_streaming_requests_include_usage() -> None:
+    """Streaming through the bridge must ask the backend for usage, otherwise
+    OpenAI-compatible providers emit no usage-only chunk and the trailing-chunk
+    capture has nothing to report."""
+
+    async def mock_stream() -> AsyncIterator[ChatCompletionChunk]:
+        yield ChatCompletionChunk(
+            id="chunk-1",
+            model="gpt-4",
+            created=0,
+            object="chat.completion.chunk",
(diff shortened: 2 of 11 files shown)


src/any_llm/utils/messages_compat.py
```diff
@@ -70,6 +70,15 @@ def messages_params_to_completion_params(params: MessagesParams) -> dict[str, An
...
         result["stream"] = params.stream
+        if params.stream:
+            # OpenAI-compatible backends omit token usage from streamed chunks
+            # unless asked for it, so the streamed Messages bridge would report
+            # zero tokens. Request the trailing usage-only chunk that the
+            # streaming wrapper flushes into the closing ``message_delta``.
+            # Providers that don't support ``stream_options`` strip it in their
+            # own param conversion, and the native Anthropic provider never
+            # reaches this bridge (it overrides ``_amessages``).
+            result["stream_options"] = {"include_usage": True}
```
khaledosman, line 81: Minimax slips through this claim: `MinimaxProvider._convert_completion_response_async` filters chunks with `if chunk.choices and chunk.choices[0].delta`, so the usage-only trailing chunk (empty `choices`) is dropped before it reaches the bridge and streamed messages through minimax still meter zero. Not a regression, but worth a follow-up to pass usage-bearing chunks through the minimax filter.

tests/unit/test_messages.py
```diff
@@ -405,6 +405,73 @@ async def mock_stream() -> Any:
...
+
+    completion_params = mock_provider._acompletion.call_args.args[0]
+    assert completion_params.stream is True
+    assert completion_params.stream_options == {"include_usage": True}
+
+
+@pytest.mark.asyncio
+async def test_default_amessages_non_streaming_omits_include_usage() -> None:
+    """A non-streaming bridge call has no usage-only chunk to request."""
+    from any_llm.types.completion import ChatCompletion, ChatCompletionMessage, Choice
```
khaledosman: `any_llm.types.completion` is not an optional dependency and the file already imports from it at the top. Per the testing guidelines, inline imports are reserved for optional provider SDKs; move `ChatCompletion, ChatCompletionMessage, Choice` into the top-level import block.

khaledosman (commented):
Reviewed the fix end-to-end: verified the bridge chain into the #1151 trailing-chunk flush, and checked the provider blast radius against actual SDK signatures (groq `create` and xai `chat.create` have no `stream_options` parameter, azure `complete()` does not model it, huggingface-hub 1.3.2 does, and cohere/together/cerebras/ollama/mistral already strip it; bedrock/sagemaker/gemini/lmstudio build kwargs explicitly and ignore it). Groq and xAI still meter streamed usage despite the strip since both surface usage natively on chunks. Touched test files, ruff check and format pass locally; CI is green.

Two minor inline comments below. One observation that fits neither diff line: every `BaseOpenAIProvider` subclass (perplexity, deepinfra, dashscope, moonshot, ...) now sends `stream_options` to its remote endpoint on the streamed messages path. The OpenAI SDK accepts it, but backend tolerance can't be verified statically; one run with the `run-integration-tests` label would derisk that.

🤖 This review was created by Claude Code.

khaledosman (dismissed):

Issue #1197, opened 2026-07-23
Provider policy: two tiers, a config registry, and a first-class OpenAI-compatible path


njbrake:
> _Note: this issue was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

## Problem

We keep receiving PRs that add OpenAI-compatible gateways as providers (#1194 Kenari,
#1187 Eden AI, #1150 Atlas Cloud, Neosantara, etc.). Each one is individually cheap:
subclass `BaseOpenAIProvider`, set `API_BASE`, env var names, and capability flags.
But the list is unbounded, and every merge makes it harder to refuse the next one
without being inconsistent.

The cost of these providers is not the code. It is:

- the endorsement signal of being listed as an official provider
- the `tests/conftest.py` model map (126 entries) rotting as gateways deprecate models
- integration test noise and the eventual dead-gateway cleanup
- review attention for PRs that are 90% boilerplate

Meanwhile the answer we would like to give ("just use the OpenAI provider with a
custom `api_base`") works mechanically but is semantically a hack: the provider
reports itself as `openai`, and `LLMProvider.from_string` rejects any name not in
the enum, so users cannot represent their gateway under its own name.

## Current state (audited)

- 51 providers. **12 are pure config** (zero method overrides, only class attributes):
  atlascloud, dashscope, databricks, deepinfra, moonshot, mzai, nebius, neosantara,
  perplexity, qiniu, telnyx, plus `openai` itself.
- ~8 more have a single small override (optional-key handling or a param remap).
- CI holds integration-test keys for **22 of 51** providers. The config-only gateway
  providers (kenari, edenai, atlascloud, requesty, telnyx, qiniu, ...) have no keys,
  so their integration tests already silently skip. Exceptions: moonshot, nebius,
  and perplexity are config-only but do have CI keys.
- Provider resolution is convention-based (`_create_provider` in `any_llm.py` builds
  the class name from the enum value and imports the module), so a registry can slot
  in without restructuring.

### 1. Two provider tiers (policy)

- **Verified providers**: we hold API keys, integration tests run in CI, we fix
  breakage. This is what we promote in the README and docs.
- **Community providers**: verified live by the contributor at PR time, maintained
  by the community. No CI keys, excluded from the integration test matrix. Docs
  label them as such. This formalizes what is already true for these providers today.

The tier is a support promise, not a code-shape distinction. A config-only provider
[...]

PR #1198, opened 2026-07-23
feat: add first-class OpenAI-compatible custom endpoint path

njbrake:
## Description

Adds a first-class, supported way to point any-llm at any OpenAI-compatible endpoint that does not have a dedicated provider, via a new factory:

```python
from any_llm import AnyLLM

llm = AnyLLM.create_openai_compatible(
    name="mygateway",
    api_base="https://mygateway.example/v1",
    api_key="your-key",  # optional for keyless local servers
)
resp = llm.completion(model="some-model", messages=[{"role": "user", "content": "Hello!"}])
```

Previously the only way to reach an unlisted gateway was `AnyLLM.create("openai", api_base=...)`, which misreports the provider identity as `openai`, and `LLMProvider.from_string` rejects any name not in the enum, so users could not represent their gateway under its own name. The returned provider reports the caller's chosen `name` and is usable exactly like any other provider instance.

Design notes:

- **Instance-based, not enum routed.** The custom endpoint is represented by a provider instance you hold and call methods on. The provider enum stays the identity registry; named `provider:model` routing for community gateways is the follow-up registry work in the parent issue.
- **Keyless tolerated** (mirrors vllm/llamafile): `api_key` falls back to `OPENAI_COMPATIBLE_API_KEY`, then to a placeholder, so local servers are never blocked by `MissingApiKeyError`.
- **Capability flags** follow the OpenAI-compatible defaults from `BaseOpenAIProvider`; calling a capability the endpoint does not implement surfaces the endpoint's own error. Explicit per-provider flags are the registry's job in a later PR.

Implementation:

- `src/any_llm/providers/openai/custom.py`: `OpenAICompatibleProvider(BaseOpenAIProvider)`
- `src/any_llm/any_llm.py`: `AnyLLM.create_openai_compatible(name, api_base, api_key=None, **kwargs)`
- `tests/unit/providers/test_openai_compatible_provider.py`: 11 tests (identity, base-URL binding, empty-api_base guard, keyless placeholder, explicit/env key precedence, capability defaults, kwargs forwarding, enum exclusion)
- `docs/quickstart.md`: "Custom OpenAI-compatible Endpoints" section

Testing:

- `pytest tests/unit/providers/test_openai_compatible_provider.py`: 11 passed
- `pre-commit run --files <changed>`: ruff, ruff-format, mypy strict, codespell all pass

## PR Type

- 🆕 New Feature

## Relevant issues

[...]

PR #1201, opened 2026-07-24
feat: add config registry for OpenAI-compatible gateway providers

njbrake:
## Description

Adds the config registry from the #1197 provider policy and migrates all config-only OpenAI-compatible providers to it: gateway providers become data rows instead of code folders.

```python
"atlascloud": OpenAICompatibleProviderConfig(
    name="atlascloud",
    api_base="https://api.atlascloud.ai/v1",
    env_api_key_name="ATLASCLOUD_API_KEY",
    env_api_base_name="ATLASCLOUD_API_BASE",
    provider_documentation_url="https://www.atlascloud.ai/docs",
    supports_completion_reasoning=True,
),
```

The loader (`AnyLLM._create_provider` and `AnyLLM.get_provider_class`) checks the registry before falling through to the existing folder-per-provider import convention. A row mints a cached `BaseOpenAIProvider` subclass whose class name follows the `{Name}Provider` convention, so `metadata.class_name` and class identity are stable across a migration.

Migrated providers (the full config-only set from #1197): atlascloud, dashscope, databricks, deepinfra, moonshot, mzai, nebius, neosantara, perplexity, qiniu, telnyx. Each folder shrinks to an import shim, and every existing per-provider unit test passes without modification, demonstrating the compatibility contract: `LLMProvider.<NAME>` enum entries, `AnyLLM.create("<name>")`, `<name>:model` strings, `get_provider_class` identity, metadata, env var resolution, and both deep-import paths all behave as before.

Design notes:

- **Rows replicate effective flags.** Each row carries the flags the folder class actually exposed, including values inherited from `BaseOpenAIProvider` defaults (for example moderation). Flag corrections are deliberately out of scope; migrations must be behavior-neutral.
- **`api_base` is optional in the row schema** for databricks, whose base URL only comes from `DATABRICKS_HOST` or an explicit argument, mirroring the folder class it replaces.
- **Community rows already route.** A registry name with no enum entry resolves through `AnyLLM.create(name)` and `get_provider_class(name)` (covered by tests using an injected row). `provider:model` string routing for registry-only names is deliberately not wired yet; `split_model_provider` returns `LLMProvider` and widening that type is its own change.
- **Conservative flag defaults for new rows.** Rows default to completion, streaming, and list_models only; everything else is opt-in per row. Capability flags are part of the row because `BaseOpenAIProvider` raises from the `SUPPORTS_*` attributes.
- **Zero-knob for now.** Any gateway needing custom auth, param remaps, or response translation keeps a code folder. Whether the registry grows one or two knobs to absorb the middle tier (~8 providers) stays an open question in #1197.

Follow-ups (tracked in #1197): two-tier docs listing, CONTRIBUTING acceptance rule, string routing plus docs presentation for community rows, and eventually retiring the import shims.

Testing:

- `pytest tests/unit`: 1570 passed, 92 skipped, including every migrated provider's existing test file unchanged (atlascloud, dashscope, databricks, deepinfra, moonshot, neosantara, perplexity, qiniu, telnyx; mzai and nebius had no test files)
- New `tests/unit/test_registry.py`: 13 tests (lookup normalization, class caching, loader precedence, enum and string resolution, metadata fidelity, community-row routing, conservative defaults, unknown-name errors, shim identity)
- `pytest tests/docs`: 4 passed
- `scripts/generate_provider_table.py` runs clean against the registry-backed metadata
- `pre-commit` (ruff, ruff-format, mypy strict, codespell) clean on changed files

## PR Type

- 🆕 New Feature

## Relevant issues

Part of #1197 (config registry and migration steps). Does not close the issue.

## Checklist
- [x] New and existing tests pass local

### DIFF
--- a/src/any_llm/any_llm.py
+++ b/src/any_llm/any_llm.py
@@ -241,7 +241,15 @@ class AnyLLM(ABC):
     def _create_provider(
         cls, provider_key: str | LLMProvider, api_key: str | None = None, api_base: str | None = None, **kwargs: Any
     ) -> AnyLLM:
-        """Dynamically load and create an instance of a provider based on the naming convention."""
+        """Dynamically load and create an instance of a provider.
+
+        Registry rows (config-only OpenAI-compatible gateways) resolve first;
+        everything else falls through to the folder-per-provider naming convention.
+        """
+        registry_class = cls._get_registry_provider_class(provider_key)
+        if registry_class is not None:
+            return registry_class(api_key=api_key, api_base=api_base, **kwargs)
+
         provider_key = LLMProvider.from_string(provider_key).value
 
         provider_class_name = f"{provider_key.capitalize()}Provider"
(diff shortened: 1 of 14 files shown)


njbrake:
> _Note: this comment was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

@tbille same for this one: review and merge whenever you're comfortable. It is stacked on #1198, so it should land second; GitHub will retarget it to main once #1198 merges.

tbille (approved):
Nice one! Let's get this merged once lint fixed

PR #1230, opened 2026-08-05
fix(openai): suppress parsed response serializer warnings

### LINKED ISSUE #1203: [BUG] OpenAI structured output emits a pydantic serializer warning on every call while dumping a discarded field
**`any-llm` version:** 1.21.0 (`any-llm-sdk`, openai 2.24.0, pydantic 2.12.5)
**Python Version**: 3.13

### Describe the bug

Every OpenAI structured-output call (pydantic `response_format`) emits a pydantic `UserWarning`:

```
UserWarning: Pydantic serializer warnings:
  PydanticSerializationUnexpectedValue(Expected `none` - serialized value may not be as expected
  [field_name='parsed', input_value=Answer(answer='Earth'), input_type=Answer])
```

The source is `_convert_chat_completion` in `providers/openai/utils.py`, which calls `response.model_dump()` on the OpenAI SDK's `ParsedChatCompletion`. Its generic `parsed` field is declared against an unparameterized TypeVar (schema default `None`), so serializing the actual parsed model instance triggers the warning.

The detail that makes this purely cosmetic to fix: the dumped `parsed` value is never used. `_convert_parsed_chat_completion` discards it and re-attaches `parsed` from the original response objects immediately after:

```python
def _convert_parsed_chat_completion(response: OpenAIParsedChatCompletion[Any]) -> ParsedChatCompletion[Any]:
    """Convert an OpenAI ParsedChatCompletion preserving the .parsed field on each choice."""
    base = _convert_chat_completion(response)  # <- model_dump() here warns about `parsed`
    parsed_completion: ParsedChatCompletion[Any] = ParsedChatCompletion.model_validate(base, from_attributes=True)
    for base_choice, parsed_choice in zip(response.choices, parsed_completion.choices, strict=True):
        parsed_choice.message.parsed = base_choice.message.parsed  # <- dumped value replaced anyway
    return parsed_completion
```

So the warning is emitted for serializing a field whose serialized form is thrown away.

Impact: log noise once per process in production, and a hard failure for any consumer running pytest with `filterwarnings = ["error"]` (we carry an `"ignore:Pydantic serializer warnings:UserWarning"` entry solely for this).

### Example w/ Code

```python
import asyncio
import warnings

from pydantic import BaseModel

from any_llm import acompletion


class Answer(BaseModel):
    answer: str


warnings.simplefilter("error")  # same effect as pytest filterwarnings = ["error"]

asyncio.run(
    acompletion(
        model="openai:gpt-4o-mini",
        messages=[{"role": "user", "content": "What planet do we live on? One word."}],
        response_format=Answer,
    )
)
# UserWarning: Pydantic serializer warnings: PydanticSerializationUnexpectedValue(
#   Expected `none` - serialized value may not be as expected [field_name='parsed', ...])
```

### Suggested fix

Silence the dump for the value that is discarded anyway, e.g. `response.model_dump(warnings=False)` in `_convert_chat_completion` (or null out `choice.message.parsed` on a copy before dumping). No behavior change either way, since `parsed` is re-attached from the original objects.

Happy to send the PR.

mikemikimike:
## Description

Suppress the Pydantic serializer warning emitted when converting parsed OpenAI Chat Completions. The OpenAI SDK can construct a parsed response with an unbound TypeVar schema even though the parsed value is valid; passing `warnings=False` to `model_dump` avoids a warning for this discarded serialization while preserving the parsed value. Adds an offline regression test using the SDK's real `parse_chat_completion` path.

## PR Type

- [x] Bug Fix

## Relevant issues

Fixes #1203

## Checklist

- [ ] New and existing tests pass locally (focused tests pass; full unit suite timed out locally)
- [ ] **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

- AI Model used: GPT-5
- Any other info you'd like to share: The change and regression test were reviewed against the issue's offline reproduction.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of parsed OpenAI chat-completion responses.
  * Prevented unnecessary serialisation warnings during response processing.
  * Ensured parsed answers remain intact when converted.

* **Tests**
  * Added coverage for parsed responses and warning-free conversion.

### DIFF
--- a/src/any_llm/providers/openai/utils.py
+++ b/src/any_llm/providers/openai/utils.py
@@ -79,7 +79,7 @@ def _convert_chat_completion(response: OpenAIChatCompletion) -> ChatCompletion:
             type(response.created),
         )
         response.created = int(response.created)
-    normalized = _normalize_openai_dict_response(response.model_dump())
+    normalized = _normalize_openai_dict_response(response.model_dump(warnings=False))
     return ChatCompletion.model_validate(normalized)
 
 
--- a/tests/unit/providers/test_openai_utils.py
+++ b/tests/unit/providers/test_openai_utils.py
@@ -1,11 +1,46 @@
+import warnings
+
 import pytest
+from openai.lib._parsing._completions import parse_chat_completion
 from openai.types.chat.chat_completion import ChatCompletion as OpenAIChatCompletion
 from openai.types.chat.chat_completion_chunk import ChatCompletionChunk as OpenAIChatCompletionChunk
+from pydantic import BaseModel
 
 from any_llm.exceptions import ProviderError
 from any_llm.providers.openai.base import BaseOpenAIProvider
-from any_llm.providers.openai.utils import _convert_chat_completion
+from any_llm.providers.openai.utils import _convert_chat_completion, _convert_parsed_chat_completion
 from any_llm.types.completion import ChatCompletion, ChatCompletionChunk
+
+
+class Answer(BaseModel):
+    answer: str
+
+
+def test_convert_parsed_chat_completion_does_not_emit_serializer_warning() -> None:
+    """The SDK's parsed response path must not warn while preserving parsed data."""
+    raw = OpenAIChatCompletion.model_validate(
+        {
+            "id": "chatcmpl-offline",
+            "object": "chat.completion",
+            "created": 0,
+            "model": "gpt-4o-mini",
+            "choices": [
+                {
+                    "index": 0,
+                    "finish_reason": "stop",
+                    "message": {"role": "assistant", "content": '{"answer":"Earth"}'},
+                }
+            ],
+        }
+    )
+    parsed = parse_chat_completion(response_format=Answer, input_tools=[], chat_completion=raw)
+
+    with warnings.catch_warnings(record=True) as caught:
+        warnings.simplefilter("always")
+        converted = _convert_parsed_chat_completion(parsed)
+
+    assert not any("Pydantic serializer warnings" in str(item.message) for item in caught)
+    assert converted.choices[0].message.parsed == Answer(answer="Earth")
 
 
 def test_convert_chat_completion_with_empty_response() -> None:


njbrake (approved):

Approving. Nice catch, and nice test.

I reproduced the underlying mechanism rather than taking the description on faith: `client.chat.completions.parse()` builds each choice via `construct_type_unchecked(type_=ParsedChoice[ResponseFormatT], ...)` with `ResponseFormatT` still an unbound TypeVar at runtime, so the message class comes back as `ParsedChatCompletionMessage[TypeVar]` with a `None` schema for `parsed`. That is what makes the dump warn once per structured-output call. Replicating the pre-PR body emits exactly 1 `Pydantic serializer warnings` UserWarning; the post-PR path emits 0, with `message.parsed` intact in both.

Two things I checked before signing off:

1. `warnings=False` is diagnostic-only here. `parsed.model_dump()` and `parsed.model_dump(warnings=False)` produce identical dicts, so there is no behavior change. The dumped `parsed` is discarded anyway, since `_convert_parsed_chat_completion` reassigns it from the original choices right after.

2. The suppression is broader than the parsed path, since `_convert_chat_completion` handles every OpenAI-compatible provider. I built a `ChatCompletion` with a deliberately wrong `refusal=12345` (the class of provider quirk the Sambanova `created`-as-float handling exists for) and confirmed the following `ChatCompletion.model_validate(normalized)` still raises `ValidationError`. Only the redundant pre-validation hint is lost, so nothing actionable gets silenced.

This also matches what the repo already does at every other dump-then-revalidate hop (`openai/base.py`, `huggingface/huggingface.py`, `utils/structured_output.py`), so it closes the last gap rather than introducing a new pattern.

One non-blocking suggestion for a follow-up if you feel like it: the test's key assertion is negative, so if a future OpenAI SDK release parameterizes that class properly, the test would pass even with the fix reverted. Asserting the precondition first (that a plain `model_dump()` does warn, or that the message class name is `ParsedChatCompletionMessage[TypeVar]`) would pin it. I confirmed that precondition holds against the currently pinned SDK.

`uv run pytest tests/unit` is green locally (1607 passed, 92 skipped).

PR #1245, opened 2026-08-06
fix(reasoning): handle XML reasoning tags split mid-buffer

### LINKED ISSUE #1242: [BUG] Streaming XML reasoning parser fails when tags are split across chunks
**`any-llm` version:** `main` at `8f3bb1e84d82602762ef9e78ca5168bcb291eaa9` (`0.0.0.dev0`)  
**Python Version:** `3.13.14`

### Describe the bug

I was checking the streaming XML reasoning parser and found that it does not handle every valid tag split across chunks correctly.

This seems to be a bug in the existing streaming reasoning feature, not arequest for new behaviour. The docstring of `process_streaming_reasoning_chunks()` says that it buffers across chunks to detect and extract reasoning tags that may be split across multiple chunks. There is also an existing `test_wrap_chunks_handles_split_tags` test, but it only covers a partial opening tag at the beginning of the buffer.

Two other chunk boundaries currently fail.

A closing tag split after reasoning content:

```text
"<think>reasoning</th" + "ink>answer"
```

Expected result:

```text
reasoning = "reasoning"
content = "answer"
```

Actual result on `main`:

```text
reasoning = ""
content = ""
```

An opening tag split after normal content:

```text
"preface <th" + "ink>reasoning</think>answer"
```

Expected result:

```text
reasoning = "reasoning"
content = "preface answer"
```

Actual result:

```text
reasoning = ""
content = "preface <think>reasoning</think>answer"
```

This means that changing only the upstream chunk boundary can either drop the reasoning and final answer or return the raw reasoning block as visible content.

From what I could trace, `is_partial_reasoning_tag()` checks `text.startswith(tag[:i])`. That works when the partial tag starts at offset zero, but not when it is a suffix of a buffer containing earlier text. In the closing tag case, the partial fragment is also moved into `reasoning_buffer`, so the next chunk cannot complete the tag.

I verified both cases using regression tests on an untouched `main` checkout. The original 13 tests in `tests/unit/test_xml_reasoning.py` pass, while both new boundary tests fail.

### Example w/ Code

The following cases reproduce the problem using the existing helpers in `tests/unit/test_xml_reasoning.py`:

```python
@pytest.mark.asyncio
@pytest.mark.parametrize(
    ("parts", "expected_content", "expected_reasoning"),
    [
        (["<think>reasoning</th", "ink>answer"], "answer", "reasoning"),
        (
            ["preface <th", "ink>reasoning</think>answer"],
            "preface answer",
            "reasoning",
        ),
    ],
)
async def test_wrap_chunks_handles_tags_split_after_content(
    parts: list[str],
    expected_content: str,
    expected_reasoning: str,
) -> None:
    chunks = [_make_chunk(content=part) for part in parts]
    result = await _collect_chunks(
        wrap_chunks_with_xml_reasoning(_async_iter_chunks(chunks))
    )

    full_content = "".join(
        chunk.choices[0].delta.content or ""
        for chunk in result
        if chunk.choices
    )
    full_reasoning = "".join(
        chunk.choices[0].delta.reasoning.content
        for chunk in result
        if chunk.choices and chunk.choices[0].delta.reasoning
    )

    assert full_content == expected_content
    assert full_reasoning == expected_reasoning
```

I also tested a local fix that keeps the longest suffix of the buffer that is a prefix of a reasoning tag. After that change, the results were:

```text
tests/unit/test_xml_reasoning.py: 19 passed
tests/unit/providers/test_sambanova.py: 6 passed
tests/unit/providers/test_huggingface_provider.py: 15 passed
```

I can open a PR with the fix and regression tests if this approach looks okay.

njbrake:
## Description

Thanks @Vansh-Sharma27 for the precise repro in #1242, it made the patch straightforward.

`is_partial_reasoning_tag()` checked `text.startswith(tag[:i])`, so it only detected a partial tag sitting at offset 0 of the buffer. With any earlier text in front, a split tag was flushed as content (opening case) or swallowed into `reasoning_buffer` where the next chunk could never complete it (closing case).

Replaced it with `partial_reasoning_tag_suffix_len()`, which returns the length of the longest buffer suffix that is a proper prefix of a reasoning tag; only that suffix stays buffered. While tracing, an unterminated `<think>` block turned out to drop the entire response, so the stream now flushes whatever is still held back at the end.

Before / after:

| chunks | before | after |
| --- | --- | --- |
| `"<think>reasoning</th"`, `"ink>answer"` | `content=""`, `reasoning=""` | `content="answer"`, `reasoning="reasoning"` |
| `"preface <th"`, `"ink>reasoning</think>answer"` | `content="preface <think>reasoning</think>answer"` | `content="preface answer"`, `reasoning="reasoning"` |
| `"<think>unterminated reasoning"` | `content=""`, `reasoning=""` | `reasoning="unterminated reasoning"` |

## PR Type

- 🐛 Bug Fix

## Relevant issues

Fixes #1242

## Checklist

- [x] New and existing tests pass locally (`tests/unit`: 1896 passed, 93 skipped)
- [x] Documentation was updated where necessary (n/a)
- [x] **AI Usage:**

## AI Usage Information

- Any other info you'd like to share: root cause and fix reviewed by @njbrake; the reporter's diagnosis in #1242 pointed at the same suffix matching approach.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved streaming responses when reasoning tags are split across multiple chunks.
  * Prevented incomplete reasoning tags from appearing prematurely in displayed content.
  * Preserved and correctly released buffered content and reasoning when streams end.
  * Improved handling of empty streams, unfinished reasoning sections, and non-content chunks.
  * Prevented terminal and response metadata from being duplicated during streaming.

* **Tests**
  * Added coverage for split and partial tags, unterminated reasoning blocks, configurable terminal chunks, optional metadata, and empty responses.

### DIFF
--- a/src/any_llm/providers/openai/xml_reasoning.py
+++ b/src/any_llm/providers/openai/xml_reasoning.py
@@ -34,6 +34,32 @@ def set_chunk_reasoning(chunk: ChatCompletionChunk, reasoning: str) -> ChatCompl
     return chunk
 
 
+def is_terminal_chunk(chunk: ChatCompletionChunk) -> bool:
+    """Return whether a chunk carries a terminal finish reason."""
+    return bool(chunk.choices and chunk.choices[0].finish_reason)
+
+
+def clear_chunk_stream_metadata(chunk: ChatCompletionChunk) -> ChatCompletionChunk:
+    """Remove metadata that must not be repeated on a synthetic flush chunk."""
+    if not chunk.choices:
+        return chunk
+
+    choice = chunk.choices[0]
+    choice.finish_reason = None
+    choice.logprobs = None
+    choice.delta.function_call = None
+    choice.delta.refusal = None
+    choice.delta.reasoning = None
+    choice.delta.role = None
+    choice.delta.tool_calls = None
+    choice.delta.extra_content = None
+    chunk.moderation = None
+    chunk.service_tier = None
+    chunk.system_fingerprint = None
+    chunk.usage = None
+    return chunk
+
+
 def wrap_chunks_with_xml_reasoning(
     chunks: AsyncIterator[ChatCompletionChunk],
 ) -> AsyncIterator[ChatCompletionChunk]:
@@ -53,6 +79,8 @@ def wrap_chunks_with_xml_reasoning(
         get_content=get_chunk_content,
         set_content=set_chunk_content,
         set_reasoning=set_chunk_reasoning,
+        is_terminal=is_terminal_chunk,
+        clear_stream_metadata=clear_chunk_stream_metadata,
     )
 
 
--- a/src/any_llm/utils/reasoning.py
+++ b/src/any_llm/utils/reasoning.py
@@ -1,6 +1,6 @@
 import re
 from collections.abc import AsyncIterator
-from typing import Any, TypeVar
+from typing import Any, Literal, TypeVar
 
 from any_llm.constants import REASONING_FIELD_NAMES
 
@@ -25,7 +25,7 @@ def find_reasoning_tag(text: str, opening: bool = True) -> tuple[int, str] | Non
     return (earliest_pos, earliest_tag) if earliest_tag else None
 
 
-def partial_reasoning_tag_suffix_len(text: str, opening: bool = True) -> int:
+def partial_reasoning_tag_suffix_len(text: str, *, tag_kind: Literal["opening", "closing"]) -> int:
     """Length of the longest suffix of text that is a proper prefix of a reasoning tag.
 
     A non-zero result means the tail of ``text`` may become a complete reasoning tag once
@@ -34,7 +34,7 @@ def partial_reasoning_tag_suffix_len(text: str, opening: bool = True) -> int:
     """
     longest = 0
     for tag_name in REASONING_FIELD_NAMES:
-        tag = f"<{tag_name}>" if opening else f"</{tag_name}>"
+        tag = f"<{tag_name}>" if tag_kind == "opening" else f"</{tag_name}>"
         for i in range(min(len(text), len(tag) - 1), 0, -1):
             if text.endswith(tag[:i]):
                 longest = max(longest, i)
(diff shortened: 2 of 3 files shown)


Vansh-Sharma27:
I tested the current head at `6daeb33de6efb97a2250cacd7dc01dd1ffcd8326` in clean base and PR worktrees with the actual project dependencies and chunk models.

The core suffix fix holds up. Both original #1242 reproductions fail on the base and pass on this PR. I ran 220 generated boundary cases covering every opening and closing tag split across all four `REASONING_FIELD_NAMES` values. All 220 pass on the PR, while 158 fail on the base. The targeted XML reasoning and affected provider tests pass, and the commit hooks are clean as well.

I reproduced the terminal ordering case that CodeRabbit mentioned. With these input chunks:

```text
0: content="prefix <th", finish_reason=None
1: content=None, finish_reason="stop"
```

I get this output order:

```text
0: content="prefix ", finish_reason=None
1: content=None, finish_reason="stop"
2: content="<th", finish_reason=None
```

Part of the payload is emitted after the terminal event. A consumer that stops on the first `finish_reason` misses the final `"<th"`.

There is also a related case when `finish_reason` is already attached to the last chunk carrying content. With this single input:

```text
content="<think>reasoning", finish_reason="length"
```

The output is:

```text
0: reasoning=None, finish_reason="length"
1: reasoning="reasoning", finish_reason="length"
```

Here the reasoning arrives after the first terminal event and `finish_reason` is emitted twice. The existing Hugging Face conversion path can represent a chunk carrying both content and `finish_reason`, although I did not have provider credentials to verify how often that shape occurs in a live remote stream.

I tested the EOF clone with project chunks that have a valid schema and carry sideband fields. `role`, `tool_calls`, `extra_content`, and `finish_reason` were each present once on input and twice on output because the already emitted `last_chunk` was copied again for the final flush.

For the EOF case with a partially closed tag:

```text
content="<think>reasoning</th"
```

The current output is:

```text
reasoning="reasoning</th"
```

I do not think that result is automatically wrong, since the EOF path is meant to preserve whatever remains buffered. However, it is currently undocumented and not covered by a committed test, so adding the test would make the chosen EOF policy explicit.

Would it be better to keep the verified suffix fix in this PR and handle EOF recovery separately, or adjust finalisation so buffered payload is emitted before or together with exactly one terminal event and sideband fields are forwarded only once? The original midstream fix does not depend on the EOF extension and passes every boundary case I tried.

PR #1254, opened 2026-08-10
fix(reasoning): preserve metadata chunk order

### LINKED ISSUE #1252: [BUG] XML reasoning parser delays metadata chunks until the end of the stream
**`any-llm` version:** `main` at `f2b11abfb5befb7d89b723eb170e33e60d275ec9` (`0.0.0.dev0`)

**Python Version:** `3.13.14`

### Describe the bug

I checked the streaming XML reasoning parser again after #1245 was merged. The original split tag problem from #1242 is fixed, but I found an ordering problem in the new `held_chunks` handling.

The problem happens when the parser is holding a partial XML tag and receives a chunk whose `delta.content` is `None`. That chunk can still contain useful stream data such as `role`, `tool_calls`, `extra_content`, or usage.

At the moment, the parser adds such chunks to `held_chunks`. It does not release them when the partial tag is completed. They are emitted only after the complete input stream has ended.

Because of this, a metadata chunk that arrived before later content can be returned after that content. It can even be returned after a later chunk containing `finish_reason="stop"`.

For example, this input order:

```text
c0: content="<think"
c1: content=None, extra_content={"marker": "mid stream"}
c2: content=">reasoning</think>answer"
c3: content=None, finish_reason="stop"
```

currently produces this output order:

```text
c2: content="answer", reasoning="reasoning"
c3: finish_reason="stop"
c1: extra_content={"marker": "mid stream"}
```

I expected the metadata chunk to remain before the later content and terminal chunk:

```text
c1
c2
c3
```

This is a stream ordering problem rather than a request for new behaviour. A caller may stop reading after `finish_reason`, so metadata moved behind that chunk may never be processed. Reordering a tool call delta can also change the order in which a caller receives tool call data.

### Example w/ Code

This test uses the real chunk models and the public XML reasoning wrapper:

```python
from collections.abc import AsyncIterator
from typing import Any, Literal

import pytest

from any_llm.providers.openai.xml_reasoning import wrap_chunks_with_xml_reasoning
from any_llm.types.completion import ChatCompletionChunk, ChoiceDelta, ChunkChoice


def make_chunk(
    chunk_id: str,
    content: str | None = None,
    *,
    extra_content: dict[str, Any] | None = None,
    finish_reason: Literal["stop"] | None = None,
) -> ChatCompletionChunk:
    return ChatCompletionChunk(
        id=chunk_id,
        choices=[
            ChunkChoice(
                index=0,
                finish_reason=finish_reason,
                delta=ChoiceDelta(
                    role=None,
                    content=content,
                    extra_content=extra_content,
                ),
            )
        ],
        created=1234567890,
        model="test model",
        object="chat.completion.chunk",
    )


async def iter_chunks(
    chunks: list[ChatCompletionChunk],
) -> AsyncIterator[ChatCompletionChunk]:
    for chunk in chunks:
        yield chunk


@pytest.mark.asyncio
async def test_partial_tag_keeps_metadata_chunk_order() -> None:
    chunks = [
        make_chunk("c0", "<think"),
        make_chunk("c1", extra_content={"marker": "mid stream"}),
        make_chunk("c2", ">reasoning</think>answer"),
        make_chunk("c3", finish_reason="stop"),
    ]

    result = [
        chunk
        async for chunk in wrap_chunks_with_xml_reasoning(iter_chunks(chunks))
    ]

    assert [chunk.id for chunk in result] == ["c1", "c2", "c3"]
```

On merged `main`, the assertion fails with:

```text
Expected: ["c1", "c2", "c3"]
Actual:   ["c2", "c3", "c1"]
```

The same behaviour was reproduced with a real `ChoiceDeltaToolCall`, a partial closing tag, a terminal content chunk, an empty choices usage chunk, and extra normal content after the tag was completed.

### What I found in the implementation

In `process_streaming_reasoning_chunks()`, a contentless chunk is appended to `held_chunks` whenever `buffer` or `reasoning_buffer` contains data:

```python
if not content:
    if buffer or reasoning_buffer:
        held_chunks.append(original_chunk)
```

Every held chunk is then emitted at the very end:

```python
for held_chunk in held_chunks:
    yield held_chunk
```

The pre merge parent `455ebe1f2f34358c186b94b0f58e433b5d764c84` yielded contentless chunks immediately. Running the same seven cases there produced no ordering violation. Running them on merged `main` reproduced the delayed ordering in every trace. The automated checker counted ten violations across six cases. The usage case also changed order, but it was not counted because that chunk had `choices=[]` instead of delta metadata.

### Validation

I ran the tests using the repository dependencies and the real project models.

```text
Merged main XML reasoning tests: 33 passed
Affected provider tests: 27 passed
Merged main full unit suite: 1986 passed, 66 skipped, 8 failed
Pre merge parent full unit suite: 1966 passed, 66 skipped, 8 failed
Configured repository checks: passed
```

The same eight full suite failures occurred on both revisions. Seven came from the missing optional `botocore[crt]` dependency, and one came from Windows symlink permissions.

No live provider request was made. The reproduction is a local parser test using the installed project types. The affected streaming wrapper is used by SambaNova, Portkey, MiniMax, and Hugging Face.

### References

1. Original report: https://github.com/mozilla-ai/any-llm/issues/1242

2. Merged fix: https://github.com/mozilla-ai/any-llm/pull/1245

3. Buffering branch on the tested commit: https://github.com/mozilla-ai/any-llm/blob/f2b11abfb5befb7d89b723eb170e33e60d275ec9/src/any_llm/utils/reasoning.py#L80-L88

4. End of stream release of held chunks: https://github.com/mozilla-ai/any-llm/blob/f2b11abfb5befb7d89b723eb170e33e60d275ec9/src/any_llm/utils/reasoning.py#L151-L178

5. XML reasoning wrapper: https://github.com/mozilla-ai/any-llm/blob/f2b11abfb5befb7d89b723eb170e33e60d275ec9/src/any_llm/providers/openai/xml_reasoning.py#L63-L84

6. Final review that mentioned the same ordering case: https://github.com/mozilla-ai/any-llm/pull/1245#pullrequestreview-4885939437

mikemikimike:
## Description
The XML reasoning stream wrapper currently delays every contentless chunk while a partial reasoning tag is buffered. This can move metadata such as role, tool calls, or provider extensions after later content and terminal chunks.

This change keeps ordinary metadata chunks in their original stream order while continuing to defer contentless terminal chunks until buffered content is flushed.

## PR Type

- 🐛 Bug Fix

## Relevant issues

Fixes #1252

## Checklist

- [x] AI Usage:
    - [ ] No AI was used.

## AI Usage Information

- AI Model used: GPT-5
- Any other info you'd like to share: The change was validated with the repository unit suite and local pre-commit checks.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved streaming behaviour so empty terminal chunks are no longer unnecessarily delayed.
  * Preserved the correct ordering and metadata of metadata-only chunks when reasoning content spans multiple streamed chunks.
  * Ensured final content and terminal chunks remain correctly processed during reasoning extraction.

* **Tests**
  * Added coverage for chunk ordering, metadata preservation, and reasoning/content extraction across streamed responses.

### DIFF
--- a/src/any_llm/utils/reasoning.py
+++ b/src/any_llm/utils/reasoning.py
@@ -81,7 +81,7 @@ async def process_streaming_reasoning_chunks(
         content = get_content(original_chunk)
 
         if not content:
-            if buffer or reasoning_buffer:
+            if is_terminal(original_chunk) and (buffer or reasoning_buffer):
                 held_chunks.append(original_chunk)
             else:
                 yield original_chunk
--- a/tests/unit/test_xml_reasoning.py
+++ b/tests/unit/test_xml_reasoning.py
@@ -24,10 +24,11 @@ def _make_chunk(
     finish_reason: Literal["stop", "length"] | None = None,
     role: Literal["assistant"] | None = "assistant",
     extra_content: dict[str, Any] | None = None,
+    chunk_id: str = "test-chunk",
 ) -> ChatCompletionChunk:
     """Create a minimal ChatCompletionChunk for testing."""
     return ChatCompletionChunk(
-        id="test-chunk",
+        id=chunk_id,
         choices=[
             ChunkChoice(
                 index=0,
@@ -338,6 +339,23 @@ async def test_wrap_chunks_preserves_terminal_metadata_after_partial_tag() -> No
 
     assert _accumulate(result) == ("prefix answer", "reasoning")
     assert [chunk.choices[0].finish_reason for chunk in result] == [None, "stop"]
+
+
+@pytest.mark.asyncio
+async def test_wrap_chunks_preserves_metadata_order_around_partial_tag() -> None:
+    """Metadata chunks stay in order while a partial tag is buffered."""
+    chunks = [
+        _make_chunk(content="<think", chunk_id="opening"),
+        _make_chunk(content=None, role=None, extra_content={"marker": "metadata"}, chunk_id="metadata"),
+        _make_chunk(content=">reasoning</think>answer", role=None, chunk_id="content"),
+        _make_chunk(content=None, finish_reason="stop", role=None, chunk_id="terminal"),
+    ]
+
+    result = await _collect_chunks(wrap_chunks_with_xml_reasoning(_async_iter_chunks(chunks)))
+
+    assert [chunk.id for chunk in result] == ["metadata", "content", "terminal"]
+    assert _accumulate(result) == ("answer", "reasoning")
+    assert result[0].choices[0].delta.extra_content == {"marker": "metadata"}
 
 
 @pytest.mark.asyncio


njbrake (approved):

Approving. Reproduced the #1252 failure on main (`['c2', 'c3', 'c1']`) and confirmed this branch gives `['c1', 'c2', 'c3']`, along with five related cases: metadata between reasoning body chunks, two consecutive metadata chunks, a partial closing tag, a usage chunk with empty choices, and a terminal chunk that carries content. Full unit suite and pre-commit clean locally, and CI is green across the matrix.

The useful property this gives us is that `held_chunks` can now only ever contain terminal chunks, so the terminal chunk is always last. That was the part of #1252 that mattered, since a caller that stops on `finish_reason` could previously miss metadata entirely.

Two residual inversions I checked and am accepting, both reachable only when the stream ends with a non-empty buffer (unterminated or truncated tag):

1. A usage chunk (`choices=[]`, so `is_terminal` is false) can now be emitted before the terminal chunk. Nothing is dropped, and consumers read usage off whichever chunk carries it.
2. A metadata chunk can precede the EOF flush of content that arrived before it. This looks inherent to the buffering design: the buffered tail cannot be emitted until the later chunk has already gone by, and splitting metadata off from content would break the single-flush-chunk behavior added in #1245.

Both are milder than what they replace. Thanks for the clear repro in the issue, it made this quick to verify.

PR #1256, opened 2026-08-11
fix(messages): keep sync streaming on a single event loop

### LINKED ISSUE #1253: [BUG] Sync messages streaming consumes the response on a different event loop
**`any-llm` version:** `main` at `f2b11abfb5befb7d89b723eb170e33e60d275ec9` (`0.0.0.dev0`)

**Python Version:** `3.13.14`

**OpenAI SDK Version:** `2.52.0`

**httpx Version:** `0.28.1`

### Describe the bug

I was comparing the sync streaming paths and found that `messages(stream=True)` creates and consumes its async response on two different event loops.

`AnyLLM.messages()` first calls `run_async_in_sync(self.amessages(...))`. For an OpenAI compatible provider, this opens the streaming HTTP response and returns an async iterator. After that call finishes, `messages()` passes the iterator to `async_iter_to_sync_iter()`, which consumes it on a new worker thread event loop.

The OpenAI SDK stream contains httpx and anyio objects that belong to the loop where the response was opened. Reading that stream from the second loop fails with:

```text
RuntimeError: <asyncio.locks.Event object at ...> is bound to a different event loop
```

The sync completion streaming path does not have this problem. It uses `async_coro_to_sync_iter()`, so creation and consumption happen on the same worker loop.

### Example w/ Code

The following script uses a local HTTP server and the real OpenAI SDK. It does not need an API key or a remote provider.

```python
import json
import threading
from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer
from typing import Any

from any_llm import AnyLLM


class StreamingHandler(BaseHTTPRequestHandler):
    protocol_version = "HTTP/1.1"

    def do_POST(self) -> None:
        content_length = int(self.headers.get("Content-Length", "0"))
        self.rfile.read(content_length)

        events: list[dict[str, Any]] = [
            {
                "id": "chunk 1",
                "object": "chat.completion.chunk",
                "created": 1234567890,
                "model": "test model",
                "choices": [
                    {
                        "index": 0,
                        "delta": {
                            "role": "assistant",
                            "content": "Hello",
                        },
                        "finish_reason": None,
                    }
                ],
            },
            {
                "id": "chunk 2",
                "object": "chat.completion.chunk",
                "created": 1234567890,
                "model": "test model",
                "choices": [
                    {
                        "index": 0,
                        "delta": {"content": " world"},
                        "finish_reason": None,
                    }
                ],
            },
            {
                "id": "chunk 3",
                "object": "chat.completion.chunk",
                "created": 1234567890,
                "model": "test model",
                "choices": [
                    {
                        "index": 0,
                        "delta": {},
                        "finish_reason": "stop",
                    }
                ],
            },
        ]

        body = "".join(
            f"data: {json.dumps(event)}\n\n" for event in events
        )
        body += "data: [DONE]\n\n"
        encoded = body.encode()

        self.send_response(200)
        self.send_header("Content-Type", "text/event-stream")
        self.send_header("Content-Length", str(len(encoded)))
        self.send_header("Connection", "close")
        self.end_headers()
        self.wfile.write(encoded)
        self.wfile.flush()

    def log_message(self, format: str, *args: Any) -> None:
        return


server = ThreadingHTTPServer(("127.0.0.1", 0), StreamingHandler)
server.daemon_threads = True
thread = threading.Thread(target=server.serve_forever, daemon=True)
thread.start()

api_base = f"http://127.0.0.1:{server.server_address[1]}/v1"

try:
    completion_client = AnyLLM.create_openai_compatible(
        name="local completion",
        api_base=api_base,
        api_key="test",
    )
    completion_result = list(
        completion_client.completion(
            model="test model",
            messages=[{"role": "user", "content": "Hello"}],
            stream=True,
        )
    )
    print("completion chunks:", len(completion_result))

    messages_client = AnyLLM.create_openai_compatible(
        name="local messages",
        api_base=api_base,
        api_key="test",
    )
    message_result = list(
        messages_client.messages(
            model="test model",
            messages=[{"role": "user", "content": "Hello"}],
            max_tokens=32,
            stream=True,
        )
    )
    print("message events:", len(message_result))
finally:
    server.shutdown()
    server.server_close()
    thread.join()
```

The completion control receives all three chunks. The Messages call fails while it starts reading the response:

```text
completion chunks: 3
RuntimeError: <asyncio.locks.Event object at ...> is bound to a different event loop
```

I repeated the Messages call three times and all three attempts failed in the same way. Calling it from inside an already running loop with `allow_running_loop=True` failed with `RuntimeError: Event loop is closed`.

### What I found in the implementation

The current Messages wrapper does this in two stages:

```python
response = run_async_in_sync(
    self.amessages(...),
    allow_running_loop=allow_running_loop,
)

return async_iter_to_sync_iter(response)
```

The completion and Responses streaming wrappers instead pass the coroutine directly to `async_coro_to_sync_iter()`. That helper obtains the iterator and consumes it inside one worker loop.

Using the same single loop approach for Messages streaming looks like one possible direction. I am mainly reporting the behaviour and the difference between the wrappers.

### Validation

The relevant unit files still pass because the current sync Messages test covers only a non streaming response.

```text
tests/unit/test_utils.py: 4 passed
tests/unit/test_messages.py: 49 passed, 1 skipped
```

No live provider was called. The local server returns valid SSE chunks and the request goes through the real OpenAI SDK, httpx, and anyio. This confirms the loop failure on the OpenAI compatible Messages bridge without depending on a remote service.

### Related issues checked

I checked the existing event loop issues before reporting this.

1. #1118 and #1119 are about AsyncClient cleanup messages during test teardown.

2. #477 discusses exceptions inside the sync and async wrapper in general.

3. #200 tracks the earlier async migration work.

I could not find an issue for the specific case where sync Messages streaming creates the response on one loop and consumes it on another.

### References

1. Current `messages()` implementation: https://github.com/mozilla-ai/any-llm/blob/f2b11abfb5befb7d89b723eb170e33e60d275ec9/src/any_llm/any_llm.py#L801-L827

2. Async iterator bridge: https://github.com/mozilla-ai/any-llm/blob/f2b11abfb5befb7d89b723eb170e33e60d275ec9/src/any_llm/utils/aio.py#L87-L188

3. Sync completion streaming control path: https://github.com/mozilla-ai/any-llm/blob/f2b11abfb5befb7d89b723eb170e33e60d275ec9/src/any_llm/any_llm.py#L576-L611

4. Sync Responses streaming control path: https://github.com/mozilla-ai/any-llm/blob/f2b11abfb5befb7d89b723eb170e33e60d275ec9/src/any_llm/any_llm.py#L1041-L1061

5. Async migration discussion: https://github.com/mozilla-ai/any-llm/issues/200

6. Sync wrapper exception discussion: https://github.com/mozilla-ai/any-llm/issues/477

7. Separate AsyncClient teardown issue: https://github.com/mozilla-ai/any-llm/issues/1118

8. Related teardown warning PR: https://github.com/mozilla-ai/any-llm/pull/1119

champ18ion:
## Description
`AnyLLM.messages(stream=True)` opened the async response via `run_async_in_sync(self.amessages(...))` on one worker event loop, then handed the resulting async iterator to `async_iter_to_sync_iter()`, which consumed it on a *different* event loop. For OpenAI-compatible providers, the SDK's httpx/anyio objects are bound to the loop that opened the stream, so consuming from a second loop raised:

```
RuntimeError: <asyncio.locks.Event object at ...> is bound to a different event loop
```

`completion()` and `Responses()` streaming don't hit this because they pass the coroutine straight into `async_coro_to_sync_iter()`, so the response is opened and consumed on the same worker loop.

- Route the `messages()` sync streaming path through `async_coro_to_sync_iter()`, mirroring the existing `completion()`/`Responses()` pattern.
- Pass `allow_running_loop` through to the non-streaming fallback's `async_iter_to_sync_iter()` call for consistency with `responses()`.
- Add a regression test in `tests/unit/test_messages.py` based on the repro from the issue: a local `ThreadingHTTPServer` emitting SSE chunks, driven through the real OpenAI SDK/httpx/anyio, no live provider or API key required. Confirmed it hangs/fails against the pre-fix code and passes with the fix.

## PR Type

- 🐛 Bug Fix

## Relevant issues
Fixes #1253

## Checklist
- [x] **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

- Any other info you'd like to share: I picked up this issue (assigned by @nathan-brake), diagnosed and posted the fix approach in the issue thread myself, then had Claude Code implement the one-line-swap plan I'd already confirmed, write the regression test, and run the local test/lint/type-check suite. I reviewed the diff and test output before opening this PR. Happy to discuss the approach directly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved synchronous streaming responses with OpenAI-compatible providers.
  * Streaming now works more reliably when an event loop is already active.
  * Preserved message-specific options for streaming and non-streaming responses.
  * Streaming request errors are reported when iteration begins.
  * Improved handling of asynchronous response streams when no explicit streaming option is provided.

### DIFF
--- a/src/any_llm/any_llm.py
+++ b/src/any_llm/any_llm.py
@@ -809,6 +809,14 @@ class AnyLLM(ABC):
     ) -> MessageResponse | ParsedMessage[Any] | ParsedBetaMessage[Any] | Iterator[MessageStreamEvent]:
         """Create a message using the Anthropic Messages API synchronously.
 
+        With `stream=True` the request is opened lazily on the first iteration, so errors
+        raised while opening it (a rejected `output_format` combination, an unsupported
+        `context_management`/`betas` request, or a provider auth or connection failure)
+        surface from the first `next()` rather than from this call. This matches
+        [AnyLLM.completion][any_llm.any_llm.AnyLLM.completion] and
+        [AnyLLM.responses][any_llm.any_llm.AnyLLM.responses]; wrap the iteration rather
+        than the call to catch them.
+
         See [AnyLLM.amessages][any_llm.any_llm.AnyLLM.amessages]
         """
         if allow_running_loop is None:
--- a/src/any_llm/any_llm.py
+++ b/src/any_llm/any_llm.py
@@ -813,6 +813,20 @@ class AnyLLM(ABC):
         """
         if allow_running_loop is None:
             allow_running_loop = INSIDE_NOTEBOOK
+        if kwargs.get("stream"):
+            return async_coro_to_sync_iter(
+                cast(
+                    "Coroutine[Any, Any, AsyncIterator[MessageStreamEvent]]",
+                    self.amessages(
+                        prompt_cache_key=prompt_cache_key,
+                        context_management=context_management,
+                        betas=betas,
+                        **kwargs,
+                    ),
+                ),
+                allow_running_loop=allow_running_loop,
+            )
+
         response = run_async_in_sync(
             self.amessages(
                 prompt_cache_key=prompt_cache_key,
@@ -824,7 +838,7 @@ class AnyLLM(ABC):
         )
         if isinstance(response, (MessageResponse, ParsedMessage, ParsedBetaMessage)):
             return response
-        return async_iter_to_sync_iter(response)
+        return async_iter_to_sync_iter(response, allow_running_loop=allow_running_loop)
 
     @handle_exceptions(wrap_streaming=True)
     async def amessages(
--- a/tests/unit/test_messages.py
+++ b/tests/unit/test_messages.py
@@ -54,6 +54,7 @@ from any_llm.types.messages import (
     ParsedTextBlock,
     StopReason,
 )
+from any_llm.utils.aio import async_iter_to_sync_iter
 
 
 def test_stop_reason_aliases_anthropic_beta_type() -> None:
@@ -1320,12 +1321,15 @@ def test_sync_messages_streaming_consumes_response_on_a_single_event_loop() -> N
     assert text == "Hello world"
 
 
-def test_sync_messages_bridges_an_iterator_returned_without_stream() -> None:
+def test_sync_messages_bridges_an_iterator_and_forwards_allow_running_loop() -> None:
     """A provider that hands back an iterator without `stream` still bridges to a sync iterator.
 
-    The streaming branch keys off `kwargs["stream"]`, so this covers the fallback
-    that converts a non-streaming call's async iterator, mirroring the same
-    fallback in `completion()` and `responses()`.
+    The streaming branch keys off `kwargs["stream"]`, so this covers the fallback that
+    converts a non-streaming call's async iterator, mirroring the same fallback in
+    `completion()` and `responses()`. `allow_running_loop` is asserted on the bridge call
+    rather than through behavior: the fallback is only reachable with no running loop, so
+    the flag changes nothing observable there, and dropping it would silently leave the
+    bridge on its own default of True.
     """
 
     async def event_stream() -> AsyncIterator[MessageStreamEvent]:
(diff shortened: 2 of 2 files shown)


njbrake (approved):

Approving. Good diagnosis, and the regression test is the right shape: it drives the real OpenAI SDK, httpx, and anyio over a local socket, so it reproduces the cross-loop failure instead of mocking around it. I confirmed it fails against pre-fix `main` and passes with the fix.

I pushed four commits to your branch:

1. Lint fixes CI had not run yet: `@override` on `log_message` (mypy `explicit-override`), dropped the local `AnyLLM` import since it is already imported at module scope, cast the stream so `event.type` type-checks, and asserted the streamed text and trailing `message_stop`.
2. A test for the non-streaming iterator fallback. Forwarding `allow_running_loop` left that line uncovered once the streaming branch took every real streaming call, and `codecov/patch` was failing at 66%.
3. A docstring note that streaming errors now surface from the first `next()` rather than from the call. That is a real contract change on a public method, and it matches `completion()` and `responses()`.
4. An assertion that `allow_running_loop` actually reaches the bridge, per your follow-up finding. The commit 2 test would have passed with the keyword deleted.

On the suggestion to bound the stream collection with a timeout helper: skipped. `pyproject.toml` already sets a 120s global pytest-timeout, and the pre-fix failure surfaces in under a second rather than hanging.

Worth knowing for future work here: nothing in `tests/integration` covers the sync `messages()`, `completion()`, or `responses()` streaming paths. They all use the async API, which is why this bug shipped without CI noticing. Your unit test is currently the only coverage of sync `messages(stream=True)`.

The integration run on this branch is red for unrelated reasons: a Google billing block on the gemini project (403 PERMISSION_DENIED) and an otari wallet at -$0.13. Both reproduce identically on `main`, same 23 failures.

PR #1263, opened 2026-08-11
feat(timeout): centralize per-request timeout handling across providers

### LINKED ISSUE #1262: Centralize per-request timeout handling across providers (opt-in support, like prompt_cache_key)
### Summary

The first-class `timeout` parameter added in #1255 is forwarded to every provider through `**kwargs`. Providers whose SDK accepts a `timeout` keyword (OpenAI-based, groq, cerebras, together, anthropic) handle it for free, but providers with a different SDK shape do not, and each needed an individual fix:

- **TypeError** before any request is sent: `ollama`, and (before it stripped the kwarg) `xai`, `mistral`, `cohere`.
- **Silent mis-send** into the request body / generation params: `sagemaker` (`json.dumps`'d into `Body`), `watsonx` (`params=`).
- **Folded into a provider config** that has no timeout concept: `lmstudio`.

In #1255 these were handled provider-by-provider (map where the SDK supports it — mistral `timeout_ms`, cohere `request_options`; warn+drop where it doesn't — xai, ollama, watsonx, sagemaker, lmstudio). `huggingface` already stripped it.

### The problem with the current approach

Per-provider handling is whack-a-mole: it is not enforced anywhere, so a **new provider added tomorrow** that forwards kwargs to its SDK will silently `TypeError` or mis-send `timeout` again, with no test or type to catch it. The two silent mis-sends (`sagemaker`, `watsonx`) are the dangerous class — no error, corrupted request.

### Proposed design

Centralize the contract the way `prompt_cache_key` already does with `PROMPT_CACHE_KEY_SUPPORT` + `_validate_prompt_cache_key`. Sketch:

- A per-provider capability declaration for request-timeout support (e.g. `native` / `mapped` / `unsupported`).
- The base layer stops blindly injecting `timeout` into `**kwargs` for every provider. It routes `timeout` only to providers that opt in; unsupported providers get a clean `UnsupportedParameterError` (or a single centralized warn+drop) instead of each leaking a vendor `TypeError` or silently mis-sending.
- New providers must declare their support explicitly, so the failure mode becomes a loud, testable default rather than a silent per-SDK surprise.

This also makes the `timeout=None` semantics (currently "unset", documented in #1255) a single place to reason about rather than N.

### Context

- Follow-up to #1255 (review by @njbrake surfaced the cross-provider scope).
- Related prior per-provider timeout work: #901 (Gemini), #1240 (Bedrock).

peteski22:
## Description

Follow-up to #1255. That PR added a first-class `timeout` parameter and forwarded it to every provider through `**kwargs`, handling each SDK shape individually (map where supported, warn+drop where not). As #1262 notes, that per-provider approach is whack-a-mole: a new provider that forwards kwargs to its SDK would silently `TypeError` or mis-send `timeout` again, with nothing to catch it. The two silent mis-sends (`sagemaker`, `watsonx`) are the dangerous class.

This PR centralizes the contract the way `prompt_cache_key` already does with `PROMPT_CACHE_KEY_SUPPORT`:

- A per-provider `TIMEOUT_SUPPORT` capability: `Literal["unsupported", "native", "mapped"]`, defaulting to `unsupported`.
- A single guard in the base `acompletion` (`_validate_and_forward_timeout`) that routes `timeout`:
  - `native`/`mapped` → forwarded via `kwargs` (mapped providers translate it in their own conversion);
  - `unsupported` → raises a clear `UnsupportedParameterError` (hinting to set it on the client via `client_args`) instead of leaking a vendor `TypeError` or silently mis-sending the value.
- `timeout=None` is treated as unset in one place, so an unbounded timeout is not expressible.

A new provider now inherits the safe `unsupported` default and must opt in explicitly, so the failure mode is a loud, testable default rather than a silent per-SDK surprise.

### Provider classification
- **native** (SDK accepts `timeout` directly): OpenAI base (covers all OpenAI-compatible and config-registry providers), Anthropic base, `cerebras`, `groq`, `together`.
- **mapped** (translated): Google base / `gemini`+`vertexai` (`http_options` ms), `bedrock` (per-timeout boto3 client), `cohere` (`request_options`), `mistral` (`timeout_ms`).
- **unsupported** (no per-request timeout): `huggingface`, `lmstudio`, `ollama`, `sagemaker`, `watsonx`, `xai`, `azure`. The now-dead per-provider warn+drop code is removed; each records *why* it is unsupported in a comment (e.g. `watsonx`/`sagemaker` would otherwise mis-send `timeout` into the request body).

### Behavior change
For `unsupported` providers, passing a per-request `timeout` now **raises `UnsupportedParameterError`** rather than logging a warning and dropping it (the #1255 behavior). This matches the `prompt_cache_key` precedent and avoids silently proceeding without a requested bound. Client-level timeouts via `client_args` are unaffected. Since the typed `timeout` shipped only in #1255, regression exposure is minimal.

### Note on `azure`
`azure` (Azure AI Inference SDK) is deliberately left at the conservative `unsupported` default: it was untouched by #1255, is not verified to honor a per-request timeout, and its SDK rejects unknown transport kwargs. It is flagged in-code for a maintainer to promote to `native`/`mapped` with a verifying test if appropriate.

### Notes on tests
- Added `test_acompletion_rejects_timeout_for_unsupported_provider` (central rejection path). The existing forward and `None`-drop tests plus the mapped-translation tests (`cohere`/`mistral`/`gemini`) continue to cover the supported paths; the obsolete per-provider warn+drop tests were removed.
- `pre-commit` (ruff + ruff-format + mypy strict) is clean for the changed code (no new mypy errors versus the `main` baseline).
- The only failing tests in a local `tests/unit` run are pre-existing and identical to `main`: the `mistralai` `ThinkChunk.signature` SDK drift and the `lmstudio` optional SDK not being installed. Neither is touched by this change.
- Integration tests were not run locally (they need real keys or the `run-integration-tests` label). The change is provider-SDK-agnostic at the boundary; a real `cohere`/`mistral`/`gemini`/`bedrock` run would confirm the mapped translation end to end.

## PR Type

- 🆕 New Feature
- 💅 Refactor

## Relevant issues

Fixes #1262
Follow-up to #1255 (per-provider timeout handling)

## Checklist

- [x] **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

- Any other info you'd like to share: Claude was used collaboratively. It assisted with investigation, design, drafting, and iteration under human review and direction, including the decision to raise `UnsupportedParameterError` (over a centralized warn+drop) and restoring the per-provider rationale for the `unsupported` classifications.

## Summary by CodeRabbit

* **New Features**
  * Added per-request timeout support to synchronous and asynchronous completion and messages requests.
  * Timeout values are automatically forwarded or converted for compatible providers.
  * Omitting a timeout preserves provider defaults.
  * Client-level timeout configuration remains available where per-request timeouts are unsupported.

* **Bug Fixes**
  * Unsupported per-request timeouts are now rejected clearly before requests are sent.
  * Improved timeout forwarding and validation across synchronous and asynchronous APIs.


njbrake (dismissed):

Approving. I verified every classification against the installed SDKs rather than the descriptions: `openai`, `anthropic`, `cerebras`, `groq` and `together` all take an explicit `timeout` on their create call; `huggingface` and `ollama` accept neither `timeout` nor `**kwargs`, so a forwarded value would have been a `TypeError`; `cohere` exposes only `request_options`. The single guard genuinely covers everything, since no provider overrides `acompletion` and the sync path delegates to it on both branches.

The `azure` classification was the one that could have removed working behavior, since it was the only provider silently passing `timeout` through on main. It didn't: I probed it against a local endpoint that sleeps 5 seconds, and `timeout=1` returned after 5.01s, identical to no timeout at all. azure-core's retry policy consumes `timeout` as an absolute retry budget and only clamps `connection_timeout`, so it never bounded a slow response. Raising is strictly better than that silent no-op.

I pushed a follow-up commit to the branch: assertions on each provider's `TIMEOUT_SUPPORT` (the classifications had no test coverage, since the only new test mutates the attribute on an `openai` instance), the docstrings in `api.py` and `acompletion` which still said timeout is "passed through to the provider's client/SDK", and two comment corrections. The bedrock comment pointed at `_convert_completion_params` but the pop is in `_completion`, and the azure comment said the SDK rejects unknown transport kwargs, which the probe above disproves.

One thing left for a follow-up, not blocking here since it predates this PR: `TIMEOUT_SUPPORT` is a class attribute, but bedrock's behaviour is per instance. `_client_for_timeout` returns the unmodified client and only warns when `_boto_session is None`, which is what a user-supplied `client=` sets. So `BedrockProvider(client=...)` with a timeout still proceeds unbounded, which is the silent drop this PR otherwise eliminates. Worth an issue.

Also worth adding `TIMEOUT_SUPPORT` to the provider-authoring section of CONTRIBUTING.md, since "a new provider must opt in explicitly" is the headline benefit and that section currently only mentions the `SUPPORTS_*` flags.

PR #1265, opened 2026-08-12
fix(gemini): route dict response schemas through response_json_schema

### LINKED ISSUE #1143: Gemini dict response_format fails for OpenAI strict schemas with additionalProperties
### Describe the bug

`any-llm` accepts OpenAI-style dict `response_format` for Gemini after #969 / #541, but it forwards the nested `json_schema.schema` directly to `google.genai` as `generation_config.response_schema`.

OpenAI strict JSON schemas commonly include `additionalProperties: false`. When that schema is passed to Gemini through `response_schema`, `google.genai` serializes it as `additional_properties`, and Gemini rejects the request:

```text
ClientError: 400 INVALID_ARGUMENT. {'error': {'code': 400, 'message': 'Invalid JSON payload received. Unknown name "additional_properties" at \'generation_config.response_schema\': Cannot find field.', 'status': 'INVALID_ARGUMENT', 'details': [{'@type': 'type.googleapis.com/google.rpc.BadRequest', 'fieldViolations': [{'field': 'generation_config.response_schema', 'description': 'Invalid JSON payload received. Unknown name "additional_properties" at \'generation_config.response_schema\': Cannot find field.'}]}]}}
```

This means the dict response_format support added for #541 works only for very simple schemas. It fails for a normal OpenAI strict structured-output schema.

Related prior work:

- #541 was closed as implemented by #969.
- #1003 was a duplicate follow-up for the same Gemini dict `response_format` support.
- #1094 / #1095 fixed a similar Gemini JSON Schema conversion issue for tool definitions by routing `$defs` / `$ref` schemas through `parameters_json_schema` instead of `types.Schema(...)`.

### Example w/ Code

```python
import asyncio
from any_llm import AnyLLM

schema = {
    "type": "object",
    "properties": {
        "name": {"type": "string"},
    },
    "required": ["name"],
    "additionalProperties": False,
}

response_format = {
    "type": "json_schema",
    "json_schema": {
        "name": "NameResponse",
        "strict": True,
        "schema": schema,
    },
}

async def main():
    llm = AnyLLM.create("gemini")
    await llm.acompletion(
        model="gemini-3-flash-preview",
        messages=[{"role": "user", "content": "Return JSON with name = ok"}],
        response_format=response_format,
        stream=False,
    )

asyncio.run(main())
```

### Expected behavior

The Gemini provider should accept OpenAI-compatible `response_format={"type":"json_schema", ...}` values that include `additionalProperties: false`, or sanitize/convert the OpenAI JSON Schema into the subset accepted by Gemini's `response_schema`.

### Actual behavior

The request fails with Gemini `400 INVALID_ARGUMENT` because `additionalProperties` is forwarded into `generation_config.response_schema` and becomes unsupported `additional_properties`.

### Environment

I reproduced this directly with `AnyLLM.create("gemini")`, not only through a gateway.

- `google-genai`: 1.70.0
- `pydantic`: 2.12.5
- Upstream main still forwards the schema directly in `src/any_llm/providers/gemini/base.py`:

```python
kwargs["response_schema"] = response_format["json_schema"]["schema"]
```

### Suggested fix direction

The Gemini provider likely needs a response-schema conversion/sanitization step instead of directly assigning the OpenAI schema dict to `response_schema`.

At minimum, tests should cover an OpenAI strict schema with `additionalProperties: false`, since that is the standard shape produced by OpenAI-style structured outputs.

mikemikimike:
## Description

Route OpenAI-style dictionary `response_format` schemas through Google GenAI's `response_json_schema` field. The field accepts the full JSON Schema dialect used by OpenAI-compatible callers, including `additionalProperties`, while preserving the existing `response_schema` path for Pydantic and dataclass response formats.

The regression coverage exercises both non-streaming and streaming requests with `additionalProperties: false` and verifies that the schema is not sent through the restricted `response_schema` field.

## PR Type

- [x] Bug Fix

## Relevant issues

Fixes #1143

## Validation

- `\.venv\Scripts\python.exe -m pytest tests/unit/providers/test_gemini_provider.py -q -p no:rerunfailures --tb=short`: 120 passed
- `ruff check --ignore CPY001 src/any_llm/providers/gemini/base.py tests/unit/providers/test_gemini_provider.py`: passed
- `ruff format --check src/any_llm/providers/gemini/base.py tests/unit/providers/test_gemini_provider.py`: passed
- `git diff --check`: passed
- Full `tests/unit` was collected successfully (2,116 tests) but did not finish within approximately six minutes locally; no failure was reported before it was stopped.
- Mypy reports five pre-existing errors in the changed module under the local dependency environment; the patch adds no new type-check diagnostics.

## Checklist

- [x] I have added unit tests that prove my fix/feature works.
- [ ] New and existing tests pass locally (the focused provider suite passes; the full unit suite timed out locally).
- [x] Documentation was updated where necessary (no documentation change is required).
- [ ] No AI was used.

## AI Usage Information

- AI Model used: GPT-5
- Any other info you'd like to share: The implementation and tests were reviewed against the existing provider call path and local test suite.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Gemini structured JSON responses by correctly applying the requested schema format.
  * Ensured generated and nested JSON objects reject unspecified properties for more predictable results.
  * Applied the fix consistently to standard and streaming completions.
  * Improved handling of dictionary-based JSON schemas with strict property definitions.

### DIFF
--- a/src/any_llm/providers/gemini/base.py
+++ b/src/any_llm/providers/gemini/base.py
@@ -157,7 +157,11 @@ class GoogleProvider(AnyLLM):
         response_format = params.response_format
         if is_structured_output_type(response_format):
             kwargs["response_mime_type"] = "application/json"
-            kwargs["response_schema"] = get_json_schema(response_format)
+            schema = get_json_schema(response_format)
+            if "additionalProperties" in schema:
+                kwargs["response_json_schema"] = schema
+            else:
+                kwargs["response_schema"] = schema
         elif isinstance(response_format, dict):
             response_type = response_format.get("type")
             if response_type == "json_schema":
--- a/src/any_llm/providers/gemini/base.py
+++ b/src/any_llm/providers/gemini/base.py
@@ -162,7 +162,7 @@ class GoogleProvider(AnyLLM):
             response_type = response_format.get("type")
             if response_type == "json_schema":
                 kwargs["response_mime_type"] = "application/json"
-                kwargs["response_schema"] = response_format["json_schema"]["schema"]
+                kwargs["response_json_schema"] = response_format["json_schema"]["schema"]
             elif response_type == "json_object":
                 kwargs["response_mime_type"] = "application/json"
             elif response_type == "text":
--- a/src/any_llm/providers/gemini/base.py
+++ b/src/any_llm/providers/gemini/base.py
@@ -37,6 +37,7 @@ try:
         _convert_tool_spec,
         _create_openai_chunk_from_google_chunk,
         _create_openai_embedding_response_from_google,
+        _has_additional_properties,
     )
 except ImportError as e:
     MISSING_PACKAGES_ERROR = e
@@ -158,7 +159,7 @@ class GoogleProvider(AnyLLM):
         if is_structured_output_type(response_format):
             kwargs["response_mime_type"] = "application/json"
             schema = get_json_schema(response_format)
-            if "additionalProperties" in schema:
+            if _has_additional_properties(schema):
                 kwargs["response_json_schema"] = schema
             else:
                 kwargs["response_schema"] = schema
--- a/src/any_llm/providers/gemini/utils.py
+++ b/src/any_llm/providers/gemini/utils.py
@@ -45,6 +45,22 @@ def _has_json_schema_refs(schema: Any) -> bool:
     return False
 
 
+def _has_additional_properties(schema: Any) -> bool:
+    """Return True if *schema* sets ``additionalProperties`` anywhere, nested included.
+
+    ``google.genai.types.Schema`` serializes the key as ``additional_properties``, which the
+    Gemini API rejects with a 400. Such schemas must be routed through
+    ``GenerateContentConfig.response_json_schema``, which the SDK forwards as raw JSON Schema.
+    """
+    if isinstance(schema, dict):
+        if "additionalProperties" in schema:
+            return True
+        return any(_has_additional_properties(v) for v in schema.values())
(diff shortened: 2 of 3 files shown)


ebarkhordar (commented):
The routing looks right, and `response_json_schema` is the field that takes the full dialect.

Possibly out of scope, but the same fault has a second entrance above your change. `is_structured_output_type` sends Pydantic and dataclass response formats to `get_json_schema` (`src/any_llm/utils/structured_output.py:32`), which is `model_json_schema()`, and a model declaring `ConfigDict(extra="forbid")` emits `additionalProperties: false` there.

Measured at 83b06e2 in a clean `python:3.12-slim` container (`pip install -e ".[gemini]"`, google-genai 2.17.0, pydantic 2.13.4), reading the `GenerateContentConfig` that `_convert_completion_params` returns:

| response_format | response_json_schema | response_schema |
| --- | --- | --- |
| the issue's dict | full schema, `additionalProperties: false` | None |
| `BaseModel` with `extra="forbid"` | None | `{"additionalProperties": false, ...}` |
| plain `@dataclass` | None | no `additionalProperties` |

Row two still hands `response_schema` the payload shape #1143 quotes in its 400. I have no Gemini key here, so that is the precondition and not a reproduced failure, and `extra="forbid"` is opt-in rather than the default, which is why I am asking rather than calling it a bug in this PR: is that branch a deliberate follow-up, or would keying the choice on whether the schema carries `additionalProperties`, instead of on the input type, close both entrances at once?

Issue #1260, opened 2026-08-11
Add integration coverage for sync streaming (completion/responses/messages)


champ18ion:
### Describe the bug/gap

The integration suite has no coverage of the *sync* streaming paths — `AnyLLM.completion(stream=True)`, `AnyLLM.responses(stream=True)`, and `AnyLLM.messages(stream=True)`. Every existing streaming integration test calls the async (`a`-prefixed) method instead, e.g. `test_custom_path_streaming` in [`tests/integration/test_openai_compatible.py`](https://github.com/mozilla-ai/any-llm/blob/main/tests/integration/test_openai_compatible.py#L43-L55) drives `llm.acompletion(..., stream=True)`.

This is why #1253 shipped without CI catching it: the bug was specific to the sync bridge (`run_async_in_sync` + `async_iter_to_sync_iter` opening and consuming the response on different event loops), and nothing in the integration suite ever exercised that bridge under a real streaming response. The unit regression test added in #1256 is currently the only coverage of sync `messages(stream=True)`, and there's equivalent async-only coverage for `completion()` and `responses()`.

### Proposed fix

Add sync twins of the existing async streaming integration tests:

- `tests/integration/test_completion.py` — a sync version of the streaming test, calling `.completion(stream=True)` instead of `.acompletion(stream=True)`.
- `tests/integration/test_responses.py` — same, for `.responses(stream=True)`.
- `tests/integration/test_messages.py` — same, for `.messages(stream=True)`.
- Optionally `tests/integration/test_openai_compatible.py::test_custom_path_streaming` — add a sync counterpart alongside the existing async one, since that's the custom/OpenAI-compatible path where #1253 actually manifested.

These would reuse the same live-endpoint fixtures, API-key skip guards, and assertions already in place — just swapping the call site from the async method to the sync one. No new test infrastructure needed.

### Context

Flagged by @njbrake during review of #1256 (see [review comment](https://github.com/mozilla-ai/any-llm/pull/1256#pullrequestreview)):

> Worth knowing for future work here: nothing in `tests/integration` covers the sync `messages()`, `completion()`, or `responses()` streaming paths. They all use the async API, which is why this bug shipped without CI noticing. Your unit test is currently the only coverage of sync `messages(stream=True)`.

champ18ion:
I'd like to pick this one up — could someone assign it to me (@champ18ion)? Following the same sync-twin pattern used to fix #1253.

njbrake:
[...]

PR #1267, opened 2026-08-12
test(integration): add sync streaming coverage for completion and messages

### LINKED ISSUE #1284: [BUG] Every sync xAI call fails: grpc channel is bound to the wrong event loop
**`any-llm` version:** `main` at `7148d4d`

**Python Version**: 3.13 and 3.14 (both affected)

### Describe the bug

Every synchronous xAI call fails before reaching the network. The async equivalents work. This affects `completion()`, `completion(stream=True)`, and `list_models()`.

The failure takes one of two forms depending on the thread's event loop state:

```text
RuntimeError: There is no current event loop in thread 'MainThread'.
RuntimeError: Task <...> got Future <...> attached to a different loop
```

Root cause: `XaiProvider._init_client` builds `XaiAsyncClient` eagerly, and the xAI SDK builds its `grpc.aio` channels in the constructor. grpc binds a channel to whichever event loop is current in the constructing thread. The sync API runs each request on a separate worker loop (`any_llm.utils.aio`), so the channel never belongs to the loop that drives the RPC.

Arranging for a loop to exist before construction does not help, since that loop is still not the one running the call. Upgrading xai-sdk does not help either: 1.18.0 still builds both channels in `AsyncClient._init`.

### Example w/ Code

```python
from any_llm import AnyLLM

llm = AnyLLM.create("xai")  # XAI_API_KEY set

for chunk in llm.completion(
    model="grok-3-mini",
    messages=[{"role": "user", "content": "hi"}],
    stream=True,
):
    print(chunk)
```

With a deliberately invalid key, the async path returns `INVALID_ARGUMENT: Incorrect API key provided` from x.ai, while every sync path above raises the loop error without ever reaching the API.

This went unnoticed because the integration suite only drove the async methods. #1267 adds the sync coverage that surfaced it, and carries the fix: build the client per running event loop, on first use, so the channel always belongs to the loop that will drive it.

champ18ion:
## Description
The integration suite only ever exercised the async (`a`-prefixed) streaming methods — `acompletion()`, `amessages()`, `aresponses()` — so nothing covered the sync bridge (`run_async_in_sync()` / `async_coro_to_sync_iter()`) against a real streaming HTTP response. That gap is exactly why #1253 shipped without CI catching it: the bug was specific to the sync bridge opening and consuming the response on two different event loops.

Adds sync twins of the existing async streaming integration tests, reusing the same fixtures, `MissingApiKeyError`/`UnsupportedParameterError` skip guards, and assertions, just swapping the call site from the `a`-method to the sync one:

- `test_streaming_completion_sync` in `tests/integration/test_streaming.py` — mirrors `test_streaming_completion_async`, drives `completion(stream=True)` across all providers.
- `test_messages_streaming_sync` in `tests/integration/test_messages.py` — mirrors `test_messages_streaming`, drives `messages(stream=True)` across all providers.
- `test_custom_path_streaming_sync` in `tests/integration/test_openai_compatible.py` — mirrors `test_custom_path_streaming`, drives `completion(stream=True)` through the OpenAI-compatible custom path against a live OpenAI endpoint. This is the exact path where #1253 manifested.

This new coverage surfaced a real, separate bug: every sync xAI call was broken (`XaiProvider` built its grpc channel eagerly and bound it to whichever loop existed at construction, so the sync bridge's worker loop never matched). @njbrake filed #1284 for it and fixed it directly on this branch: the client is now built per running event loop, on first use, so the channel always belongs to the loop driving the call. Thanks for catching and fixing that one.

**Scope note:** `responses()` streaming has no existing async integration test to mirror either (only unit coverage), so it's left out here rather than growing this into a bigger, separate task.

## PR Type

- 🐛 Bug Fix
- 🚦 Infrastructure

## Relevant issues
Refs #1260 (follow-up from #1256 / #1253)
Fixes #1284

## Checklist
- [x] **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

- Any other info you'd like to share: I filed #1260 myself off a gap @njbrake pointed out while reviewing #1256, then had Claude Code write the sync twins mirroring the existing async tests, run ruff/mypy, and confirm the new tests collect and skip cleanly without API keys. When CodeRabbit and then @njbrake flagged real bugs (a skip-guard ordering issue, then the xai event-loop bug this coverage surfaced), I reproduced each independently before accepting the fix, and reset to @njbrake's version once his fix on the xai issue turned out more robust than the one I'd been writing. I reviewed the diff before opening and updating this PR.

### DIFF
--- a/src/any_llm/providers/xai/xai.py
+++ b/src/any_llm/providers/xai/xai.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import asyncio
 import dataclasses
 import json
 from typing import TYPE_CHECKING, Any
@@ -118,6 +119,15 @@ class XaiProvider(AnyLLM):
 
     @override
     def _init_client(self, api_key: str | None = None, api_base: str | None = None, **kwargs: Any) -> None:
+        # grpc.aio's channel construction binds to "the current event loop" for this
+        # thread via asyncio.get_event_loop(). That call raises once a previous loop
+        # (e.g. from an earlier asyncio.run()) has been created and closed on this
+        # thread, which _init_client can hit when constructed synchronously and
+        # outside of any active event loop.
+        try:
+            asyncio.get_event_loop()
+        except RuntimeError:
+            asyncio.set_event_loop(asyncio.new_event_loop())
         self.client = XaiAsyncClient(api_key=api_key, **kwargs)
 
     @override
--- a/src/any_llm/providers/xai/xai.py
+++ b/src/any_llm/providers/xai/xai.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+import contextlib
 import dataclasses
 import json
 from typing import TYPE_CHECKING, Any
@@ -34,6 +35,20 @@ if TYPE_CHECKING:
     from any_llm.types.model import Model
 
 
+def _close_client(client: XaiAsyncClient) -> None:
+    """Release the grpc sockets held by a client whose event loop has already closed.
+
+    XaiAsyncClient.close() is a coroutine and its channels belong to a loop that is gone, so
+    there is nothing left to await it on; the cygrpc channel underneath is what actually
+    holds the socket and closes synchronously. Cleanup must never break a request, so if
+    either library moves these internals this degrades to the previous behavior, an idle
+    socket held until the process exits, rather than raising.
+    """
+    for channel in (client._api_channel, client._management_channel):
+        with contextlib.suppress(Exception):
+            channel._channel.close()  # type: ignore[union-attr]
+
+
 class XaiProvider(AnyLLM):
     API_BASE = "https://api.x.ai/v1"
     ENV_API_KEY_NAME = "XAI_API_KEY"
@@ -57,8 +72,6 @@ class XaiProvider(AnyLLM):
     TIMEOUT_SUPPORT = "unsupported"
 
     MISSING_PACKAGES_ERROR = MISSING_PACKAGES_ERROR
-
-    client: XaiAsyncClient
 
     @staticmethod
     @override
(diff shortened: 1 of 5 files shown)


njbrake:
Hi @champ18ion thank you for the PR! Could you also fix the bug so that the xai tests pass?

Both fail at AnyLLM.create(provider, ...) and never reach the streaming bridge:

src/any_llm/providers/xai/xai.py:121: in _init_client
    self.client = XaiAsyncClient(api_key=api_key, **kwargs)
grpc/aio/_channel.py:382: self._loop = cygrpc.get_working_loop()
E  RuntimeError: There is no current event loop in thread 'MainThread'.

XaiProvider._init_client builds the async client eagerly in __init__, and grpc demands a current event loop in the calling thread at channel construction. I isolated it against the locked grpcio==1.83.0:

champ18ion:
Thanks for catching that, @njbrake! Reproduced it independently outside the test suite:

```python
import asyncio
from any_llm import AnyLLM

async def main():
    await asyncio.sleep(0)

asyncio.run(main())
AnyLLM.create("xai", api_key="test-key")  # raises the same RuntimeError
```

Root cause: `XaiProvider._init_client` builds `XaiAsyncClient` eagerly, and `grpc.aio`'s channel construction binds to "the current event loop" for the calling thread via `asyncio.get_event_loop()`. That call raises once a prior loop on the thread has been created and closed (e.g. by an earlier `asyncio.run()`) — exactly the state my new sync integration tests hit when they call `AnyLLM.create("xai", ...)` synchronously after other async provider tests have already run in the same worker.

Fix: ensure a loop is set on the current thread before constructing the client, in `_init_client`. Added a regression unit test that reproduces the closed-loop state directly (deliberately not mocking `XaiAsyncClient`, since the bug is in the real grpc channel construction) and confirmed it fails without the fix and passes with it. Pushing now.

njbrake:
@champ18ion Might be that it's still in progress but I don't think you pushed anything after your most recent comment 🙏

champ18ion:
Sumimasen, I said I was pushing but actually forgot to run the commit. Catching it now, pushing shortly.

champ18ion:
Hey @njbrake, quick heads up, I rebased onto main to clear the conflict from #1280 and force pushed. Should've probably just merged instead, but it's a clean rebase (only a trivial append conflict in the xai test file, resolved by keeping both tests) and everything still passes. Sorry for the extra noise on the PR today between this and the earlier push mixup. Appreciate you catching the xai bug, that was a good find. Let me know if you want any changes, otherwise this should be ready whenever you get a chance to take a look.

champ18ion:
@njbrake, Saw you pushed a fix directly, thanks! I'd actually gone one round deeper too after CodeRabbit's comment (the construction fix alone still failed at the first RPC from a different loop, confirmed that locally), but yours is better: keyed by loop instead of just tracking the last one, so it holds up under concurrent loops, and the explicit socket cleanup for closed loops avoids the channel leak mine would've had. Reset my branch to yours instead of pushing a redundant version on top. Verified it locally too, all tests, ruff, and mypy pass. Thanks for jumping in on this one.

njbrake:
Heads up: I pushed the commit now sitting at `c1eb0c9` to your branch, replacing the xai fix, and opened #1284 to track the underlying bug. Your three integration tests are untouched and ready. Thanks for rebasing it onto main, that also cleared the `wait-for-http.sh` failure from #1283.

The loop binding approach could not work, and it is worth knowing why. With your fix applied, `AnyLLM.create("xai", ...)` stops raising, but the first RPC then fails with `RuntimeError: ... attached to a different loop`. Reproduced on 3.13 and 3.14.

The root cause sits a layer below the symptom you were handed. grpc binds a channel to whichever event loop is current in the constructing thread, and the sync bridge runs every request on a fresh worker loop. So a loop arranged at construction time is never the one that drives the call.

The replacement builds the client lazily, once per running event loop, so the channel always belongs to the loop that will use it. Clients whose loop has closed get dropped and their sockets released, since otherwise a sync caller accumulates one live grpc channel per request.

I also replaced the unit test. Asserting `provider.client is not None` passes with or without the bug, because it never makes an RPC.

Sorry for going straight at your branch. Happy to back it out if you would rather take it another way.

champ18ion:
No need to back it out, your version is better and it's already what's on the branch, happy with it. Thanks for the writeup on why the construction-only fix couldn't have worked, good to have that on record.

No need to touch #1284 separately either. Your commit here already fixes it, and the PR description now says "Fixes #1284" so it'll close automatically on merge.

PR #1278, opened 2026-08-12
fix(ollama): send tool history in Ollama's native format

jonigl:
## Description

`OllamaProvider._acompletion` rewrites tool history before sending it: tool results become `user` messages, and assistant tool calls are flattened into the assistant's text as raw OpenAI JSON. Ollama accepts both natively, so nothing needs flattening (https://docs.ollama.com/capabilities/tool-calling).

I hit this in [ollmcp](https://github.com/jonigl/mcp-client-for-ollama), where the model would sometimes print a tool call as raw JSON instead of calling the tool, but only once a tool had already run in the conversation. It was not just small local models — gemma4:31b-cloud through Ollama Cloud did it too. With this branch I can no longer reproduce it.

Four things go wrong on the released path:

- the tool result is attributed to the **user**
- `json.dumps` wraps it in quotes and escapes its newlines into literal `\n`, so a multi-line page snapshot arrives as one line of escape sequences
- the assistant's tool call is shown as **text**, which is a few-shot demonstration of the wrong output format. Smaller models copy it and emit the next tool call as prose instead of a structured call, so the caller gets a JSON blob as the answer.
- the trailing `\n[]`, because the branch tests `"tool_calls" in message` rather than its truth, so every assistant turn carrying an empty list gets a bare `[]` glued to its text

The fix keeps `role: "tool"` and sets `tool_name`, resolved from the matching assistant tool call (OpenAI addresses a tool result by `tool_call_id`, Ollama by name; templates such as gpt-oss render it). Assistant tool calls are forwarded in their own field, with `arguments` converted from OpenAI's JSON string to the mapping `ollama._types.Message.ToolCall` requires. The conversion runs before the streaming fork, so both paths are covered.

It also fixes a crash on the same path: `model_dump(exclude_none=True)` drops `content` when it is `None`, and the old code indexed `input_message["content"]` directly, raising `KeyError: 'content'`.

One judgment call worth your opinion: `json.loads` on `arguments` now raises on malformed JSON, where the old code passed it through silently. I went with failing loudly, since a tool call with unparseable arguments cannot be executed either way. Happy to downgrade it to a warning.

### Here a simple way to check what is going on

No Ollama server or model needed; the messages are captured before the request goes out.

```bash
# released
uv run --no-project --with 'any-llm-sdk[ollama]' python context_dump.py

# this PR
uv run --no-project --with ollama \
  --with "any-llm-sdk @ git+https://github.com/jonigl/any-llm@fix/ollama-native-tool-messages" \
  python context_dump.py
```

<details>
<summary><code>context_dump.py</code></summary>

```python
import asyncio

from any_llm import AnyLLM

HISTORY = [
    {"role": "user", "content": "Open the news page and summarize it"},
    {"role": "assistant", "content": "", "tool_calls": [
        {"id": "c1", "type": "function", "function": {
            "name": "browser_navigate", "arguments": '{"url": "https://news.google.com/"}'}}]},
    {"role": "tool", "tool_call_id": "c1",
     "content": "### Page state\n- generic [ref=e1]: Before you continue\n- button [ref=e2]: Accept all"},
    {"role": "assistant", "content": "I opened the page.", "tool_calls": []},
]


async def spy(*args, **kwargs):
    for message in kwargs.get("messages") or args[1]:
        print(message)
    raise SystemExit


llm = AnyLLM.create("ollama", api_base="http://localhost:11434")
llm.client.chat = spy
asyncio.run(llm.acompletion(model="x", messages=HISTORY, stream=False))
```

</details>

### Verification

image

## PR Type

- 🐛 Bug Fix

## Checklist

- [x] **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

- AI Model used: Opus 5
- AI Developer Tool used: Claude Code + vscode

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Ollama tool-call handling for streaming and non-streaming responses.
  * Preserved tool messages, names and assistant tool calls when forwarding conversation history.
  * Normalised missing or empty tool-call arguments and content for more reliable requests.

* **Tests**
  * Added coverage for tool calls, tool results, missing names, and empty content scenarios.

### DIFF
--- a/src/any_llm/providers/ollama/ollama.py
+++ b/src/any_llm/providers/ollama/ollama.py
@@ -267,9 +267,12 @@ class OllamaProvider(AnyLLM):
         cleaned_messages = []
         for input_message in params.messages:
             if input_message["role"] == "tool":
+                # OpenAI also allows a tool result as a list of content parts, while Ollama
+                # takes a plain string, so flatten it to its text the same way user turns are.
+                tool_content, _ = self._extract_images_from_message(input_message)
                 cleaned_message: dict[str, Any] = {
                     "role": "tool",
-                    "content": input_message.get("content") or "",
+                    "content": tool_content or "",
                 }
                 # OpenAI identifies a tool result by call id, Ollama by tool name.
                 tool_name = input_message.get("name") or tool_names_by_call_id.get(
--- a/src/any_llm/providers/ollama/ollama.py
+++ b/src/any_llm/providers/ollama/ollama.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import json
 from typing import TYPE_CHECKING, Any, Literal
 
 from typing_extensions import override
@@ -15,6 +14,7 @@ try:
 
     from .utils import (
         _convert_models_list,
+        _convert_tool_calls,
         _create_chat_completion_from_ollama_response,
         _create_openai_chunk_from_ollama_chunk,
         _create_openai_embedding_response_from_ollama,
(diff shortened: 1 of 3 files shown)


jonigl:
I will take a look to those 2 CodeRabbit comments tomorrow :)

njbrake (approved):
Approved. I reproduced all four defects against a live Ollama server and confirmed the replacement format against ollama-python at the pinned floor (0.6.0, where `Message.tool_name` and `ToolCall.Function.arguments: Mapping[str, Any]` both already exist) and against Ollama's tool-calling docs. The payload the provider now emits passes the client's own validation, and a real server returns a structured tool call from it. The `KeyError: 'content'` also reproduces on main and is fixed here.

On your open question about `json.loads` raising: keep it. anthropic, bedrock, and gemini all parse `arguments` unguarded (`anthropic/utils.py:174`, `bedrock/utils.py:385`, `gemini/utils.py:227`), so failing loudly matches what the rest of the repo already does. Same reasoning applies to CodeRabbit's "require a mapping" comment: `'[]'`, `'null'`, and `'1'` already fail at ollama's own validation with "Input should be a valid dictionary", so the check would move the error rather than create it. I'd skip it. Its second comment (a malformed JSON test case) is worth adding if you want it, but it isn't blocking.

I pushed one commit to your branch: OpenAI also allows a tool result's `content` as a list of content parts, which used to survive as a `json.dumps` string and now raises a ValidationError before the request goes out. It reuses `_extract_images_from_message`, same as the user turns.

Follow-up on our side, not yours: `tests/integration/test_agent_loop.py` skips all local providers with "does not support tools", so this path has no integration coverage. That skip is now wrong for ollama.

jonigl:
hey @njbrake, thanks for the review and the merge! 🙂 (appreciate you reproducing all four defects)

off topic: I like what you are building with [agent-of-empires](https://github.com/agent-of-empires/agent-of-empires) (and great name by the way!). I love tmux ❤️, so putting it within reach of more people is a great idea 🚀

njbrake:
@jonigl thank you! agent-of-empires and otari are the other passion projects I have alongside of any-llm, if you give them a try or have any thoughts feel free to hit me up on discord/x/gh/etc/etc 😄

PR #1279, opened 2026-08-13
fix(gemini): accumulate non-streaming reasoning parts

### LINKED ISSUE #1277: [BUG] Gemini non-streaming converter overwrites reasoning across thought parts instead of accumulating

**`any-llm` version:** 1.25.0
**Python Version**: 3.13

### Describe the bug

The non-streaming Gemini converter overwrites reasoning text instead of accumulating it, so a response carrying more than one thought part keeps only the last one. The streaming converter in the same file gets this right.

Non-streaming, [`_convert_response_to_response_dict`](https://github.com/mozilla-ai/any-llm/blob/main/src/any_llm/providers/gemini/utils.py#L350):

```python
for part in parts or []:
    if getattr(part, "thought", None):
        reasoning = part.text        # overwrite
```

Streaming, [`_create_openai_chunk_from_google_chunk`](https://github.com/mozilla-ai/any-llm/blob/main/src/any_llm/providers/gemini/utils.py#L452):

```python
for part in parts or []:
    if part.thought:
        reasoning_content += part.text or ""   # accumulate
```

Same request, streamed vs not, can therefore return different reasoning. The non-streaming result is silently truncated, with no error.

### Example w/ Code

Any non-streaming Gemini completion with `reasoning_effort` set where the candidate comes back with two or more `thought` parts. The `include_thoughts=True` path makes this reachable in normal use.

```python
result = await AnyLLM.create(LLMProvider.GEMINI).acompletion(
    model="gemini-3-flash-preview",
    messages=[{"role": "user", "content": "..."}],
    reasoning_effort="high",
)
result.choices[0].message.reasoning.content  # only the final thought part
```

### Suggested fix

Accumulate in the non-streaming path the way the streaming path does, and add a unit test with a multi-part thought response. `vertexai` shares this code. Found while reviewing #1275; unrelated to that change.

mikemikimike:
## Description
Fix Gemini non-streaming response conversion so reasoning text from multiple thought parts is accumulated instead of overwritten. This keeps non-streaming reasoning consistent with the existing streaming converter and preserves `None` when no thought parts are present.

## PR Type

- [x] Bug Fix

## Relevant issues

Fixes #1277

## Checklist

## Validation

- `uv run --frozen pytest -q tests/unit/providers/test_gemini_provider.py` - 130 passed
- `uv run --frozen pytest -q tests/unit` - 2108 passed, 69 skipped
- Targeted pre-commit hooks for changed files passed: ruff, ruff-format, mypy, codespell, line-ending checks

## AI Usage Information

- AI Model used: GPT-5
- Any other info: Implemented in an isolated worktree from `upstream/main`; no provider API key was used.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Gemini response handling so reasoning and answer text from multiple parts are accumulated correctly.
  * Preserved the separation between reasoning and regular answer content.
  * Normalised responses without reasoning so they are handled consistently.
  * Ignored unsupported response parts without affecting valid content.

* **Tests**
  * Added regression coverage for combined reasoning, text-only responses, and multi-part answer text.
  * Added coverage for responses containing parts without text or function calls.

### DIFF
--- a/src/any_llm/providers/gemini/utils.py
+++ b/src/any_llm/providers/gemini/utils.py
@@ -384,8 +384,8 @@ def _convert_response_to_response_dict(response: types.GenerateContentResponse) 
                     tool_call_dict["extra_content"] = extra_content
 
                 tool_calls_list.append(tool_call_dict)
-            elif getattr(part, "text", None):
-                text_content = part.text
+            elif part_text := getattr(part, "text", None):
+                text_content = (text_content or "") + part_text
 
         # Truncated or filtered responses produce a choice even without content or tool
         # calls, e.g. a thinking model that spent the whole max_output_tokens budget on
--- a/src/any_llm/providers/gemini/utils.py
+++ b/src/any_llm/providers/gemini/utils.py
@@ -363,7 +363,7 @@ def _convert_response_to_response_dict(response: types.GenerateContentResponse) 
 
         for part in parts or []:
             if getattr(part, "thought", None):
-                reasoning = part.text
+                reasoning = (reasoning or "") + (part.text or "")
             elif function_call := getattr(part, "function_call", None):
                 args_dict = {}
                 if args := getattr(function_call, "args", None):
--- a/src/any_llm/providers/gemini/utils.py
+++ b/src/any_llm/providers/gemini/utils.py
@@ -396,7 +396,7 @@ def _convert_response_to_response_dict(response: types.GenerateContentResponse) 
                     "message": {
                         "role": "assistant",
                         "content": None if tool_calls_list else text_content,
-                        "reasoning": reasoning,
+                        "reasoning": reasoning or None,
                         "tool_calls": tool_calls_list or None,
                     },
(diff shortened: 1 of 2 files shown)


njbrake (approved):
Approving. The fix follows the right precedent, `_create_openai_chunk_from_google_chunk` in the same file.

I pushed three commits while reviewing:

- `fdc5a60`: the original one-liner turned a textless thought part, one carrying only a `thought_signature`, into `Reasoning(content="")` instead of `None`, since `(None or "") + (None or "")` is `""`. Now `reasoning or None`, matching the streaming converter.
- `5abf339`: the same overwrite sat one branch below, on `text_content`. google-genai's own `GenerateContentResponse.text` concatenates every non-thought text part, so multi-part text is a real shape and the last one was winning.
- `d88556f`: covers the changed `elif`'s false arm (a part with neither text nor a function call), which Codecov had flagged as a partial branch.

Integration tests ran green against real keys, including `test_completion_reasoning[gemini]` and `test_completion_reasoning_streaming[gemini]`.

Thanks for the tidy fix and for including the regression test.

PR #1281, opened 2026-08-13
fix(gemini): use thinking level for Gemini 3.5+

### LINKED ISSUE #1276: [BUG] Gemini provider only sends thinking_budget, which Gemini 3.5+ will reject

**`any-llm` version:** 1.25.0
**Python Version**: 3.13

### Describe the bug

`GoogleProvider._convert_completion_params` only ever sends `thinking_budget` for `reasoning_effort` ([base.py#L134-L140](https://github.com/mozilla-ai/any-llm/blob/main/src/any_llm/providers/gemini/base.py#L134-L140)). Gemini 3.0 still accepts that for backward compatibility, so this works today. Google has signalled it goes away for Gemini 3.5 and newer, at which point every explicit `reasoning_effort` will fail against a current model.

The signal is in `google-genai` itself. `ReinforcementTuningHyperParameters.thinking_level` (types.py, SDK 2.16.0) reads:

> Starting from Gemini 3.5 models, the old thinking_budget will no longer be supported and will result in a user error if set. Instead, users should use the thinking_level parameter to control the maximum thinking depth.

That text is on a tuning config rather than the inference `ThinkingConfig`, so treat it as direction rather than proof for `generate_content`. The [Gemini 3 guide](https://ai.google.dev/gemini-api/docs/gemini-3) points the same way: `thinking_budget` is "still supported for backward compatibility, but we recommend migrating to `thinking_level`".

### Example w/ Code

```python
await AnyLLM.create(LLMProvider.GEMINI).acompletion(
    model="gemini-3.5-flash",
    messages=[{"role": "user", "content": "hi"}],
    reasoning_effort="low",   # becomes thinking_budget=1024
)
```

### Suggested fix

Branch on the model version in `_convert_completion_params`: send `thinking_level` for Gemini 3.5+, keep `thinking_budget` below that. `types.ThinkingLevel` accepts MINIMAL/LOW/MEDIUM/HIGH only, so `xhigh` and `max` collapse onto HIGH. Never send both, which is its own 400. This affects `vertexai` too, since both share `GoogleProvider`.

Not urgent: it only bites when the test suite or a user moves to a 3.5+ model. Related to #1275, which moves the reasoning tests to `gemini-3-flash-preview` and notes this in a comment.

mikemikimike:
## Description

Use Gemini thinking levels for Gemini 3.5+ models while preserving thinking-budget fallback for older models and SDKs without thinking-level support. The model-version parser also handles future minor and major versions.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure

## Relevant issues

Fixes #1276

## Checklist

- [x] AI Usage: AI was used for drafting/refactoring.

## AI Usage Information

- AI Model used: GPT-5
- Any other info you'd like to share: Used to inspect review feedback and draft focused regression coverage.

## Summary by CodeRabbit

* **New Features**
  * Added support for configurable reasoning levels on newer Gemini models, including high-effort reasoning.
  * Reasoning settings are now automatically selected based on the Gemini model version.

* **Bug Fixes**
  * Preserved existing thinking-budget behaviour for older Gemini models.
  * Improved compatibility across supported Gemini model generations.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,12 +39,12 @@ mistral = [
 anthropic = []
 
 gemini = [
-  "google-genai",
+  "google-genai>=1.51.0",
   "google-cloud-storage",
 ]
 
 vertexai = [
-  "google-genai",
+  "google-genai>=1.51.0",
   "google-cloud-storage",
 ]
 
--- a/src/any_llm/providers/gemini/base.py
+++ b/src/any_llm/providers/gemini/base.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import json
+import re
 from typing import TYPE_CHECKING, Any, ClassVar, cast
 
 from typing_extensions import override
@@ -77,6 +78,17 @@ REASONING_EFFORT_TO_THINKING_LEVELS = {
     "max": "HIGH",
 }
 _SUPPORTED_BATCH_ENDPOINTS = frozenset({"/v1/chat/completions"})
+_THINKING_LEVEL_MIN_GEMINI_VERSION = (3, 5)
+_GEMINI_VERSION_PATTERN = re.compile(r"gemini-(\d+)(?:\.(\d+))?")
+
+
+def _uses_thinking_level(model_id: str) -> bool:
+    """Gemini 3.5 and newer reject `thinking_budget` and expect `thinking_level` instead."""
+    match = _GEMINI_VERSION_PATTERN.search(model_id.lower())
+    if match is None:
+        return False
+    major, minor = int(match.group(1)), int(match.group(2) or 0)
+    return (major, minor) >= _THINKING_LEVEL_MIN_GEMINI_VERSION
 
 
 class GoogleProvider(AnyLLM):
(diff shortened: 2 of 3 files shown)


njbrake (dismissed):
Approving. The version branch only fires for model ids that parse as 3.5 or newer, so every model that exists today, including the `gemini-3-flash-preview` the integration suite pins, keeps the `thinking_budget` path. The regression surface on current users is nil.

I pushed a commit to your branch fixing the two new tests. `CompletionParams` rejects an empty `messages` list and the thinking config lives at `result["config"].thinking_config`, so neither test ever ran. The stand-in `ThinkingConfig` was also being coerced back into an all-`None` real one by `GenerateContentConfig` validation, so it asserted nothing about the SDK either. Same commit swaps the `gemini-3.5`/`gemini-4` prefix check for a version parse (CodeRabbit's point: `gemini-3.6` and `gemini-5` were falling through), and passes `types.ThinkingLevel` rather than a plain string, which the `dict[str, Any]` splat had been hiding from mypy.

Worth knowing rather than acting on: nobody can confirm yet whether 3.5 will accept `MINIMAL` and `MEDIUM` or only `LOW` and `HIGH` the way Gemini 3 does. If it lands narrower, the mapping needs a one line revisit.

Thanks for picking this up.

njbrake (approved):
Re-approving after the follow-up commits.

Two changes since my last approval. `google-genai>=1.51.0` is now pinned in the `gemini` and `vertexai` extras, which is the version that introduced `ThinkingConfig.thinking_level`, so the runtime `model_fields` check is gone and the parameter is chosen on the model version alone. Stating the requirement in the dependency declaration also fails at install time rather than as an opaque 400 from Google. Please leave it that way rather than restoring the capability sniffing.

I kept your two improvements from the meantime: the `ThinkingLevel` enum members in the mapping, and the `(?:^|/)` anchor, which correctly declines a full Vertex resource path. Tests now cover both sides of the 3.5 cutoff, including `gemini-3.4-flash` per CodeRabbit.

On CodeRabbit's remaining `ThinkingLevel` comment: it reads as stale, since its premise is that `google-genai` is unpinned and the same commit pins it. Its suggested fix would restore the check we removed, so I would skip it.

PR #1287, opened 2026-08-14
fix(otari): forward beta params and request IDs

### LINKED ISSUE #1225: Remove Otari beta guard after context management support ships
Blocked by mozilla-ai/otari#486.

After a fixed Otari release is available:

- bump the minimum `otari` version
- remove the temporary `context_management` and `betas` guard from `OtariProvider._amessages`
- add forwarding tests

The guard was introduced in #1224.

HareeshBahuleyan:
## Why

AnyLLM still blocks Otari context management parameters after SDK support shipped, and native Messages calls discard the Otari request ID needed for observability.

## What changed

Remove the temporary beta parameter guard, forward `context_management` and `betas`, and use the SDK response metadata API. Non-streaming responses expose `request_id`, while streaming responses attach it to the provider-neutral message-start event.

## Notes

- Stacked on #1286. Retarget this PR to `main` after the dependency PR merges.
- Verified with `otari 0.3.0` and `46` Otari provider unit tests.

Fixes #1225.

## PR Type

- Bug Fix
- New Feature

## Relevant issues

Fixes #1225.

## Checklist

- [x] **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

- AI Model used: GPT-5.6
- AI Developer Tool used: Pi

## Summary by CodeRabbit

* **New Features**
  * Responses now include the provider request ID when available, including streamed responses.
  * Anthropic beta parameters are forwarded for supported requests.
  * Structured output remains available through the existing completion experience.

* **Bug Fixes**
  * Improved handling of metadata-wrapped responses and streaming events.
  * Context management and beta parameters are now rejected only when incompatible with structured output.

### DIFF
--- a/src/any_llm/providers/otari/otari.py
+++ b/src/any_llm/providers/otari/otari.py
@@ -346,10 +346,6 @@ class OtariProvider(BaseOpenAIProvider):
         config). otari's gateway serves /messages natively, so delegate to the otari
         SDK's ``message()`` to preserve them.
         """
-        if params.context_management is not None or params.betas:
-            msg = "context_management and betas require a provider with a native Anthropic Messages API"
-            raise NotImplementedError(msg)
-
         if params.output_format is not None:
             # Structured output is handled by the base Messages<->Completions bridge, which
             # routes output_format through otari's completion path. A follow-up could adopt
--- a/src/any_llm/providers/otari/otari.py
+++ b/src/any_llm/providers/otari/otari.py
@@ -141,7 +141,7 @@ _MESSAGE_STREAM_EVENT_TYPES: dict[str, type[BaseModel]] = {
 }
 
 
-def _message_stream_event_from_dict(event: dict[str, Any]) -> MessageStreamEvent | None:
+def _message_stream_event_from_dict(event: dict[str, Any], request_id: str | None = None) -> MessageStreamEvent | None:
     """Validate a raw otari message SSE event dict into a typed MessageStreamEvent.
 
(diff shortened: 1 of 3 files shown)


njbrake (dismissed):
Approving. I verified the core behavior against real `otari==0.3.0` driven through a local gateway rather than the mocks: `context_management` and `betas` land in the POST body on both the streaming and non-streaming paths, and `request_id` comes back as `X-Otari-Request-ID` on both. The stream timing is right too; the SDK sets `AsyncOtariStream.request_id` before the first event is yielded, so reading it after the first `__anext__` is correct.

I pushed one commit (7d15c37) for the issue CodeRabbit raised. `output_format` combined with `context_management` or `betas` delegated to the base guard, which reports that the provider has no native Anthropic Messages API. That is false for otari; the Completions bridge just cannot carry those params. Same exception type, accurate message, plus a regression test for both params. Adopting otari's native `/messages` structured output (its `MessagesRequest` does have an `output_format` field) is the real fix and belongs in the follow-up your code comment already flags.

Two things worth a follow-up rather than a change here. `MessageResponse.request_id` currently has one producer: the Anthropic provider does a plain `model_validate` and never sets it, and otari's `with_response_metadata` also covers `completion` and `response`, which stay unwrapped. Worth deciding whether the field is otari-specific or the start of a general convention. Separately, `assert stream.request_id_reads == 1` pins an implementation detail; reading the attribute every iteration would be equally correct and would fail that assertion.

Nice work, and thanks for splitting the guard removal and the request ID work into separate commits.

Issue #1189, opened 2026-07-21
OpenAI provider passes reasoning_effort through unconditionally even when the model rejects it combined with tools


tbille:
**`any-llm` version:** 1.21.0

### Describe the bug
For some newer OpenAI reasoning models (e.g. `gpt-5.6-sol`, which OpenAI's docs describe as intended primarily for the Responses API), calling `acompletion()` with both `tools` and an explicit non-`"auto"` `reasoning_effort` raises a 400 from OpenAI:

```
Error code: 400 - {'error': {'message': "Function tools with reasoning_effort are not
supported for gpt-5.6-sol in /v1/chat/completions. To use function tools, use
/v1/responses or set reasoning_effort to 'none'.", 'type': 'invalid_request_error',
'param': 'reasoning_effort', 'code': None}}
```

`BaseOpenAIProvider._acompletion` (`providers/openai/base.py`) only special-cases `reasoning_effort == "auto"` -> `None`; any explicit value is passed straight through to `client.chat.completions.create()` regardless of whether `tools` is set or which model is targeted. any-llm currently has no signal that this model/param/tool combination will fail before the round trip to OpenAI.

### Reproduction
```python
from any_llm import acompletion

await acompletion(
    model="openai/gpt-5.6-sol",
    messages=[{"role": "user", "content": "What's the weather in Paris?"}],
    tools=[{"type": "function", "function": {"name": "get_weather", "parameters": {}}}],
    reasoning_effort="medium",
)
# -> openai.BadRequestError, param="reasoning_effort"
```

### Related prior art
This is the same category of problem as #270 / #271 (`gpt-4.1` rejecting `reasoning_effort` entirely) — a model-specific parameter-support gap the SDK currently can't see ahead of time. That fix addressed the *default* value; this case additionally depends on whether `tools` is present, so a default-only fix won't cover it.

### Possible directions (open to your take on the right approach)
1. Surface a clearer, earlier error from any-llm itself (e.g. detect the upstream `param="reasoning_effort"` + tools-present condition and raise a more actionable `any_llm` exception before/around the SDK call) rather than passing through OpenAI's raw error.
2. Since `BaseOpenAIProvider` already implements `_aresponses` for the same provider, consider whether `_acompletion` could transparently bridge through the Responses API for model/param combinations Chat Completions doesn't support, keeping the `acompletion()` call signature stable for callers.
[...]

Issue #1107, opened 2026-06-03
Reasoning effort levels for Anthropic


lanesket:
## Question

Is the `reasoning_effort: "xhigh"` → Anthropic `output_config.effort: "max"` mapping intentional, or a bug?

Anthropic exposes `xhigh` and `max` as **two distinct** effort levels (`low < medium < high < xhigh < max`), but  any-llm maps the canonical `xhigh` onto Anthropic `max`.

1. Requesting `xhigh` silently produces a **higher** level than asked (`max`) - more thinking tokens / cost than intended.
2. I can't use real xhigh? I guess
3. Anthropic's `max` is only reachable **by accident** (by sending `xhigh`), and not addressable on purpose - `"max"` isn't a valid `ReasoningEffort` input value.

`src/any_llm/providers/anthropic/utils.py`:
```python
REASONING_EFFORT_TO_ANTHROPIC_EFFORT = {
    "minimal": "low",
    "low": "low",
    "medium": "medium",
    "high": "high",
    "xhigh": "max",   # <-- here, skips the real xhigh
}
```

`src/any_llm/types/completion.py`:
```python
ReasoningEffort = Literal["none", "minimal", "low", "medium", "high", "xhigh", "auto"]
# no "max" -> Anthropic's top level cannot be requested explicitly
```

May be I'm wrong, just want to know what's the logic behind it

PR #708, opened 2026-01-05
feat: add ReasoningEffort alias, and support 'none' as str value

njbrake:
## Description
You were allowed to set reasoning_effort to `None`, but not to the str `none`. This PR 

1. Adds support for passing 'none' (the str value)
2. Consolidates the reasoning_effort into a single alias that can be referenced

This allows for continued support for OpenAI API which introduced 'none' as an accepted input to reasoning_effort.

## Checklist
- [x] I have added unit tests that prove my fix/feature works
- [x] New and existing tests pass locally
- [x] Documentation was updated where necessary
- [x] I have read and followed the [contribution guidelines](https://github.com/mozilla-ai/any-llm/blob/main/CONTRIBUTING.md)

---

> [!NOTE]
> Adds a normalized `ReasoningEffort` alias (now includes `'none'`) and propagates consistent handling across the API and providers.
> 
> - **Types/API**: Define `ReasoningEffort` in `types/completion.py`; update `CompletionParams`, `AnyLLM.acompletion`, and `api.completion/acompletion` to use it
> - **Provider behavior**: Standardize reasoning handling
>   - Anthropic: set `thinking={type:'disabled'}` when `reasoning_effort` is `None`/`'none'`; map effort to budgets otherwise
>   - Gemini: set `thinking_config` with `include_thoughts=False` for `None`/`'none'`; budgeted config for explicit levels
>   - Bedrock: treat `'none'` like no reasoning (omit `additionalModelRequestFields`)
>   - Azure, HuggingFace, Watsonx: coerce `'auto'`/`'none'` to `None`
>   - Cerebras, Cohere, Groq, Mistral, Ollama, Together, xAI, OpenRouter: drop `reasoning_effort` when `'auto'`/`'none'`; retain existing conversions
>   - OpenAI base: use aliased `ReasoningEffort` for defaults
> - **Tests**: Add/extend unit tests for Anthropic, AWS Bedrock, Cerebras, Cohere, Gemini, Groq, Mistral, Ollama, OpenRouter, Together, Watsonx, xAI to verify `'none'` disables or removes reasoning parameters
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1419cec6b2171baeae5b07d8e1ab0aad9c3deee1. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #1294, opened 2026-08-16
fix(gemini): reasoning_effort="none" disables thinking instead of hiding it

### LINKED ISSUE #1293: [BUG] Gemini: reasoning_effort="none" hides thoughts instead of disabling thinking
**`any-llm` version:** main (c0f3ffb); also present in 1.25.0
**Python Version**: 3.13

### Describe the bug

`_convert_completion_params` maps `reasoning_effort="none"` (and `None`) to `ThinkingConfig(include_thoughts=False)` ([base.py#L159-L161](https://github.com/mozilla-ai/any-llm/blob/c0f3ffb/src/any_llm/providers/gemini/base.py#L159-L161)). `include_thoughts` only hides thought summaries — the model still thinks and still bills those tokens. 

Disabling thinking is `thinking_budget=0`.

Every other provider that implements `"none"` genuinely disables thinking (anthropic sends `thinking={"type": "disabled"}`; cohere, deepseek, and ollama likewise). Gemini's implementation is inconsistent.

### Example w/ Code

```python
llm = AnyLLM.create("gemini")
resp = await llm.acompletion(
    model="gemini-2.5-flash",
    messages=[{"role": "user", "content": "What is 15% of 240? Think it through carefully before answering."}],
    reasoning_effort="none",
)
print(resp.usage)
# prompt_tokens=20  completion_tokens=168  total_tokens=736
```

The model thought anyway: 736 - 20 - 168 = **548 thought tokens billed** on a call that asked for none. (They show up in no usage field, mozilla-ai/any-llm#1290.)

The same question with thinking actually disabled through the raw SDK:

```python
client.models.generate_content(
    model="gemini-2.5-flash",
    contents="What is 15% of 240? Think it through carefully before answering.",
    config=GenerateContentConfig(thinking_config=ThinkingConfig(thinking_budget=0)),
)
# usage_metadata: prompt=20  candidates=263  thoughts=None  total=283
```

Same prompt, no thoughts, 736 vs 283 total tokens. `thinking_budget=0` is what `reasoning_effort="none"` should send.

PR incoming.

JamMaster1999:
## Description

`reasoning_effort="none"` (and `None`) currently maps to `ThinkingConfig(include_thoughts=False)`, which hides thought summaries instead of turning off thinking. 

`"none"` already means off everywhere else in this codebase:
- anthropic sends `thinking={"type": "disabled"}`, 
- cohere and deepseek send their disable flags, 
- ollama sends `think=False`,
- `messages_compat.py` maps Anthropic's `disabled` back to `reasoning_effort="none"`. 
 
**Gemini is the one provider where "none" means "keep thinking just hide the thinking summaries".**

The fix, per path:

- **`thinking_budget` models**: send `ThinkingConfig(include_thoughts=False, thinking_budget=0)`, so thinking is actually off.
- **Models that reject `thinking_budget=0`** (e.g. `gemini-2.5-pro`): Google's own 400 surfaces as the typed error, reacting to the provider's error rather than predicting it with a model list. 
   - This is the chosen approach for this repo (mozilla-ai/any-llm#1189). 
   - The live error is self-explanatory: `[gemini] 400 INVALID_ARGUMENT ... "Budget 0 is invalid. This model only works in thinking mode."`
- **`thinking_level` models (3.5+)**: `thinking_level` has no off value since the lowest tier is `MINIMAL`.
   - So `"none"` maps to `ThinkingLevel.MINIMAL` with a `logger.warning`, the same clamp-to-nearest convention already used in mozilla-ai/any-llm#1241 and mozilla-ai/any-llm#1108.

<details>
<summary>Why clamp with warning, instead of error?</summary>

On 3.5/3.6 the clamp billed zero thought tokens (table below), so `"none"` callers get what they asked for. Erroring up front instead would mean guessing from the version string which models those are, the kind of guess mozilla-ai/any-llm#1189 declined. A model that doesn't support `MINIMAL` rejects it, and the caller gets the typed 400 (3.7, note below).

The warning is because `MINIMAL` is the least thinking the model allows, not a promise of zero, and mozilla-ai/any-llm#1107 treats surprise thinking costs as a bug.

</details>

`None` stays merged with `"none"` as it has been since mozilla-ai/any-llm#708; `"auto"` remains the unset default, so callers who don't set the parameter are untouched.

Verified live across every path with `reasoning_effort="none"` (gap = thought tokens billed but hidden; before the fix, `gemini-2.5-flash` billed 424 hidden thought tokens):

| gemini models | path | result |
|---|---|---|
| 2.5-flash, 2.5-flash-lite, 3-flash-preview, 3.1-flash-lite | budget → 0 | OK, gap 0 |
| 2.5-pro, 3.1-pro-preview | budget → 0, rejected | typed `InvalidRequestError`: "Budget 0 is invalid. This model only works in thinking mode." |
| 3.5-flash, 3.5-flash-lite, 3.6-flash | level → MINIMAL | OK, gap 0, warning emitted |
| 3.7-flash | level → MINIMAL, rejected | typed `InvalidRequestError`: "Thinking level MINIMAL is not supported", details below |

Every failure surfaces as a typed any-llm exception; no raw `google.genai` errors leak.

<details>
<summary>Why gemini-3.7-flash rejects the clamp</summary>

`gemini-3.7-flash` cannot disable thinking: `MINIMAL` is gone, the new floor `LOW` thinks for real (220 thought tokens on the test prompt), and `thinking_budget=0` is silently ignored but still billed. So the 400 is correct, same as the thinking-only pros, and clamping to `LOW` would just bring back the bug this PR fixes. The removal also breaks the existing `"minimal"` mapping (mozilla-ai/any-llm#1281) on 3.7.

For reference, `thinking_budget=0` today is honored on 3.5, rejected on 3.6, and ignored-and-billed on 3.7. Three behaviors in three consecutive minor versions is why this PR trusts the SDK enum and provider errors instead of a model matrix.

</details>

### Behavior change

`reasoning_effort="none"` previously kept thinking active (billed, hidden); it now genuinely disables it. Models that cannot run without thinking now raise Google's 400 instead of silently billing. Callers who relied on "hidden but active" thinking should use `"auto"` or a graded level. Default (`"auto"`) behavior is unchanged.

## PR Type

- 🐛 Bug Fix

## Relevant issues

Fixes #1293

## Checklist

- [x] **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling when reasoning is disabled for Gemini models.
  * Models without an off setting now use minimal thinking and provide a one-time warning.
  * Improved conversion of Gemini and asynchronous API errors into clearer request errors, including HTTP status details.

* **Documentation**
  * Clarified how to disable reasoning and which settings to use for models that require thinking.

* **Tests**
  * Expanded coverage for reasoning configurations, unsupported thinking settings, and API error handling.

### DIFF
--- a/src/any_llm/providers/gemini/base.py
+++ b/src/any_llm/providers/gemini/base.py
@@ -8,6 +8,7 @@ from typing_extensions import override
 
 from any_llm.any_llm import AnyLLM
 from any_llm.exceptions import BatchNotCompleteError, InvalidRequestError, UnsupportedParameterError
+from any_llm.logging import logger
 from any_llm.types.completion import (
     ChatCompletion,
     ChatCompletionChunk,
@@ -148,7 +149,17 @@ class GoogleProvider(AnyLLM):
             kwargs["presence_penalty"] = params.presence_penalty
         if params.reasoning_effort != "auto":
             if params.reasoning_effort is None or params.reasoning_effort == "none":
-                kwargs["thinking_config"] = types.ThinkingConfig(include_thoughts=False)
+                if _uses_thinking_level(params.model_id):
+                    # thinking_level has no off tier; MINIMAL is the nearest expressible value
+                    logger.warning(
+                        "%s cannot disable thinking; clamping reasoning_effort='none' to thinking_level=MINIMAL",
+                        params.model_id,
+                    )
+                    kwargs["thinking_config"] = types.ThinkingConfig(
+                        include_thoughts=False, thinking_level=types.ThinkingLevel.MINIMAL
+                    )
+                else:
+                    kwargs["thinking_config"] = types.ThinkingConfig(include_thoughts=False, thinking_budget=0)
             elif _uses_thinking_level(params.model_id):
                 kwargs["thinking_config"] = types.ThinkingConfig(
                     include_thoughts=True, thinking_level=REASONING_EFFORT_TO_THINKING_LEVELS[params.reasoning_effort]
--- a/src/any_llm/providers/gemini/base.py
+++ b/src/any_llm/providers/gemini/base.py
@@ -71,6 +71,7 @@ REASONING_EFFORT_TO_THINKING_LEVELS = {
 _SUPPORTED_BATCH_ENDPOINTS = frozenset({"/v1/chat/completions"})
 _THINKING_LEVEL_MIN_GEMINI_VERSION = (3, 5)
 _GEMINI_VERSION_PATTERN = re.compile(r"(?:^|/)gemini-(\d+)(?:\.(\d+))?")
+_MINIMAL_CLAMP_WARNED: set[str] = set()
 
 
 def _uses_thinking_level(model_id: str) -> bool:
@@ -150,11 +151,12 @@ class GoogleProvider(AnyLLM):
         if params.reasoning_effort != "auto":
             if params.reasoning_effort is None or params.reasoning_effort == "none":
                 if _uses_thinking_level(params.model_id):
-                    # thinking_level has no off tier; MINIMAL is the nearest expressible value
-                    logger.warning(
-                        "%s cannot disable thinking; clamping reasoning_effort='none' to thinking_level=MINIMAL",
-                        params.model_id,
-                    )
+                    if params.model_id not in _MINIMAL_CLAMP_WARNED:
+                        _MINIMAL_CLAMP_WARNED.add(params.model_id)
+                        logger.warning(
+                            "%s has no thinking off tier; clamping reasoning_effort='none' to thinking_level=MINIMAL",
+                            params.model_id,
+                        )
                     kwargs["thinking_config"] = types.ThinkingConfig(
                         include_thoughts=False, thinking_level=types.ThinkingLevel.MINIMAL
                     )
--- a/tests/unit/providers/test_gemini_exceptions.py
+++ b/tests/unit/providers/test_gemini_exceptions.py
@@ -1,6 +1,7 @@
 # ruff: noqa: E402
 from __future__ import annotations
 
+import httpx
 import pytest
 
 google_genai = pytest.importorskip("google.genai")
@@ -63,4 +64,37 @@ def test_api_error_generic() -> None:
 
     assert isinstance(result, ProviderError)
     assert result.provider_name == "gemini"
+    assert result.original_exception is original
+
+
+def test_client_error_when_model_rejects_zero_thinking_budget() -> None:
+    """Models that only work in thinking mode reject thinking_budget=0 with a 400."""
+    original = ClientError(
+        code=400,
+        response_json={"error": {"message": "Budget 0 is invalid. This model only works in thinking mode."}},
+    )
+
+    result = convert_exception(original, "gemini")
+
+    assert isinstance(result, InvalidRequestError)
+    assert "Budget 0 is invalid" in str(result)
+    assert result.original_exception is original
+
+
+def test_client_error_when_model_rejects_minimal_thinking_level() -> None:
+    """Models without a MINIMAL thinking level reject the clamp with a 400."""
+    original = ClientError(
+        code=400,
+        response_json={
+            "error": {
+                "message": "Thinking level MINIMAL is not supported for this model. Please retry with other thinking level."
+            }
+        },
+        response=httpx.Response(400),
+    )
+
+    result = convert_exception(original, "gemini")
+
+    assert isinstance(result, InvalidRequestError)
+    assert "Thinking level MINIMAL is not supported" in str(result)
     assert result.original_exception is original
--- a/tests/unit/providers/test_gemini_exceptions.py
+++ b/tests/unit/providers/test_gemini_exceptions.py
@@ -98,3 +98,22 @@ def test_client_error_when_model_rejects_minimal_thinking_level() -> None:
     assert isinstance(result, InvalidRequestError)
     assert "Thinking level MINIMAL is not supported" in str(result)
     assert result.original_exception is original
+
+
+def test_client_error_without_attached_response_still_maps_to_invalid_request() -> None:
+    """With no usable response object, the status Google puts in the error body still classifies the 400."""
+    original = ClientError(
+        code=400,
+        response_json={
+            "error": {
+                "code": 400,
+                "message": "Thinking level MINIMAL is not supported for this model. Please retry with other thinking level.",
+                "status": "INVALID_ARGUMENT",
+            }
+        },
+    )
+
+    result = convert_exception(original, "gemini")
+
+    assert isinstance(result, InvalidRequestError)
+    assert "Thinking level MINIMAL is not supported" in str(result)
--- a/tests/unit/providers/test_gemini_provider.py
+++ b/tests/unit/providers/test_gemini_provider.py
@@ -662,25 +662,30 @@ def test_older_gemini_models_keep_thinking_budget(model_id: str) -> None:
 
 
 @pytest.mark.parametrize("reasoning_effort", [None, "none"])
-def test_none_disables_thinking_on_thinking_budget_models(reasoning_effort: ReasoningEffort | None) -> None:
-    """Test that reasoning_effort='none' sends thinking_budget=0 instead of only hiding thoughts."""
-    result = GoogleProvider._convert_completion_params(
-        CompletionParams(
-            model_id="gemini-2.5-flash",
-            messages=[{"role": "user", "content": "Hello"}],
-            reasoning_effort=reasoning_effort,
-        ),
-        provider_name="gemini",
-    )
+def test_none_disables_thinking_on_thinking_budget_models(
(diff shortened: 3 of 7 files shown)


JamMaster1999:
Good catch from the bot, one correction on scope: the missing enum members are pre-existing on main, not introduced here. #1281 set the `google-genai>=1.51.0` floor while its module-level `REASONING_EFFORT_TO_THINKING_LEVELS` references `ThinkingLevel.MINIMAL` and `MEDIUM` — but 1.51.0's enum only has `LOW` and `HIGH` (both missing members first appear in 1.56.0). So importing the provider at the declared minimum already fails on main today. Since this PR adds another reference, I've bumped the floor to `>=1.56.0` here and added regression tests for both rejection paths.

tests/unit/providers/test_gemini_exceptions.py
```diff
@@ -64,3 +65,36 @@ def test_api_error_generic() -> None:
...
+def test_client_error_when_model_rejects_minimal_thinking_level() -> None:
+    """Models without a MINIMAL thinking level reject the clamp with a 400."""
+    original = ClientError(
+        code=400,
+        response_json={
+            "error": {
+                "message": "Thinking level MINIMAL is not supported for this model. Please retry with other thinking level."
+            }
+        },
+        response=httpx.Response(400),
```
dpoulopoulos, line 93: **[minor]** This test only passes because the attached `httpx.Response` carries `status_code`. When `aiohttp` is importable, google-genai uses it for async calls and attaches an `aiohttp.ClientResponse`, which exposes `status`, not `status_code`; `_extract_status_code` then returns `None`, and the message "Thinking level MINIMAL is not supported..." matches no pattern in `_ERROR_PATTERNS`, so the caller gets `ProviderError` instead of `InvalidRequestError`. Since the clamp relies on that typed error as its only safety net, let `_extract_status_code` also read `response.status`, and add a case without an attached response.

src/any_llm/providers/gemini/base.py
```diff
@@ -158,7 +159,17 @@ def _convert_completion_params(params: CompletionParams, **kwargs: Any) -> dict[
             kwargs["presence_penalty"] = params.presence_penalty
         if params.reasoning_effort != "auto":
             if params.reasoning_effort is None or params.reasoning_effort == "none":
-                kwargs["thinking_config"] = types.ThinkingConfig(include_thoughts=False)
+                if _uses_thinking_level(params.model_id):
+                    # thinking_level has no off tier; MINIMAL is the nearest expressible value
+                    logger.warning(
+                        "%s cannot disable thinking; clamping reasoning_effort='none' to thinking_level=MINIMAL",
```
dpoulopoulos: **[minor]** The text says the model "cannot disable thinking", but your own measurements show MINIMAL bills zero thought tokens on 3.5 and 3.6, so the warning reads worse than the truth for the common case. It also fires on every request, which floods logs for a caller that sets `reasoning_effort="none"` on each call. Reword it as a clamp to the lowest tier, and consider warning once per model.

src/any_llm/providers/gemini/base.py
```diff
@@ -158,7 +159,17 @@ def _convert_completion_params(params: CompletionParams, **kwargs: Any) -> dict[
             kwargs["presence_penalty"] = params.presence_penalty
         if params.reasoning_effort != "auto":
             if params.reasoning_effort is None or params.reasoning_effort == "none":
```
dpoulopoulos, line 152: **[nit]** Callers who used `reasoning_effort="none"` with a thinking-only model such as gemini-2.5-pro now get a 400 where the call used to succeed. The reasoning section of `docs/quickstart.md` says nothing about this. One line there would save users a support round trip.

tests/unit/providers/test_gemini_provider.py
```diff
@@ -644,6 +644,42 @@ def test_older_gemini_models_keep_thinking_budget(model_id: str) -> None:
...
+                messages=[{"role": "user", "content": "Hello"}],
+                reasoning_effort=reasoning_effort,
+            ),
+            provider_name="gemini",
+        )
+
+    assert result["config"].thinking_config == types.ThinkingConfig(
+        include_thoughts=False, thinking_level=types.ThinkingLevel.MINIMAL
+    )
+    assert "cannot disable thinking" in caplog.text
```
dpoulopoulos: **[nit]** The tests check that the warning appears for a thinking-level model, but nothing checks that the budget path stays silent. A change that warns for every model would still pass. Add `assert not caplog.text` (or `caplog.records == []`) to the budget-path test.

dpoulopoulos (commented):
Thanks @JamMaster1999. Good, well-scoped fix. The two paths are right: `thinking_budget=0` for budget models and the `MINIMAL` clamp for 3.5+ models, and the `google-genai` floor of 1.56.0 is correct because `ThinkingLevel.MINIMAL` and `MEDIUM` first exist in that release, so the module-level map broke the provider import on `1.51` to `1.55`.

I checked correctness of both branches, the exception mapping, the unit tests, the docs, and the effect on the integration suite (it only uses `reasoning_effort="low"`, so nothing there changes).

Two small points: the `MINIMAL` rejection maps to `InvalidRequestError` only when the SDK attaches an `httpx` response, and the per-call warning text overstates the cost.

tests/unit/providers/test_gemini_exceptions.py
JamMaster1999: `_extract_status_code` now reads `response.status` as well, pinned by `test_convert_exception_carries_status_from_aiohttp_style_response`. Also added `test_client_error_without_attached_response_still_maps_to_invalid_request`, with the body Google actually sends (`status: INVALID_ARGUMENT`); with no response attached that classifies through the message patterns. The synthetic body in the earlier test was what lacked the status, so on the live aiohttp path the 400 was already typed, but the handler no longer depends on that. 2c76c9d.

src/any_llm/providers/gemini/base.py
JamMaster1999: Reworded to `<model> has no thinking off tier; clamping reasoning_effort='none' to thinking_level=MINIMAL`, and it now fires once per model through a module-level set. `test_minimal_clamp_warns_once_per_model` pins that. 2c76c9d.

src/any_llm/providers/gemini/base.py
JamMaster1999: Added one line to the reasoning section of `docs/quickstart.md`: `"none"` (or `None`) turns thinking off, and a model that only runs with thinking on, such as `gemini-2.5-pro`, rejects it with `InvalidRequestError`. 2c76c9d.

tests/unit/providers/test_gemini_provider.py
JamMaster1999: Added `assert not caplog.records` to the budget-path test. 2c76c9d.

JamMaster1999:
@dpoulopoulos all four points addressed in 2c76c9d, and main is merged in so `run-linter` picks up the `test_openai_exceptions.py` fix that was failing here. Ready for another look.

src/any_llm/providers/gemini/base.py
```diff
@@ -158,7 +159,17 @@ def _convert_completion_params(params: CompletionParams, **kwargs: Any) -> dict[
...
-                kwargs["thinking_config"] = types.ThinkingConfig(include_thoughts=False)
+                if _uses_thinking_level(params.model_id):
+                    # thinking_level has no off tier; MINIMAL is the nearest expressible value
+                    logger.warning(
+                        "%s cannot disable thinking; clamping reasoning_effort='none' to thinking_level=MINIMAL",
+                        params.model_id,
+                    )
+                    kwargs["thinking_config"] = types.ThinkingConfig(
+                        include_thoughts=False, thinking_level=types.ThinkingLevel.MINIMAL
+                    )
```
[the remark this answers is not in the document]
JamMaster1999: Addressed in 988a023 on 2026-08-16: the `google-genai` floor is `>=1.56.0` in both extras, which is the first release with `ThinkingLevel.MINIMAL` and `MEDIUM`. The enum was already referenced on main by #1281 under the old `1.51.0` floor, so the bump also fixes the provider import at the declared minimum.

JamMaster1999:
@dpoulopoulos @njbrake this is ready for another look. All four review points were addressed in 2c76c9d (aiohttp `status`, once-per-model warning, docs line, silent budget path) and it merges clean onto 1.27.0. 

We consume any-llm downstream and carry this on a fork until it lands, so a review would be much appreciated. CI still needs a workflow approval.

JamMaster1999:
Closing this. #1381 landed the core of it: `reasoning_effort="none"` now sends `thinking_budget=0` on the budget models and rejects on the thinking-only ones, which is what this PR argued for. The one piece it didn't pick up is the aiohttp `status` read in `_extract_status_code`, which is now its own small PR: https://github.com/mozilla-ai/any-llm/pull/1387

PR #1296, opened 2026-08-17
fix(gemini): surface inline_data image parts in both converters

### LINKED ISSUE #1295: [BUG] Gemini converter drops inline_data parts
**Python Version**: 3.13

### Describe the bug

Two symptoms:

### Example w/ Code

```python
llm = AnyLLM.create("gemini")
resp = await llm.acompletion(
)
print(resp.usage)
```

```
```

```python
)])
# []
```

### Proposed fix

tonycoder-hub:
## Description

Gemini response converters dropped `inline_data` parts, so image-model output was lost (text+image kept text only; image-only produced no choice).

Both converters now render `image/*` inline blobs as OpenAI-style `image_url` data URIs on `message.images` / `ChoiceDelta.images`. Non-image blobs are left alone (see #1295).

## PR Type

- 🐛 Bug Fix

## Relevant issues

Fixes #1295

## Checklist

- [x] **AI Usage:**

`uv run pytest tests/unit`: 2164 passed, 69 skipped. pre-commit/mypy clean. No live Gemini call (no API key).

## AI Usage Information

- AI Developer Tool used: Cursor Cloud Agent

## Summary by CodeRabbit

* **New Features**
  * Added support for generated images in chat completions, including image-only responses, multiple images, image iteration, and streaming image updates.
  * Images are available as OpenAI-compatible base64 `image_url` content blocks and can be included in follow-up messages.

* **Documentation**
  * Added a quickstart section explaining generated image handling, requested modalities, streaming, and follow-up conversations.


JamMaster1999:
Verified this live on `gemini-2.5-flash-image`: `message.images` populated, image-only responses produce a proper choice, streaming works, and the block round-trips as input. Looks right to me as the #1295 author, and I'll open the audio half as a follow-up once this merges.

tonycoder-hub:
Closing as stale — opened on or before 2026-08-17 and still unmerged.

PR #1308, opened 2026-08-17
fix(responses): accept per-request timeout on aresponses

JamMaster1999:
## Description

`aresponses()` has no `timeout` parameter, and `ResponsesParams` is `extra="forbid"`, so a caller-supplied timeout dies in validation before any request is made:

```
ValidationError: 1 validation error for ResponsesParams
timeout
  Extra inputs are not permitted [type=extra_forbidden, input_value=60, input_type=int]
```

mozilla-ai/any-llm#1263 added per-request timeouts and wired them into `acompletion` and `amessages`. `aresponses` was left out, so the Responses path is the one call path with no per-request timeout; a client-wide timeout via `client_args` is the only workaround.

The fix repeats the #1263 pattern on `aresponses`, unchanged:

- `timeout: float | None = None` on the signature, with the same `# noqa: ASYNC109` convention as `acompletion` / `amessages`.
- Routed through `_validate_and_forward_timeout` into the provider kwargs. From there `TIMEOUT_SUPPORT` governs it: `native` providers hand it to their SDK as a request option; `unsupported` providers raise `UnsupportedParameterError`.
- It never enters `ResponsesParams` — per #1263, timeout is a transport option, not request-body JSON.

No provider changes are needed: every `_aresponses` implementation already forwards `**kwargs` into its SDK call.

Verified live against OpenAI `gpt-5.1`:

| call | before | after |
|---|---|---|
| `aresponses(..., timeout=0.001)` | `ValidationError` | `ProviderError: [openai] Request timed out.` |
| `aresponses(..., timeout=60)` | `ValidationError` | completes normally |

The new unit test asserts both halves of the routing: the value reaches the `_aresponses` call kwargs, and never appears in the params model.

## PR Type

- 🐛 Bug Fix

## Relevant issues

Follow-up to mozilla-ai/any-llm#1263.

## Checklist

- [x] **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

## Summary by CodeRabbit

* **New Features**
  * Added optional timeout settings for synchronous and asynchronous Responses API requests.
  * Timeout values are forwarded to supported providers, helping requests complete within the specified duration.
  * Timeout handling is consistently available across the Responses API and its helper methods.

* **Bug Fixes**
  * Unsupported providers now reject timeout settings clearly before any request is initiated.

### DIFF
--- a/src/any_llm/any_llm.py
+++ b/src/any_llm/any_llm.py
@@ -1270,7 +1270,11 @@ class AnyLLM(ABC):
             prompt_cache_key: A key to use when reading from or writing to the prompt cache.
             prompt_cache_retention: How long to retain a prompt cache entry created by this request.
             conversation: The conversation to associate this response with (ID string or ConversationParam object).
-            timeout: Per-request timeout in seconds, forwarded to the provider's SDK call.
+            timeout: Per-request timeout in seconds, passed through to the provider's client/SDK.
+                An explicit ``None`` is treated the same as omitting it (the provider's default
+                applies), so it cannot request an unbounded timeout. Providers that have no
+                per-request timeout raise `UnsupportedParameterError`; set a timeout on their
+                client via `client_args` instead.
             extra_body: Additional fields to merge into an OpenAI-compatible Responses request body.
             **kwargs: Additional provider-specific arguments that will be passed to the provider's API call.
 
--- a/src/any_llm/any_llm.py
+++ b/src/any_llm/any_llm.py
@@ -1222,6 +1222,7 @@ class AnyLLM(ABC):
         prompt_cache_key: str | None = None,
         prompt_cache_retention: str | None = None,
         conversation: str | dict[str, Any] | None = None,
+        timeout: float | None = None,  # noqa: ASYNC109  # forwarded to the provider SDK, which owns the timeout
         extra_body: dict[str, Any] | None = None,
         **kwargs: Any,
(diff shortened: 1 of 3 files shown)


njbrake (approved):
Approving. The fix is right: `timeout` now routes through `_validate_and_forward_timeout` into the provider kwargs, so it never reaches the `extra="forbid"` `ResponsesParams`. I confirmed the bug reproduces on the base commit and that the new test fails without the fix.

I rebased onto main (the branch had conflicted with #1310) and pushed one commit on top:

- `api.py`: `timeout` is now declared on the public `responses` and `aresponses` helpers. It already worked via `**kwargs`, but it was the one parameter on those two functions that type checkers and the docstring never mentioned, unlike `completion` / `acompletion` / `messages` / `amessages` from #1263.
- Restored the docstring wording the sibling methods use, including the `UnsupportedParameterError` note.
- Tests for signature exposure, an explicit `None` staying unset, both helpers, and the unsupported-provider rejection.

One follow-up worth filing separately. `TIMEOUT_SUPPORT` is a single per-provider flag, classified against the completion path, but a provider can sit behind a different SDK on the Responses path. HuggingFace is the live case: its `responses_client` is a plain `AsyncOpenAI` whose `responses.create()` accepts a per-request timeout, yet it inherits `unsupported` from the `AsyncInferenceClient` completion path, so `aresponses(..., timeout=...)` is rejected on a path that would have honored it. It fails closed, so nothing is at risk, but the classification is wrong for that path. @peteski22, this is your subsystem, so I would value your read on whether the capability should become path aware.

Thanks @JamMaster1999.

PR #1300, opened 2026-08-17
fix(gemini): map tool_choice none and named-function form

tonycoder-hub:
## Description
Gemini `_convert_tool_choice` only knew `auto` and `required`. `"none"` raised `KeyError`, and the OpenAI named-function dict was ignored because the call site only accepted strings, so Gemini defaulted to AUTO.

`"none"` now maps to `FunctionCallingConfigMode.NONE`. The named-function dict maps to `mode=ANY` with `allowed_function_names`. Unrecognized values raise `UnsupportedParameterError`. VertexAI inherits the same conversion.

any-llm already emits both values from Anthropic `messages()` via `_convert_tool_choice_to_openai`.

## PR Type
- Bug Fix

## Relevant issues
No open issue. Distinct from #1291/#1294/#1296.

## Checklist
- [x] I understand the code I am submitting.
- [x] I have added unit tests that prove my fix/feature works
- [x] I have run this code locally and verified it fixes the issue.
- [x] New and existing tests pass locally
- [x] Documentation was updated where necessary
- [x] I have read and followed the contribution guidelines
- [x] **AI Usage:**
    - [x] This is fully AI-generated.

## AI Usage Information
- AI Model used: Claude Opus 5
- AI Developer Tool used: Cursor cloud agent
- [x] I am an AI Agent filling out this form (check box if true)

`uv run pytest tests/unit -q -n auto` → 2160 passed, 69 skipped.

## Summary by CodeRabbit

- **New Features**
  - Added support for Gemini tool-selection modes, including automatic, required, disabled, and specific-function selections.
  - Added support for selecting required tools from an allowed-tools list.

- **Bug Fixes**
  - Unsupported or malformed tool-selection values now produce clear, provider-specific errors.

- **Tests**
  - Added coverage for disabled tools, named-function selection, allowed-tool selections, and invalid settings.

PR #1309, opened 2026-08-17
fix(gemini): convert native tool dicts instead of silently dropping them

JamMaster1999:
## Description

`_convert_tool_spec` skips any dict whose `type` isn't `"function"`, so a gemini-native tool dict like `{"google_search": {}}` silently vanishes from the request — no error, no search. The typed form of the same tool already passes through via `BUILT_IN_TOOLS`; the dict spelling is the one form that dies.

Non-function dicts now go through `types.Tool.model_validate`: any gemini built-in converts, and a dict the SDK schema rejects raises a typed `InvalidRequestError`. `provider_name` is threaded into the converter for the error message, as #1300 does for `_convert_tool_choice`.

Verified live on `gemini-2.5-flash` with "What was the S&P 500's closing value last Friday?" and `tools=[{"google_search": {}}]`:

| | result |
|---|---|
| before | tool silently dropped — answers `4997.58` from stale training data |
| after | grounded — "closing value last Friday, August 14, 2026, was 7,785.76" |

Tests: a native dict reaches the request as a typed Tool; function and native tools mix in one call; unknown dicts raise. Existing direct-caller tests updated for the new argument.

## PR Type

- 🐛 Bug Fix

## Relevant issues

Same silent-drop family as mozilla-ai/any-llm#1300.

## Checklist

- [x] **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of Gemini tool specifications, including native tool dictionaries and function tools.
  * Added clearer validation errors for unsupported or invalid tool definitions.
  * Preserved valid Gemini-native tools in mixed tool configurations.
  * Improved reliability when processing incomplete or invalid tool schemas.

* **Tests**
  * Expanded coverage for missing parameters, missing properties, schema references, invalid tool dictionaries, and mixed tool configurations.

### DIFF
--- a/src/any_llm/providers/gemini/base.py
+++ b/src/any_llm/providers/gemini/base.py
@@ -164,7 +164,7 @@ class GoogleProvider(AnyLLM):
         if params.temperature is not None:
             kwargs["temperature"] = params.temperature
         if params.tools is not None:
-            kwargs["tools"] = _convert_tool_spec(params.tools)
+            kwargs["tools"] = _convert_tool_spec(params.tools, provider_name)
         if isinstance(params.tool_choice, str):
             kwargs["tool_config"] = _convert_tool_choice(params.tool_choice)
         if params.top_p is not None:
--- a/src/any_llm/providers/gemini/utils.py
+++ b/src/any_llm/providers/gemini/utils.py
@@ -7,6 +7,7 @@ from typing import Any, Literal, cast
 
 from google.genai import types
 from google.genai.pagers import Pager
+from pydantic import ValidationError
 
 from any_llm.exceptions import InvalidRequestError
 from any_llm.logging import logger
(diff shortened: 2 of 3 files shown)


njbrake (approved):
Approving. The fix sits in the one converter every Gemini completion goes through, and it covers `vertexai` too, since `VertexaiProvider` shares `GoogleProvider`. Validating against the SDK schema rather than a name table is the right call; an allowlist of `google_search` / `url_context` / `code_execution` would go stale on every SDK release.

Two things I verified rather than took on trust: `google.genai.types.Tool` sets `extra="forbid"`, so an unrecognized dict really does raise instead of validating into an empty no-op tool, and all four new tests fail against the pre-fix converter.

Non-blocking, fine as follow-ups:

- `{}` and `{"google_search": None}` still validate into an empty `types.Tool()` that gets sent. Not a regression, but the no-silent-drops principle does not quite reach them.
- The error message could name the accepted shapes. A flat function dict (`{"name": ..., "parameters": ...}`) now lands in this branch with no hint about the nested form.
- The same silent-drop shape lives in `anthropic/utils.py:407` for Anthropic server tools. Worth its own issue.

Worth flagging for anyone reading later: an unrecognized tool dict is now a hard error where it used to be ignored, so a shared `tools` list reused across providers will fail on Gemini rather than quietly losing a tool. That is the point of the change and it matches how we treat unusable request input elsewhere, but it is a behavior change.

I rebased onto current main and force-pushed; CI is green on `c10c43f`.

Issue #1053, opened 2026-04-24
[RFC] Standardizing Reasoning Continuity (Thinking Signatures) for Multi-Turn Agent Loops


liukidar:
## **Description**
Currently, `any-llm` handles reasoning models (Gemini, Claude, and OpenAI o-series) inconsistently across its primary interfaces (`acompletion`, `amessages`, and `aresponses`). While these models all utilize a "hidden state" or "thinking block" to maintain reasoning continuity, the mechanism for preserving this state across multi-turn tool-calling loops varies by provider.

To build reliable agents, we need a unified way to handle these "thinking" states so that the model doesn't "lose its train of thought" or fail validation when tools are called in sequence.

## **The Current Problem**
Based on recent testing, there is a "practical matrix" of gaps where reasoning metadata is dropped depending on which endpoint is used. Adding **`aresponses`** (the primitive for OpenAI’s Responses API) completes the picture:

| Provider | Mechanism | `acompletion` Status | `amessages` Status | `aresponses` Status |
| :--- | :--- | :--- | :--- | :--- |
| **Gemini** | `thought_signature` | **Supported** (Preserved) | **Weak** (Bridge drops signature; relies on fallback) | **Unsupported** (OpenAI-specific path) |
| **Claude** | `ThinkingBlock.signature` | **Weak** (Collapses blocks into plain text) | **Supported** (Native preservation) | **Unsupported** (OpenAI-specific path) |
| **OpenAI** | `previous_response_id` | **Unsupported** (No reasoning continuity) | **Unsupported** (Needs Responses API logic) | **Supported** (Ideal path for o-series continuity) |

### **Key Insights:**
* **Gemini** expresses continuity as `extra_content.google.thought_signature`.
* **Claude** uses a `signature` field within its thinking blocks.
* **OpenAI** reasoning continuity belongs to the **Responses API**, which tracks server-side state via `previous_response_id` rather than just passing message strings. 

This matrix highlights that `any-llm` currently requires users to switch between three different high-level methods depending on which provider they are calling if they want to maintain reasoning integrity.

## **Proposed Goals**
The goal is to move away from fragmented patches in `acompletion` and instead establish the **Messages-style path** as the canonical agentic path.

1.  **Unified Interface:** A request-level parameter (e.g., `preserve_thinking: bool`) that automatically handles the provider-specific logic for continuity.
2.  **State Preservation:**
    * Ensure `amessages` preserves Gemini's `thought_signature`.
[...]

PR #1313, opened 2026-08-19
fix(types): keep tool-call extra_content when a ChatCompletionMessage is dumped

JamMaster1999:
## Description

#1153 added `extra_content` to `ChatCompletionMessageFunctionToolCall` so gemini's `thought_signature` rides along with each tool call and gets replayed on the next turn (`gemini/utils.py:_convert_messages` reads it back). The instance carries it — but `ChatCompletionMessage` never re-declared `tool_calls`, so the field kept OpenAI's declared type, and `model_dump()` serializes against that: the subclass field is dropped.

`acompletion` dumps replayed message objects exactly this way (`any_llm.py:756`), so the documented loop — append `response.choices[0].message`, add the tool result, call again — reaches gemini with no signature. The converter then falls back to the `skip_thought_signature_validator` sentinel: no 400, but the model's reasoning context for that call is gone. `tests/integration/test_agent_loop.py` builds its history the same way. The streaming side was already right — `ChoiceDelta.tool_calls` was re-annotated in #1153; only the non-streaming message missed it.

The fix re-declares `tool_calls` with the any-llm union that already exists (`ChatCompletionMessageToolCall`), as a forward reference since the alias sits below the class. One line; validation from plain dicts is unchanged (function calls resolve to the any-llm subclass, `custom` tool calls to OpenAI's), and `ChatCompletion` round-trips still carry `extra_content`.

Nine converters had been casting their tool-call lists into the OpenAI-typed field through a `TYPE_CHECKING` alias (`ChatCompletionMessageToolCallType`). With the field typed as any-llm's own union those lists match directly and the casts point the wrong way for mypy, so the second commit deletes the aliases and casts — 9 files, no behavior change.

Verified live on `gemini-3-flash-preview`: turn 1 returns a `get_weather` call with a signature; the message object is appended and replayed with the tool result; the outgoing `Part.thought_signature` is captured at the SDK call:

| | outgoing thought_signature | equals the one returned |
|---|---|---|
| before | `skip_thought_signature_validator` sentinel | no |
| after | real signature | yes |

The test dumps a signed `ChatCompletionMessage` the way `acompletion` does and asserts the gemini converter emits the real signature. It fails on main.

## PR Type

- 🐛 Bug Fix

## Relevant issues

Completes #1153; related to the open RFC #1053.

## Checklist

- [x] **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved preservation of Google thought signatures when replaying saved chat messages.
  * Improved consistency when handling tool calls across supported providers.
  * Preserved tool-call metadata, including additional content, through message conversion and JSON round-tripping.
  * Improved validation and handling of malformed tool-call data.

* **Tests**
  * Added regression coverage for tool-call metadata, validation, serialisation, and thought-signature preservation.

### DIFF
--- a/src/any_llm/providers/anthropic/utils.py
+++ b/src/any_llm/providers/anthropic/utils.py
@@ -1,6 +1,6 @@
 import json
 from datetime import datetime
-from typing import TYPE_CHECKING, Any, cast
+from typing import Any, cast
 
 from anthropic import transform_schema
 from anthropic.types import (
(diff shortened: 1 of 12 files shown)


JamMaster1999:

PR #1314, opened 2026-08-19
fix(gemini): name tool results from the tool_call they answer

JamMaster1999:
## Description

`_convert_messages` names a gemini `function_response` from `message["name"]`, defaulting to `"unknown"`. OpenAI's tool message has no `name` — its keys are `role`, `content`, `tool_call_id` (`ChatCompletionToolMessageParam`) — so every standard tool loop reaches gemini as `name="unknown"`. The Messages bridge emits exactly that shape too (`messages_compat.py:192`), so `messages()` routed to gemini always sends `unknown`. The integration agent-loop test avoids it only by adding a `name` key by hand.

Gemini accepts the unnamed response, so this never errors. It costs accuracy when it matters: with two tools called in parallel and results supplied in a different order (legal on the OpenAI side, `tool_call_id` disambiguates), gemini has to guess which result answers which call.

The fix records `id → function name` while walking the assistant's `tool_calls` and uses it for tool messages that carry no `name`. An explicit `name` still wins; an unknown id still falls back to `"unknown"`.

Verified live with `get_temperature` / `get_population` for Toronto, results returned in reverse call order and both shaped `{"value": n}`, no `name` on the tool messages, 3 trials per model:

| | names sent | gemini-2.5-flash | gemini-3-flash-preview |
|---|---|---|---|
| before | `unknown, unknown` | 2/3 correct — one answer swapped: `temperature=2794356 population=8` | 3/3 |
| after | `get_population, get_temperature` | 3/3 | 3/3 |

The test feeds two tool results in reversed order without names and asserts the resolved names. It fails on main.

## PR Type

- 🐛 Bug Fix

## Relevant issues

None open.

## Checklist

- [x] **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of tool responses by resolving names from matching tool calls, including when results arrive out of order.
  * Explicit tool names now take precedence, with a safe fallback for missing or unrecognised names.

* **Tests**
  * Added regression coverage for tool-name resolution, explicit-name precedence, out-of-order responses, and unknown or missing tool-call IDs.

### DIFF
--- a/src/any_llm/providers/gemini/utils.py
+++ b/src/any_llm/providers/gemini/utils.py
@@ -198,6 +198,7 @@ def _convert_messages(
     """Convert messages to Google GenAI format."""
     formatted_messages = []
     system_instruction = None
+    tool_names: dict[str, str] = {}  # tool_call id -> function name, for tool results that carry no name
 
     for message in messages:
         if message["role"] == "system":
@@ -225,6 +226,7 @@ def _convert_messages(
                 parts = []
                 for i, tool_call in enumerate(message["tool_calls"]):
                     function_call = tool_call["function"]
+                    tool_names[tool_call.get("id", "")] = function_call["name"]
                     args = json.loads(function_call["arguments"]) if function_call["arguments"] else {}
 
                     # Extract thought_signature if present (OpenAI compatibility format)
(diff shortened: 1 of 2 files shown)


JamMaster1999:

PR #1316, opened 2026-08-19
fix(gemini): carry the text-part thought_signature across turns

JamMaster1999:
## Description

#1153 carries gemini's `thought_signature` on function calls. Gemini 3 also signs text: a text-only answer comes back with the signature on its final text part, and Google's guidance for that case is to return it on the next turn — recommended for reasoning quality, not enforced (no 400 if omitted, unlike function calls). Both gemini converters read the signature only inside the `function_call` branch, so a text turn's signature was dropped on the way out and, having nowhere to live on the message, never replayed.

Live on `gemini-3-flash-preview` and `gemini-3.1-pro-preview` (raw SDK): a reasoning question returns `[thought part, text part]` with the signature on the text part; `gemini-2.5-flash` signs nothing. On a stream both models deliver the signature on the last chunk in a part with **empty text** — the case Google's docs call out — which the streaming converter's `elif part.text:` skipped outright.

The fix mirrors what the anthropic converter already does for its signature: the text-part signature rides `message.extra_content` (`{"google": {"thought_signature": ...}}`) on the non-streaming message and `delta.extra_content` on the chunk that carries it, and `_convert_messages` attaches it back to the text `Part` when a text-only assistant turn is replayed. Function-call signatures keep riding the tool call as before. `_thought_signature_extra_content` is reused, so both spellings stay in sync.

Verified through `acompletion` on `gemini-3-flash-preview`:

| | `message.extra_content` | replayed text `Part.thought_signature` | stream: chunks with `delta.extra_content` |
|---|---|---|---|
| before | `None` | `None` | 0 |
| after | `{"google": {"thought_signature": ...}}` | the returned signature | 1 (the last chunk) |

Tests: a signed text part lands on `message.extra_content`; a signed empty-text part on a stream lands on `delta.extra_content`; a text-only assistant turn carrying `extra_content` replays its signature. All three fail on main.

## PR Type

- 🐛 Bug Fix

## Relevant issues

Extends #1153 to text turns; related to the open RFC #1053.

## Checklist

- [x] **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

## Summary by CodeRabbit

- **Bug Fixes**
  - Gemini completion responses now preserve thought-signature metadata.
  - Thought signatures are retained in both standard and streaming responses.
  - Assistant responses can replay preserved signatures in subsequent Gemini messages.

### DIFF
--- a/src/any_llm/providers/gemini/base.py
+++ b/src/any_llm/providers/gemini/base.py
@@ -250,6 +250,7 @@ class GoogleProvider(AnyLLM):
                 content=message_dict.get("content"),
                 tool_calls=cast("list[ChatCompletionMessageToolCallType] | None", tool_calls),
                 reasoning=Reasoning(content=reasoning_content) if reasoning_content else None,
+                extra_content=message_dict.get("extra_content"),
             )
             from typing import Literal
 
--- a/src/any_llm/providers/gemini/utils.py
+++ b/src/any_llm/providers/gemini/utils.py
@@ -364,7 +364,9 @@ def _convert_response_to_response_dict(response: types.GenerateContentResponse) 
         reasoning = None
         tool_calls_list: list[dict[str, Any]] = []
         text_content = None
-        message_extra_content = None  # gemini 3 signs the last text part too; it rides the message, like anthropic's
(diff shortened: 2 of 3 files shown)


JamMaster1999:

PR #1318, opened 2026-08-19
fix(gemini): name tool results from the tool_call they answer

JamMaster1999:
## Description

`_convert_messages` in `providers/gemini/utils.py` names each gemini `function_response` from `message["name"]` and falls back to `"unknown"`. OpenAI's tool message has no `name`: its keys are `role`, `content`, `tool_call_id` (`ChatCompletionToolMessageParam`). The Messages bridge emits that same shape (`utils/messages_compat.py`, `tool_result` → `{"role": "tool", "tool_call_id": ...}`). So every standard tool loop, through `completion()` or `messages()`, reached gemini with `name="unknown"` on every function response. The integration agent-loop test only avoids it by adding a `name` key by hand.

Gemini accepts the unnamed response, so nothing errors. But with `name="unknown"` on every response the model can only match results to calls by position. The OpenAI contract lets a caller return tool results in any order because `tool_call_id` disambiguates; gemini then attributes them to the wrong calls.

The fix records `tool_call id → function name` while walking the assistant's `tool_calls` and uses it for tool messages that carry no `name`. It lives in `_convert_messages` because that is the one place that sees both the calls and the results, for `completion()`, `messages()` and the batch converter alike. An explicit `name` still wins. A call without an id is not recorded, so a tool message whose id matches nothing, or that has no id, still gets `"unknown"`.

Live check: two tools, `get_red_code` and `get_blue_code`, both returning `{"code": n}`, called in parallel, results fed back in reverse call order with no `name` on the tool messages. Names that went out were captured by spying on `generate_content`. Same run through `completion()` and through the bridged `messages()` path, 3 trials each:

| | names sent | gemini-2.5-flash | gemini-3-flash-preview |
|---|---|---|---|
| before | `unknown, unknown` | 0/3 completion, 0/3 messages: every answer swapped (`red=9026, blue=4173`) | 0/3 completion, 0/3 messages: every answer swapped |
| after | `get_blue_code, get_red_code` | 4/4 completion (2 trials skipped: model called one tool at a time), 3/3 messages | 3/3 completion, 3/3 messages |

Two unit tests: one feeds two results in reversed order without names, one JSON and one plain string, and asserts the resolved names (fails on main); the other pins the fallbacks, explicit `name` wins and an unmatched or missing id stays `"unknown"`.

## PR Type

- 🐛 Bug Fix

## Relevant issues

None open.

## Checklist

- [x] **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of Gemini function responses when tool messages provide only a tool-call ID.
  - Preserved explicitly provided function names and used a safe fallback when no matching name is available.
  - Ensured function-response names remain correct regardless of tool-call ordering or response format.

- **Tests**
  - Added coverage for name resolution, precedence rules, unmatched tool-call IDs, and missing names.

### DIFF
--- a/src/any_llm/providers/gemini/utils.py
+++ b/src/any_llm/providers/gemini/utils.py
@@ -198,6 +198,7 @@ def _convert_messages(
     """Convert messages to Google GenAI format."""
     formatted_messages = []
     system_instruction = None
+    tool_names: dict[str, str] = {}  # tool_call id -> function name, for tool results that carry no name
 
     for message in messages:
         if message["role"] == "system":
@@ -225,6 +226,7 @@ def _convert_messages(
                 parts = []
                 for i, tool_call in enumerate(message["tool_calls"]):
                     function_call = tool_call["function"]
+                    tool_names[tool_call.get("id", "")] = function_call["name"]
                     args = json.loads(function_call["arguments"]) if function_call["arguments"] else {}
 
                     # Extract thought_signature if present (OpenAI compatibility format)
@@ -251,16 +253,13 @@ def _convert_messages(
 
             formatted_messages.append(types.Content(role="model", parts=parts))
         elif message["role"] == "tool":
+            name = message.get("name") or tool_names.get(message.get("tool_call_id", ""), "unknown")
             try:
                 content_json = json.loads(message["content"])
-                part = types.Part.from_function_response(
-                    name=message.get("name", "unknown"), response=_normalize_tool_response(content_json)
-                )
+                part = types.Part.from_function_response(name=name, response=_normalize_tool_response(content_json))
                 formatted_messages.append(types.Content(role="function", parts=[part]))
             except json.JSONDecodeError:
-                part = types.Part.from_function_response(
-                    name=message.get("name", "unknown"), response={"result": message["content"]}
-                )
+                part = types.Part.from_function_response(name=name, response={"result": message["content"]})
                 formatted_messages.append(types.Content(role="function", parts=[part]))
 
     return formatted_messages, system_instruction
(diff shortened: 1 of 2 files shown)


njbrake (approved):
Approving. The bug is real and the fix is in the right place.

I reverted `gemini/utils.py` to main and confirmed the new test fails there (`['unknown', 'unknown'] != ['get_population', 'get_temperature']`), then checked the fix against the shape a real agent loop produces rather than hand-built dicts: a `ChatCompletionMessage` dumped the way `any_llm.py:756` does it, through `CompletionParams` and `_convert_messages`, with results fed back in reverse order. Names pair correctly. `_convert_messages` has only two callers, so this one edit covers `completion()`, the `messages()` bridge, the batch converter and `vertexai`.

The approach matches `ollama.py:266-286`, which solves the identical problem with the same `tool_names_by_call_id` map, the same falsy-id guard and the same `name or lookup` precedence. Those are the only two providers that need it: anthropic, bedrock and mistral pass the id straight through, and cohere deliberately strips `name` and matches positionally. Not worth a shared helper at two callers.

One thing that is not fixed here, and I do not think it should block: name-based matching still cannot disambiguate parallel calls to the *same* tool. Two `get_weather` calls produce two responses identical apart from the payload, so order still decides. Verified locally. `types.FunctionResponse` has an `id` field described as "Populated by the client to match the corresponding function call `id`", and `FunctionCall` has a matching one, so id-based matching would close it, but that means dropping `Part.from_function_response` and touching the `thought_signature` code, which deserves its own PR and its own live check. This PR turns a failure on every response into a correct result for the common case.

Two notes for follow-up:

- `tests/integration/test_agent_loop.py:80` and `:165` hand-add `"name": tool_call.function.name`, which is not part of the OpenAI tool-message shape and is why the suite never caught this. Worth dropping so the loops exercise the real shape.
- Reviewing this turned up the same class of bug on xai, filed as #1323.

The red `run-linter` is an unrelated httpx2 typing error in a file this PR does not touch, fixed separately in #1320.

Not run: gemini integration tests, no keys on my side.

PR #1326, opened 2026-08-19
fix(xai): preserve tool call IDs in request history

### LINKED ISSUE #1323: [BUG] xai drops tool_call_id on tool results, so parallel tool calls are matched by position
**`any-llm` version:** main (1.26.0)
**Python Version**: 3.14

### Describe the bug

`XaiProvider._acompletion` drops `tool_call_id` when it converts a tool result, so xAI cannot tell which tool call a result answers. With parallel tool calls the association is positional only, and results returned in any other order get attributed to the wrong call.

`src/any_llm/providers/xai/xai.py:198`:

```python
elif message["role"] == "tool":
    xai_messages.append(tool_result(message["content"]))
```

The xAI SDK helper accepts the id and its own docstring says it matters (`xai-sdk` 1.17.0):

```
tool_result(result: str, tool_call_id: str | None = None)
```

> the assistant's tool_calls list). Essential for parallel_tool_calls or multiple tools to associate results correctly. If omitted (for single tool calls), the association may still work but is less explicit.

The same conversion loop has a second, larger gap two branches up, at `xai.py:191`:

```python
if message.get("tool_calls"):
    # No idea how to pass tool calls reconstructed in the original protobuf format.
    args.extend(str(tool_call) for tool_call in message["tool_calls"])
xai_messages.append(assistant(*args, message["content"]))
```

The assistant's tool calls are stringified into text args, so what xAI sees is a Python repr rather than tool calls. `chat.assistant(*args: str | Content)` genuinely cannot express them, but the underlying proto can: a `chat_pb2.Message` carries `tool_calls`, `tool_call_id` and `name` fields, so building the message directly instead of going through the helper would let both halves round-trip.

Passing `tool_call_id` on the result is only fully useful once the assistant side emits real tool calls with matching ids, so these are probably one fix rather than two.

### Example w/ Code

Two parallel calls with results returned in reverse order. Every `tool_result` xAI receives is content-only, so nothing links the second result back to the first call:

```python
messages = [
    {"role": "user", "content": "weather in Toronto and Paris?"},
    {"role": "assistant", "content": None, "tool_calls": [
        {"id": "c1", "type": "function", "function": {"name": "get_weather", "arguments": '{"city":"Toronto"}'}},
        {"id": "c2", "type": "function", "function": {"name": "get_weather", "arguments": '{"city":"Paris"}'}},
    ]},
    {"role": "tool", "tool_call_id": "c2", "content": '{"temp": 30}'},
    {"role": "tool", "tool_call_id": "c1", "content": '{"temp": -5}'},
]
await completion(model="xai/grok-4", messages=messages, tools=[...])
```

### Context

Found while reviewing #1318, which fixes the same class of bug for gemini (tool results reaching the provider with no link to the call they answer). gemini and ollama resolve it by mapping `tool_call_id` to the function name; xAI takes the id directly, so it needs its own fix rather than a shared helper. I have not run this against the live xAI API, so the user-visible symptom is inferred from the conversion code and the SDK contract, not observed.

original4422:
## Description

`XaiProvider` previously stringified assistant tool calls into text and omitted `tool_call_id` from tool-result messages. That made parallel results depend on position instead of the IDs returned by the model. A tool-calling assistant message with `content=None` also failed locally before a request could be sent.

This change reconstructs each assistant call as a real xAI protobuf `ToolCall`, preserving its ID, client-side tool type, function name, and JSON arguments. Tool results now carry the matching `tool_call_id`. Assistant content remains absent for `None` and is preserved for empty strings and ordinary text; assistant messages without tool calls retain their existing behavior.

The regression test sends two results in reverse call order and verifies that the protobuf messages still associate them by ID.

## PR Type

- 🐛 Bug Fix

## Relevant issues

Fixes #1323

## Validation

- `uv run pytest -q tests/unit/providers/test_xai_provider.py --reruns 0`: 18 passed
- Full unit suite with local proxy variables removed for the test process: 2258 passed, 69 skipped
- `UV_SYSTEM_PYTHON=0 uv run pre-commit run --all-files --verbose`: passed, including Ruff, Ruff format, mypy, codespell, and nbstripout

The conversion regression and fix are credential-free and were verified against the real protobuf types from `xai-sdk==1.17.0`. I do not have `XAI_API_KEY`, so I did not run or claim a live xAI request. Maintainers, please add the `run-integration-tests` label for the repository's xAI integration coverage.

## Checklist

- [x] Documentation was updated where necessary (not applicable; this restores correct provider request conversion without changing the public API)
- [x] **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

- AI Model used: GPT-5
- AI Developer Tool used: Codex desktop
- Any other info you'd like to share: The agent independently rechecked issue ownership and competing PRs, reproduced the defect offline, worked test-first, and ran the repository's full unit and pre-commit checks.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved xAI conversation handling for assistant messages, including preserved content and messages without tool calls.
  * Preserved tool-call identifiers in standard and streaming responses, with accurate tool-result mapping when results arrive in a different order.
  * Improved support for multiple tool calls, structured response formats, and serialised tool-call arguments.

* **Chores**
  * Updated the optional xAI integration to use a newer SDK version.

* **Tests**
  * Added coverage for tool-call reconstruction, content preservation, streaming identifiers, and tool-result mapping.

### DIFF
--- a/src/any_llm/providers/xai/xai.py
+++ b/src/any_llm/providers/xai/xai.py
@@ -17,6 +17,7 @@ try:
     from xai_sdk.chat import Chunk as XaiChunk
     from xai_sdk.chat import Response as XaiResponse
     from xai_sdk.chat import assistant, required_tool, system, tool_result, user
+    from xai_sdk.proto import chat_pb2
 
     from .utils import (
         _convert_models_list,
@@ -187,15 +188,27 @@ class XaiProvider(AnyLLM):
             if message["role"] == "user":
                 xai_messages.append(user(message["content"]))
             elif message["role"] == "assistant":
-                args: list[str] = []
-                if message.get("tool_calls"):
-                    # No idea how to pass tool calls reconstructed in the original protobuf format.
-                    args.extend(str(tool_call) for tool_call in message["tool_calls"])
-                xai_messages.append(assistant(*args, message["content"]))
+                content = message["content"]
+                xai_message = (
+                    assistant(content) if content is not None else chat_pb2.Message(role=chat_pb2.ROLE_ASSISTANT)
+                )
+                for tool_call in message.get("tool_calls") or []:
+                    function = tool_call["function"]
+                    xai_message.tool_calls.append(
+                        chat_pb2.ToolCall(
+                            id=tool_call["id"],
+                            type=chat_pb2.TOOL_CALL_TYPE_CLIENT_SIDE_TOOL,
+                            function=chat_pb2.FunctionCall(
+                                name=function["name"],
+                                arguments=function["arguments"],
+                            ),
+                        )
+                    )
+                xai_messages.append(xai_message)
             elif message["role"] == "system":
                 xai_messages.append(system(message["content"]))
             elif message["role"] == "tool":
-                xai_messages.append(tool_result(message["content"]))
+                xai_messages.append(tool_result(message["content"], tool_call_id=message.get("tool_call_id")))
         if params.tools is not None:
             kwargs["tools"] = _convert_openai_tools_to_xai_tools(params.tools)
 
@@ -214,15 +227,11 @@ class XaiProvider(AnyLLM):
         # since xAI's parse() only supports Pydantic BaseModel.
         response_format_pb = None
         if params.response_format is not None and dataclasses.is_dataclass(params.response_format):
-            from xai_sdk.proto import chat_pb2
-
             response_format_pb = chat_pb2.ResponseFormat(
                 format_type=chat_pb2.FORMAT_TYPE_JSON_SCHEMA,
                 schema=json.dumps(get_json_schema(params.response_format)),
             )
         elif params.response_format is not None and isinstance(params.response_format, dict):
-            from xai_sdk.proto import chat_pb2
-
             rf = params.response_format
             if rf.get("type") == "json_schema":
                 json_schema = rf.get("json_schema", {})
--- a/src/any_llm/providers/xai/xai.py
+++ b/src/any_llm/providers/xai/xai.py
@@ -188,19 +188,22 @@ class XaiProvider(AnyLLM):
             if message["role"] == "user":
                 xai_messages.append(user(message["content"]))
             elif message["role"] == "assistant":
-                content = message["content"]
+                content = message.get("content")
                 xai_message = (
                     assistant(content) if content is not None else chat_pb2.Message(role=chat_pb2.ROLE_ASSISTANT)
                 )
                 for tool_call in message.get("tool_calls") or []:
                     function = tool_call["function"]
+                    arguments = function.get("arguments")
+                    if not isinstance(arguments, str):
+                        arguments = json.dumps(arguments or {})
                     xai_message.tool_calls.append(
                         chat_pb2.ToolCall(
                             id=tool_call["id"],
                             type=chat_pb2.TOOL_CALL_TYPE_CLIENT_SIDE_TOOL,
                             function=chat_pb2.FunctionCall(
                                 name=function["name"],
-                                arguments=function["arguments"],
+                                arguments=arguments,
                             ),
                         )
                     )
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -97,7 +97,7 @@ voyage = [
 ]
 
 xai = [
-  "xai-sdk>=1.0.1",
+  "xai-sdk>=1.6.0",
 ]
 
 sagemaker = [
--- a/src/any_llm/providers/xai/utils.py
+++ b/src/any_llm/providers/xai/utils.py
@@ -1,4 +1,3 @@
-import uuid
 from collections.abc import Sequence
 from typing import Any, Literal
 
@@ -76,7 +75,7 @@ def _convert_xai_chunk_to_anyllm_chunk(chunk: XaiChunk) -> ChatCompletionChunk:
                 delta_tool_calls_list.append(
                     ChoiceDeltaToolCall(
                         index=idx,
-                        id=str(uuid.uuid4()),
+                        id=tc.id,
                         type="function",
                         function=ChoiceDeltaToolCallFunction(name=func.name, arguments=func.arguments),
                     )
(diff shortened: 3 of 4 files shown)


JamMaster1999:
Ran this branch against live xAI, since the author noted they had no key. `grok-3-mini-latest` through `acompletion`: two parallel `get_weather` calls, results fed back in reverse call order (Paris 30C first, then Toronto -5C), then asked which city is warmer.

| | turn 2 answer |
|---|---|
| `upstream/main` | "Toronto is currently 30°C, while Paris is -5°C. Toronto is warmer." (results matched by position) |
| this branch | "Paris is warmer (30C vs -5C in Toronto)." (matched by id) |

Also ran the branch locally: 18/18 in `tests/unit/providers/test_xai_provider.py`, repo-wide `mypy src/any_llm` clean, and checked the SDK shapes against `xai-sdk` 1.17.0 (`tool_result(result, tool_call_id: Optional[str])`; an id-less tool message degrades to the empty-string default). We hit the same bug class on gemini/anthropic in #1325; this closes the xai case from #1323 properly.

original4422:
Hi @tbille, when you have a chance, could you please review this PR and let me know if any changes are needed? Thanks!

src/any_llm/providers/xai/xai.py
```diff
@@ -187,15 +188,27 @@ async def _acompletion(
             if message["role"] == "user":
                 xai_messages.append(user(message["content"]))
             elif message["role"] == "assistant":
-                args: list[str] = []
-                if message.get("tool_calls"):
-                    # No idea how to pass tool calls reconstructed in the original protobuf format.
-                    args.extend(str(tool_call) for tool_call in message["tool_calls"])
-                xai_messages.append(assistant(*args, message["content"]))
+                content = message["content"]
```
dpoulopoulos: **[major]** `message["content"]` raises `KeyError` when the assistant message has no `content` key. This is the normal path: `AnyLLM.acompletion` dumps a `ChatCompletionMessage` with `exclude_none=True`, so a tool-call turn with `content=None` arrives here without the key. A hand-written `{"role": "assistant", "tool_calls": [...]}` dict fails the same way. Use `message.get("content")` and add a test for the missing key.

src/any_llm/providers/xai/xai.py
```diff
@@ -187,15 +188,27 @@ async def _acompletion(
...
+                )
+                for tool_call in message.get("tool_calls") or []:
+                    function = tool_call["function"]
+                    xai_message.tool_calls.append(
+                        chat_pb2.ToolCall(
+                            id=tool_call["id"],
+                            type=chat_pb2.TOOL_CALL_TYPE_CLIENT_SIDE_TOOL,
+                            function=chat_pb2.FunctionCall(
+                                name=function["name"],
+                                arguments=function["arguments"],
```
dpoulopoulos: **[nit]** `chat_pb2.FunctionCall(arguments=...)` needs a string. If a caller passes already parsed arguments (a dict), this raises `TypeError`. That is the same input case #1324 accepted for gemini, so xAI is now stricter than gemini.

dpoulopoulos (commented):
Thanks @original4422! Good, well-targeted fix: the protobuf tool calls and the `tool_call_id` on results are the right way to close #1323, and the unit tests use the real `chat_pb2` types instead of mocks.

Two things to handle before merge: an assistant message with no content key still raises `KeyError` at `xai.py:191`, because `AnyLLM.acompletion` dumps `ChatCompletionMessage` with `exclude_none=True;` and the code needs `xai-sdk 1.6.0` or newer, while the extra still declares `>=1.0.1`.

### General notes
- **[minor]** `pyproject.toml:100` — The extra declares `xai-sdk>=1.0.1`, but the new code needs a newer SDK. `tool_result(..., tool_call_id=...)` exists only from 1.6.0, and `ToolCall.type` / `TOOL_CALL_TYPE_CLIENT_SIDE_TOOL` only from 1.4.0. On 1.0.1 to 1.5.0 every tool message now fails with `TypeError` and every assistant tool call with `ValueError`. Raise the floor to `xai-sdk>=1.6.0`.
- **[minor]** `src/any_llm/providers/xai/utils.py:59` — The streaming path still replaces xAI's tool call ID with `str(uuid.uuid4())`, so a streamed tool call does not carry the ID the model returned. The ID fix therefore only covers non-streaming calls. This line is outside the diff, so a follow-up is fine, but please confirm it is intended.

src/any_llm/providers/xai/xai.py
original4422: Addressed in bc52e0a8: assistant content is now read with get(), so an omitted key follows the same empty-content protobuf path as an explicit None. I added the missing-key regression test. Validation: all 21 xAI provider tests passed; the full unit suite passed with 2276 passed and 69 skipped; all pre-commit hooks passed.

src/any_llm/providers/xai/xai.py
original4422: Addressed in bc52e0a8: non-string tool arguments are now JSON-serialized before constructing FunctionCall, while existing JSON strings remain unchanged. The regression covers an already-parsed dict. Validation: all 21 xAI provider tests passed; the full unit suite passed with 2276 passed and 69 skipped; all pre-commit hooks passed.

original4422:
Thanks for the review. I addressed all requested items in bc52e0a8: raised the xAI SDK floor to >=1.6.0, accepted assistant messages without a content key, serialized already-parsed tool arguments, and changed streaming conversion to preserve the tool-call ID supplied by xAI instead of generating a new UUID. Added regressions for all three behavior fixes. Validation: 21 xAI provider tests passed; the complete unit suite passed with 2276 passed and 69 skipped; all pre-commit hooks passed (including Ruff, Ruff format, mypy, codespell, and nbstripout). XAI_API_KEY is not available locally, so I did not claim a new live-provider run; the repository integration workflow remains authoritative.

original4422:
Thanks @javiermtorres for syncing this branch with the latest main. The refreshed head is mergeable and all 15 executed checks are passing (with one expected skip). When convenient, could you please confirm whether this is ready to merge, or let me know if any further changes are needed? Thanks!

javiermtorres (approved):

PR #1327, opened 2026-08-20
test(integration): send tool results in the real OpenAI shape

JamMaster1999:
## Description

The agent-loop integration tests hand-add a `name` field to every tool message they build. OpenAI tool messages have no such key (`ChatCompletionToolMessageParam` is `role`/`content`/`tool_call_id`; `name` belongs to the deprecated `role="function"` shape), so the suite exercised a shape no real caller sends, and the hand-added name short-circuited the resolution path #1318 fixed. Deleting the two lines makes the loops run on the real shape. To be precise about what that buys: it removes a false-safe path rather than turning these tests into a regression guard for #1318 itself, since the bug was a wrong answer, not an error, and these tests do not assert answer content. The `name`-first and fallback branches stay pinned by unit tests (gemini, ollama, cohere).

The key was never load-bearing anywhere: no converter requires it, cohere's `_patch_messages` exists to strip it, and the OpenAI-compatible providers were passing it through to their APIs verbatim. gemini resolves the name from `tool_call_id` since #1318 (ollama has the same map, but sits in the local-provider set this file skips). Live run of `tests/integration/test_agent_loop.py` on this branch with the keys available here: 12 passed, 70 skipped; the six failures (azureopenai, fireworks, sagemaker on both loop tests) fail identically on `upstream/main` with the same environment, for billing/endpoint reasons unrelated to this change.

## PR Type

- 🧪 Tests

## Relevant issues

Follow-up to #1318 (second bullet of the review's follow-up notes).

## Checklist

- [ ] I have added unit tests that prove my fix/feature works (test-only change; the edited tests are the deliverable)
- [x] **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

## Summary by CodeRabbit

* **Tests**
  * Improved integration coverage for agent conversations involving tool results.
  * Added checks that final responses accurately reflect returned information.
  * Strengthened validation that conversations complete successfully after tool interactions.
  * Updated test coverage for parallel and sequential tool calls, including streamlined tool-result messages.


JamMaster1999:
@njbrake this is the follow-up you flagged in your #1318 review (the integration loops hand-adding `name` to tool messages, which is why the suite never caught that bug). The other item from that review, xai dropping `tool_call_id` (#1323), is already handled by #1326 — we verified that branch against live xAI with reversed parallel results and posted the before/after there.

tests/integration/test_agent_loop.py
```diff
@@ -162,7 +161,6 @@ async def test_agent_loop_sequential_tool_calls(
                         "role": "tool",
                         "content": tool_result,
                         "tool_call_id": tool_call.id,
```
dpoulopoulos, line 209: **[nit]** Callers may still send `name` on a tool message, and any-llm accepts it. After this change no integration test sends that shape, so a provider API that rejects the extra key would no longer be caught live. Unit tests cover the conversion, not the wire call, so it may be worth keeping one loop on the `name` shape.

dpoulopoulos (dismissed):
Good, well-scoped test fix. The `name` key is not part of `ChatCompletionToolMessageParam`, so the loops now send the shape a real caller sends, and no converter needs the key: gemini and ollama fall back to the `tool_call_id` map, cohere strips it, and the other providers only passed it through. The `name`-first branch is still pinned by the gemini, ollama and cohere unit tests, so no coverage is lost there. Two test weaknesses remain in this file (the final `iteration <= max_iterations` assert can never fail, and neither loop asserts the answer content), so a wrong tool-result label still passes; please consider fixing them here. Note that Unit Tests and Lint have not run on this PR yet, so a maintainer still needs to approve the fork workflow runs. No browser check: this is a Python library with no web UI.

### General notes
- **[minor]** `tests/integration/test_agent_loop.py:167` — This assert can never fail. `iteration` is incremented only inside `while iteration < max_iterations`, so it is always `<= max_iterations`. If the model never returns a final text answer, the loop just runs out and the test still passes. Set a flag when the `tool_calls is None` branch breaks, then assert that flag, so a loop that never finishes fails.
- **[minor]** `tests/integration/test_agent_loop.py:89` — Neither loop asserts anything about the answer, so a tool result that reaches the provider under the wrong name still passes. That is the failure mode #1318 fixed. Consider asserting that the second response has text content and repeats a value from the tool result (for example `15` or both city names), so the loop becomes a real guard.

tests/integration/test_agent_loop.py
JamMaster1999: Kept `name` on the sequential loop, so that shape still goes over the wire once per provider. The parallel loop stays on the real shape, since that is the one where the `tool_call_id` mapping matters. 3bcace4.

JamMaster1999:
@dpoulopoulos both general notes addressed in 3bcace4: the sequential loop sets `answered` when the model returns a final text turn and asserts that flag, so running out of iterations now fails, and both loops assert the answer repeats a value from the tool result (`15` or `sunny`) through one `_mentions_tool_result` helper. The `name` nit is handled on the inline thread. These are integration tests and I have not run them live from this machine, so the `run-integration-tests` label is the check that counts here.

JamMaster1999:
Ran both loops live on 901fbe2 against openai, gemini, anthropic and mistral: 8/8 pass, so the new answer asserts hold on real responses. 901fbe2 also adds a direct test for `_mentions_tool_result` per the CodeRabbit note.

tests/integration/test_agent_loop.py
```diff
@@ -32,6 +32,25 @@ def get_weather(location: str) -> str:
...
+    [
+        (None, False),
+        ("", False),
+        ("It rains in Paris.", False),
+        ("It is 15C in Paris.", True),
+        ("Sunny in London.", True),
+    ],
+)
+def test_mentions_tool_result(content: str | None, expected: bool) -> None:
+    assert _mentions_tool_result(content) is expected
```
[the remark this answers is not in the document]
JamMaster1999: `FBT` is in the `ignore` list under `[tool.ruff.lint]` in `pyproject.toml`, so ruff does not report FBT001 here (same as the withdrawn finding on #1319). Keeping the positional parameter to match the other parametrized tests in the file. No change.

JamMaster1999:
@njbrake @dpoulopoulos rebased onto main (1.27.0) after #1340; the only conflict was `_call_tool` landing next to `_mentions_tool_result`, and both stay. CodeRabbit's `150F` point is fixed in 80dd2cd: the matcher now requires `15` not followed by a digit. Ran both loops live on openai, gemini, anthropic and mistral at 80dd2cd: 8/8. This is the follow-up from the #1318 review, so a look would be much appreciated. CI still needs a workflow approval.

JamMaster1999:
@HareeshBahuleyan this one was approved by @dpoulopoulos on Aug 25. After that I addressed his two notes (5644fcd), added a direct test for the predicate (f55d35f), and rebased onto 1.27.0 after #1340 landed, which is what dismissed the approval. Tests only, `tests/integration/test_agent_loop.py` is the sole file. Ran both loops live against openai, gemini, anthropic and mistral on Aug 29, 8/8 pass.

HareeshBahuleyan (approved):
Thanks @JamMaster1999 for the contribution!
Reviewed the latest head. The changes match the intended tool-message shape, and the earlier feedback has been addressed. Pre-commit and unit tests pass on the merge result with current `main` (2536 passed, 69 skipped). No additional code changes requested.

PR #1333, opened 2026-08-22
fix(gemini): preserve inline data response parts

### LINKED ISSUE #1295: [BUG] Gemini converter drops inline_data parts

mikemikimike:
## Description

Preserve Gemini `inline_data` response parts as OpenAI-compatible image items for both non-streaming and streaming responses, including media-only and thought-marked parts. The conversion now ignores incomplete mock/provider parts instead of attempting to base64-encode non-byte values.

## PR Type

- 🐛 Bug Fix

## Relevant issues

Fixes #1295

## Root cause and scope

Gemini response conversion only handled text, thought, and function-call parts. `inline_data` was silently discarded, and media-only candidates produced no choice. The fix adds the `images` response fields, converts inline bytes to data URLs in both paths, and preserves existing text/tool/finish-reason behavior. The type guard keeps text-only fixtures and incomplete provider parts out of the media conversion path.

## Validation

- Added unit coverage for mixed text + media, media-only, streaming media, and thought-marked media.
- `uv run --extra gemini pytest tests/unit/providers/test_gemini_provider.py -q --reruns 0 --timeout=20` — **182 passed**.
- `uv run --extra gemini pytest tests/unit/providers/test_gemini_provider.py -q --reruns 0 --timeout=20 -k "inline_data or image_only"` — **5 passed, 177 deselected**.
- Ruff lint and format checks on changed files — passed.
- `python -m compileall` on changed Python files — passed.
- `git diff --check` — passed.

## Checklist

- [x] **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

- AI Model used: GPT-5
- Any other info: implementation and validation were reviewed against the current source and reviewer feedback.

## Summary by CodeRabbit

- **New Features**
  - Gemini responses now support inline images alongside text, reasoning, tool calls and thought signatures.
  - Images are provided as OpenAI-compatible base64 data URLs in regular and streaming responses.
  - Image information is available in chat messages and streaming response updates.
  - Image-only responses are supported.

- **Bug Fixes**
  - Incomplete inline image data is safely ignored during response conversion.
  - Inline images now remain available when returned alongside other response content.

### DIFF
--- a/tests/unit/providers/test_anthropic_messages.py
+++ b/tests/unit/providers/test_anthropic_messages.py
@@ -7,12 +7,12 @@ from dataclasses import dataclass
 from typing import Any, Self, cast
 from unittest.mock import AsyncMock, MagicMock, Mock, patch
 
-import httpx as standard_httpx
+import httpx
 
 try:
     import httpx2
 except ImportError:
-    httpx = standard_httpx
+    pass
 else:
     httpx = httpx2
 
--- a/tests/unit/providers/test_anthropic_messages.py
+++ b/tests/unit/providers/test_anthropic_messages.py
@@ -7,7 +7,11 @@ from dataclasses import dataclass
 from typing import Any, Self, cast
 from unittest.mock import AsyncMock, MagicMock, Mock, patch
 
-import httpx
+try:
+    import httpx2 as httpx
+except ImportError:
+    import httpx
+
 import pytest
 from anthropic.types import Message, TextBlock, ThinkingBlock, ToolUseBlock, Usage
 from anthropic.types.beta import BetaMCPToolUseBlock, BetaMessage, BetaThinkingBlock, BetaUsage
--- a/tests/unit/providers/test_anthropic_messages.py
+++ b/tests/unit/providers/test_anthropic_messages.py
@@ -8,14 +8,6 @@ from typing import Any, Self, cast
 from unittest.mock import AsyncMock, MagicMock, Mock, patch
 
 import httpx
-
-try:
-    import httpx2
-except ImportError:
-    pass
-else:
-    httpx = httpx2
-
 import pytest
 from anthropic.types import Message, TextBlock, ThinkingBlock, ToolUseBlock, Usage
 from anthropic.types.beta import BetaMCPToolUseBlock, BetaMessage, BetaThinkingBlock, BetaUsage
--- a/tests/unit/providers/test_anthropic_messages.py
+++ b/tests/unit/providers/test_anthropic_messages.py
@@ -7,10 +7,14 @@ from dataclasses import dataclass
 from typing import Any, Self, cast
 from unittest.mock import AsyncMock, MagicMock, Mock, patch
 
+import httpx as standard_httpx
+
 try:
-    import httpx2 as httpx
+    import httpx2
 except ImportError:
-    import httpx
+    httpx = standard_httpx
+else:
+    httpx = httpx2
 
 import pytest
 from anthropic.types import Message, TextBlock, ThinkingBlock, ToolUseBlock, Usage
--- a/tests/unit/providers/test_gemini_provider.py
+++ b/tests/unit/providers/test_gemini_provider.py
@@ -975,6 +975,31 @@ def test_convert_streaming_response_preserves_inline_data() -> None:
     assert chunk.model_dump()["choices"][0]["delta"]["images"] == chunk.choices[0].delta.images
 
 
+def test_convert_response_preserves_thought_inline_data() -> None:
+    response = _make_gemini_response(
+        [types.Part(inline_data=types.Blob(mime_type="image/png", data=b"\x89PNG"), thought=True)],
+        types.FinishReason.STOP,
+    )
+
+    message = ChatCompletion.model_validate(_convert_response_to_response_dict(response)).choices[0].message
+
+    assert message.reasoning is None
+    assert message.images == [{"type": "image_url", "image_url": {"url": "data:image/png;base64,iVBORw=="}}]
+
+
+def test_convert_streaming_response_preserves_thought_inline_data() -> None:
+    response = _make_gemini_response(
+        [types.Part(inline_data=types.Blob(mime_type="image/png", data=b"\x89PNG"), thought=True)],
+        types.FinishReason.STOP,
+    )
+
+    chunk = _create_openai_chunk_from_google_chunk(response)
+
+    assert chunk.choices[0].delta.images == [
+        {"type": "image_url", "image_url": {"url": "data:image/png;base64,iVBORw=="}}
+    ]
+
+
 def test_convert_response_emits_choice_for_filtered_response_without_content() -> None:
     response_dict = _convert_response_to_response_dict(_make_gemini_response(None, types.FinishReason.SAFETY))
 
--- a/tests/unit/providers/test_gemini_provider.py
+++ b/tests/unit/providers/test_gemini_provider.py
@@ -39,6 +39,13 @@ from any_llm.types.completion import (
 
 TEST_IMAGE_BYTES = b"test-image-bytes"
 TEST_PDF_BYTES = b"%PDF-1.4\ntest"
+TEST_PNG_BYTES = b"\x89PNG\r\n\x1a\n"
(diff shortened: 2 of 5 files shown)


mikemikimike:
This is addressed in the current head `884cace0a3ffb27b10d6cdbf7691787347ad340b`.

`GoogleProvider._convert_completion_response()` now propagates `images` into the public `ChatCompletionMessage`, and `tests/unit/providers/test_gemini_provider.py` includes `test_convert_completion_response_preserves_public_images`, asserting both the public attribute and serialized output.

tests/unit/providers/test_gemini_provider.py
```diff
@@ -929,9 +930,7 @@ def test_convert_response_accumulates_multiple_text_parts() -> None:
     assert message["reasoning"] is None
 
 
-def test_convert_response_skips_parts_without_text_or_function_call() -> None:
-    """A candidate can mix in parts that carry neither text nor a tool call, e.g. inline image
-    data; those must not disturb the accumulated text."""
+def test_convert_response_preserves_inline_data_alongside_text() -> None:
     response = _make_gemini_response(
         [
             types.Part(inline_data=types.Blob(mime_type="image/png", data=b"\x89PNG")),
```
javiermtorres: Isn't the PNG header 8 bytes? This is definitely nonblocking, but just in case it might cause an issue with something checking content types.

tests/unit/providers/test_gemini_provider.py
```diff
@@ -945,6 +944,88 @@ def test_convert_response_skips_parts_without_text_or_function_call() -> None:
     message = response_dict["choices"][0]["message"]
     assert message["content"] == "Described."
     assert message["tool_calls"] is None
+    assert message["images"] == [{"type": "image_url", "image_url": {"url": "data:image/png;base64,iVBORw=="}}]
```
javiermtorres: Instead of hardcoding the base64, please derive it form the base64 above.

tests/unit/providers/test_gemini_provider.py
```diff
@@ -945,6 +944,88 @@ def test_convert_response_skips_parts_without_text_or_function_call() -> None:
...
+    response_dict = _convert_response_to_response_dict(
+        _make_gemini_response(
+            [types.Part(inline_data=types.Blob(mime_type="image/png", data=b"\x89PNG"))],
+            types.FinishReason.STOP,
+        )
+    )
+
+    completion = GoogleProvider._convert_completion_response((response_dict, "test-model"))
+    message = completion.choices[0].message
+    assert message.images == [{"type": "image_url", "image_url": {"url": "data:image/png;base64,iVBORw=="}}]
```
javiermtorres: Same here

tests/unit/providers/test_gemini_provider.py
```diff
@@ -945,6 +944,88 @@ def test_convert_response_skips_parts_without_text_or_function_call() -> None:
...
+
+    response_dict = _convert_response_to_response_dict(response)
+
+    assert response_dict["choices"][0]["finish_reason"] == "stop"
+    assert response_dict["choices"][0]["message"]["images"]
+
+
+def test_convert_streaming_response_preserves_inline_data() -> None:
+    response = _make_gemini_response(
+        [types.Part(inline_data=types.Blob(mime_type="audio/wav", data=b"RIFF"))],
```
javiermtorres: Again nonblocking. Strictly, the wav header is larger.

tests/unit/providers/test_gemini_provider.py
```diff
@@ -945,6 +944,88 @@ def test_convert_response_skips_parts_without_text_or_function_call() -> None:
...
+        types.FinishReason.STOP,
+    )
+
+    chunk = _create_openai_chunk_from_google_chunk(response)
+
+    assert chunk.choices[0].delta.images == [
+        {"type": "image_url", "image_url": {"url": "data:audio/wav;base64,UklGRg=="}}
```
javiermtorres: Derive the base64 value from the unencoded value.

tests/unit/providers/test_gemini_provider.py
```diff
@@ -945,6 +944,88 @@ def test_convert_response_skips_parts_without_text_or_function_call() -> None:
...
+        [types.Part(inline_data=types.Blob(mime_type="image/png", data=b"\x89PNG"), thought=True)],
+        types.FinishReason.STOP,
+    )
+
+    message = ChatCompletionMessage.model_validate(
+        _convert_response_to_response_dict(response)["choices"][0]["message"]
+    )
+
+    assert message.reasoning is None
+    assert message.images == [{"type": "image_url", "image_url": {"url": "data:image/png;base64,iVBORw=="}}]
```
javiermtorres: same here

tests/unit/providers/test_gemini_provider.py
```diff
@@ -945,6 +944,88 @@ def test_convert_response_skips_parts_without_text_or_function_call() -> None:
...
+def test_convert_streaming_response_preserves_thought_inline_data() -> None:
+    response = _make_gemini_response(
+        [types.Part(inline_data=types.Blob(mime_type="image/png", data=b"\x89PNG"), thought=True)],
+        types.FinishReason.STOP,
+    )
+
+    chunk = _create_openai_chunk_from_google_chunk(response)
+
+    assert chunk.choices[0].delta.images == [
+        {"type": "image_url", "image_url": {"url": "data:image/png;base64,iVBORw=="}}
```
javiermtorres: same here

javiermtorres:
@mikemikimike could you please check the failing unit tests? Thanks in advance.

mikemikimike:
I traced the CI failures to 	ests/unit/providers/test_anthropic_messages.py: Anthropic 1.0.0 now requires an httpx2.AsyncClient, while the tests were constructing httpx.AsyncClient instances. I pushed commit 302ff5c, which imports httpx2 with a fallback to httpx for older Anthropic SDK versions. The focused regression suite now passes: 55 passed.

mikemikimike:
The new workflow runs for commit 302ff5c are currently blocked with ction_required before any jobs start because this fork PR requires upstream approval. I do not have admin permission to approve them (403 Must have admin rights to Repository). Once approved, the Unit Tests matrix should run against the updated test import.

tests/unit/providers/test_anthropic_messages.py
```diff
@@ -7,7 +7,11 @@
 from typing import Any, Self, cast
 from unittest.mock import AsyncMock, MagicMock, Mock, patch
 
-import httpx
+try:
+    import httpx2 as httpx
+except ImportError:
+    import httpx
+
```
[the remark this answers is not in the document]
mikemikimike: I verified this against the repository's supported dependency range before changing the import. `pyproject.toml` allows `anthropic>=0.119.0`, and CI re-resolves dependencies rather than using a committed lockfile.

- With `anthropic==0.119.0` and the standard `httpx` import, the focused module passes: 55 passed.
- With the current resolver's `anthropic==1.0.0`, replacing the conditional import with standard `httpx` causes 15 tests to fail because that SDK validates for `httpx2.AsyncClient`.
- Keeping the conditional import makes the same module pass under `anthropic==1.0.0`: 55 passed.

The compatibility import is therefore required to support both the declared minimum and the current latest SDK. I left it in place; pinning `anthropic<1` would be a separate dependency-policy change if standard `httpx` is required exclusively.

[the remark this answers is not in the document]
mikemikimike: Follow-up: commit `42113bb` removes the duplicate-name pattern that caused the `mypy no-redef` failure while preserving the compatibility behavior. The focused module passes with `anthropic==1.0.0` and `httpx2` (55 passed); Ruff check, Ruff format, Python compilation, and `git diff --check` also pass locally. The conditional selection remains intentional because a direct standard `httpx` import fails against the currently resolved Anthropic SDK.

PR #1336, opened 2026-08-23
fix(mistral): trim reasoning at the opening response tag in the streaming path too

shoemoney:
## What

#1302 fixed a bug where some Mistral reasoning models (Magistral) sometimes return the real
answer wrapped in `<response>...</response>` inside the reasoning/thinking content instead of
in `content`. That fix only landed in `_create_mistral_completion_from_response` (the
non-streaming converter), `src/any_llm/providers/mistral/utils.py:209-218` on main before this
PR. `_create_openai_chunk_from_mistral_chunk` (the streaming converter, same file, ~256-334)
never got the equivalent change.

## The bug

With `stream=True`, when a `ThinkChunk` delta's thinking text contains the `<response>` markers,
`choice.delta.content` stays `None` for that chunk and the whole answer, tags included, ends up
only in `delta.reasoning.content`. A caller following the standard OpenAI streaming contract
(accumulate `delta.content` across chunks) gets an empty response. No exception, no error - it
just silently returns nothing. The identical request with `stream=False` returns the answer
correctly today (because of #1302), which is what made this easy to miss.

## The fix

Pulled the split/trim logic out of the non-streaming function into a shared
`_split_response_tag_from_reasoning()` helper and call it from both converters. That's the part
that actually matters here - the two converters had this same logic and one of them fell
behind once already; sharing it is what stops that from happening a second time.

## Streaming semantics limitation

Deltas arrive in fragments, so in principle `<response>` or `</response>` could be split across
two separate chunks. This fix only handles the case where a single chunk's reasoning text
contains both the opening and closing tag, mirroring what #1302 already does for a single
non-streaming response body. It does not buffer or reassemble text across chunks to catch a tag
boundary that lands mid-chunk-break. Noting this here rather than silently leaving it unhandled;
happy to follow up with a stateful/buffered version if that's wanted, but wanted to keep this
diff minimal and match #1302's scope.

## Testing

Added to `tests/unit/providers/test_mistral_provider.py`:
- `test_create_openai_chunk_strips_response_block_from_reasoning` - a streaming chunk whose
  thinking text contains `<response>...</response>` yields `delta.content` with the answer and
  `reasoning` trimmed to what came before the opening tag
- `test_create_openai_chunk_leaves_plain_reasoning_unchanged` - regression: a normal streaming
  chunk with plain reasoning and no tags is unaffected

Ran:
- `uv run pytest tests/unit/providers/test_mistral_provider.py -p no:rerunfailures` - 64 passed
  (62 previously existing + 2 new)
- Reverted the source change only (kept the new tests) and reran the two new tests to confirm
  they fail against the old code:
  `test_create_openai_chunk_strips_response_block_from_reasoning` failed with
  `AssertionError: assert None == 'The answer is 42.'` - `delta.content` stayed `None`, matching
  the reported bug exactly. Restored the fix afterward and reran the full file to confirm 64
  passed again.
- `uv run ruff check` / `uv run ruff format --check` on both changed files - clean
- `uv run mypy src/any_llm/providers/mistral/utils.py` - no issues

## PR Type

- 🐛 Bug Fix

## Relevant issues

Follow-up to #1302, which fixed the same bug in the non-streaming converter but missed the
streaming one.

## Checklist

- [x] **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

- Any other info you'd like to share: Written in conjunction with my pair programmer Claude. The finding was verified against upstream before any code was written, and the fix is covered by tests that fail against the unpatched source.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Mistral response handling when answers are embedded within reasoning content.
  * Recovered wrapped answers consistently in both standard and streaming responses.
  * Prevented response markers from appearing in displayed reasoning.
  * Preserved existing behaviour for ordinary reasoning without response markers.

### DIFF
--- a/src/any_llm/providers/mistral/utils.py
+++ b/src/any_llm/providers/mistral/utils.py
@@ -158,6 +158,34 @@ def _extract_mistral_content_and_reasoning(
     return content, reasoning_content, thinking_signature
 
 
+def _split_response_tag_from_reasoning(
+    content: str | None, reasoning_content: str | None
+) -> tuple[str | None, str | None]:
+    """Recover an answer that Mistral wrapped in `<response>` tags inside the thinking trace.
+
+    Some Mistral reasoning models (e.g. Magistral) sometimes return an empty/None content
+    and instead wrap the real answer in `<response>...</response>` inside the reasoning
+    content. When that happens, pull the answer out of the reasoning and trim the reasoning
+    down to what came before the opening tag.
+
+    Used by both the non-streaming and streaming converters so the two paths can't drift
+    apart again the way they did when this was only fixed in one of them (see #1302).
+
+    Note: this only handles the tags landing in a single string. In streaming, if the
+    `<response>` or `</response>` marker itself is split across two chunks, this will not
+    catch it - each chunk's reasoning text is checked independently.
+    """
+    if (
+        content is None
+        and reasoning_content
+        and "<response>" in reasoning_content
+        and "</response>" in reasoning_content
+    ):
+        content = reasoning_content.split("<response>")[1].split("</response>")[0]
+        reasoning_content = reasoning_content.split("<response>")[0]
+    return content, reasoning_content
+
+
 def _create_mistral_completion_from_response(
     response_data: MistralChatCompletionResponse, model: str
 ) -> ChatCompletion:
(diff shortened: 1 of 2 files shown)


shoemoney:
CI note: check-template failure is pre-existing on main/trunk and unrelated to this PR's changed files. Transient governance failure on run 32672248751 due to missing Checklist section in PR body; same head SHA later passed on runs 32672293502 and 32684340589 and current checks show pass, check never runs on main and is unrelated to mistral utils changes. No code fix required from this PR; rebase/label will clear it.

javiermtorres:
@shoemoney re:

> It does not buffer or reassemble text across chunks to catch a tag
boundary that lands mid-chunk-break. 

Did you observe this at some point? Maybe Mistral already handles this case?
If you can confirm seeing this we can open a specific issue on it.

PR #1344, opened 2026-08-27
fix(gemini): preserve inline media response parts

### LINKED ISSUE #1295: [BUG] Gemini converter drops inline_data parts

mikemikimike:
## Description

Gemini media responses currently discard `inline_data` parts during conversion to the OpenAI-compatible response shape. This makes image-only responses appear to have no choices and loses media from text-plus-image responses.

This change converts inline media to data URLs in both non-streaming and streaming response paths, exposes them as `images`, and emits a choice when a candidate contains media without text or tool calls.

## PR Type

- 🐛 Bug Fix

## Relevant issues

Fixes #1295

## Testing

- `python -m pytest tests/unit/providers/test_gemini_provider.py -q -k 'skips_parts or image_only'`
- `ruff check src/any_llm/providers/gemini/utils.py tests/unit/providers/test_gemini_provider.py`
- `git diff --check`

The focused tests pass locally. The full provider module includes existing integration-style cases that exceeded the local command timeout.

## Checklist

- [x] **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

- AI Model used: GPT-5
- Any other info you'd like to share: The patch and tests were reviewed locally before submission.

## Summary by CodeRabbit

- **New Features**
  - Gemini inline images are now returned as OpenAI-compatible `data:` URLs.
  - Image content is supported in both standard and streaming responses.
  - Image-only responses now produce a valid response choice.

- **Bug Fixes**
  - Inline image parts are no longer omitted when converting Gemini responses.


mikemikimike:
Thanks for pointing this out. Fixed in commit f57d60b.

`_inline_data_image` now rejects empty byte payloads in addition to missing data, and the existing regression test covers the empty-payload case alongside the other invalid inline-data cases.

Validation:
- `uv run pytest -p no:rerunfailures tests/unit/providers/test_gemini_provider.py -k "convert_response_skips_inline_data_without_image_payload or streaming_completion_with_inline_image"` — 5 passed
- `uv run mypy src/any_llm/providers/gemini/utils.py src/any_llm/types/completion.py` — passed
- `git diff --check` — passed

The repository's Ruff checks still report pre-existing copyright-header and formatting issues in the touched files; no unrelated cleanup was included.

JamMaster1999:
Thanks for picking up #1295. Two things from running the branch locally at 08cf1fb:

1. `images` never reaches the public non-streaming result. `GoogleProvider._convert_completion_response` in `gemini/base.py` builds `ChatCompletionMessage` with explicit kwargs and does not pass `images`, so `acompletion()` returns `message.images is None` even though `_convert_response_to_response_dict` fills it. Reproduced with the test helper: the intermediate dict carries the `image_url` block, `ChatCompletion.choices[0].message.images` is `None`. The tests here stop at the dict, which is why it passes. `images=message_dict.get("images")` there plus one test through `_convert_completion_response` closes it. Streaming is fine, since the chunk is built in `utils.py`.

2. Five existing tests fail on this branch and pass on main: `test_streaming_completion_includes_usage_data`, `test_streaming_completion_without_usage_metadata`, `test_convert_response_text_part_thought_signature_rides_the_message`, `test_streaming_text_part_thought_signature_rides_the_delta`, `test_streaming_completion_with_finish_reason_none`. They build parts with `Mock()`, so `part.inline_data` is a truthy `Mock` and `_inline_data_image` reaches `base64.b64encode(<Mock>)` (`utils.py:376`, `TypeError: a bytes-like object is required, not 'Mock'`). CI has not run on this PR yet (only CodeRabbit shows; Unit Tests need a workflow approval on fork PRs), so it is not visible from the checks. Either set `inline_data = None` on those mocks or check `isinstance(blob.data, bytes)` in the helper.

#1296 had the `base.py` line and used a real `types.Part` in its tests before it was closed as stale, if you want to borrow from it. Happy to re-verify on `gemini-2.5-flash-image` once these land.

mikemikimike:
Implemented in commit `e92a08e6`.

- `GoogleProvider._convert_completion_response` now forwards `message_dict["images"]` into `ChatCompletionMessage`, so non-streaming callers receive generated images.
- `_inline_data_image` now ignores non-bytes, empty, non-string MIME, and non-`image/*` payloads. This prevents `Mock()`-based parts from reaching `base64.b64encode` while preserving valid image conversion.
- Added a regression test for the public non-streaming completion path.

Validation:
- Focused Gemini regression tests: 11 passed.
- Full Gemini unit module: 211 passed; 2 pre-existing environment failures while creating the Google client (`ssl.SSLError: [SSL] unknown error` on Windows).
- Mypy passed for the changed provider/type files.
- `git diff --check` passed.

The Unit Tests and Lint workflows were created for the new head but are currently `action_required` with no jobs because fork workflow approval is required.

mikemikimike:
Addressed the latest typed-access review in commit `cb46697e`.

The non-streaming Gemini part conversion now uses the declared `types.Part` fields directly (`part.thought` and `part.function_call`) instead of `getattr`. The focused Gemini regression set still passes: 11 passed. Mypy and `git diff --check` also pass; the only remaining Ruff findings are the repository's pre-existing copyright-header checks.

javiermtorres:
@mikemikimike please address the remaining comments from copilot if appropriate.

mikemikimike:
Addressed the remaining Copilot inline-media feedback in commit 1f519fa. Added non-streaming and streaming regression coverage for audio-only STOP responses; existing mixed text+audio coverage remains. Focused inline image/audio tests (4) and pre-commit checks pass.

JamMaster1999 (changes requested):
I tested this E2E for both Gemini and OpenAI endpoints.

`gemini-2.5-flash-image` returns a valid PNG on `message.images` and `delta.images`, and the message replays as history. The images half looks ready.

The audio half regresses OpenAI. `gpt-audio-mini` streaming with `pcm16` works on main and fails on this head at the first chunk:

```
ValidationError: 2 validation errors for ChatCompletionChunk
choices.0.delta.audio.data        Field required
choices.0.delta.audio.expires_at  Field required
```

OpenAI sends `delta.audio` in pieces (`{"id", "transcript"}`, then `{"data"}`, then `{"expires_at"}`) and `ChoiceDelta.audio` here is the complete `ChatCompletionAudio`. On the Gemini side, a real `gemini-2.5-flash-preview-tts` call returns `audio/L16;codec=pcm;rate=24000`: headerless PCM whose rate lives only in the mime type, which `ChatCompletionAudio` cannot carry, so `message.audio.data` is not playable as shipped, and only the last audio part of a candidate survives. The audio tests use `audio/wav` with the bytes `WAVE`, so they never hit that path.

**Suggested implementation**, same pattern as `ChoiceDeltaToolCall`:

```python
class ChoiceDeltaAudio(BaseModel):
    """Streaming counterpart of ``ChatCompletionAudio``; every field optional, as in ``ChoiceDeltaToolCall``."""
    id: str | None = None
    data: str | None = None
    transcript: str | None = None
    expires_at: int | None = None

class ChoiceDelta(OpenAIChoiceDelta):
    ...
    audio: ChoiceDeltaAudio | None = None
```

`ChatCompletionMessage` keeps the base class's `audio`, so the redeclaration and the `AudioContent` alias go. In the Gemini converters, collect every audio blob into a list instead of keeping the last part. On the complete message, wrap `audio/L16` in a WAV header with the rate parsed from the mime type (other formats pass through). On the stream, keep pieces raw like OpenAI's `pcm16` and build `ChoiceDeltaAudio(data=..., transcript=...)`.

**Suggested tests:**

- `test_convert_chunk_response_keeps_partial_audio_delta` in `test_openai_utils.py`, parametrized over the three real pieces above through `_convert_completion_chunk_response`. Fails on this head, passes with the type.
- `test_convert_response_wraps_pcm_audio_as_wav`, parametrized over rate 24000 and 16000, with two `audio/L16` parts: asserts `RIFF`, the rate at bytes 24:28, and the joined payload.
- `test_streaming_completion_keeps_pcm_audio_raw`: two parts in one chunk, joined, no header.
- `test_convert_response_skips_empty_audio_blob`.

Both commits are on [JamMaster1999:pr1344-audio](https://github.com/JamMaster1999/any-llm/commits/pr1344-audio) on top of cf55b68 to cherry-pick (a52c375 WAV and join, bfc265f delta type). Unit suite and pre-commit pass there. Live on that branch:

| Path | Result |
|---|---|
| OpenAI `gpt-audio-mini`, stream, `pcm16` | 30 chunks, 194,400 bytes of audio, transcript intact, one id. Fails on this head, passes here. |
| OpenAI `gpt-audio-mini`, plain, `wav` | Unchanged, valid RIFF on `message.audio`. |
| Gemini `gemini-2.5-flash-preview-tts`, stream | One chunk with one raw piece of 96,526 bytes on `delta.audio`, `finish_reason` stop. |
| Gemini `gemini-2.5-flash-preview-tts`, plain | `message.audio` is a WAV that `file` reads as 16-bit mono 24000 Hz. |

JamMaster1999 (approved):
Verified at 9d0d6a5. The audio half now has the shape from the last round: `ChoiceDeltaAudio` with every field optional, all audio parts collected, `audio/L16` wrapped as WAV on the message with the rate parsed from the mime type, and raw pieces on the stream. The four suggested tests are in, and the rate parse gained a guard against bad values.

CI has not run on this PR (the fork workflows are still waiting for approval), so I ran the two workflows locally on this head:

- `uv sync --group tests --extra all` then `pytest tests/unit`: 2426 passed, 92 skipped.
- `pre-commit run --all-files`: every hook passes. mypy only reports missing `voyageai` and `ibm_watsonx_ai` because pyproject excludes both on Python 3.14; nothing in this PR touches them.

The media conversion code matches the branch I tested live in the previous review, apart from the rate guard and docstrings, so the `gpt-audio-mini` and `gemini-2.5-flash-preview-tts` results there carry over. A maintainer still needs to approve the workflow runs so Unit Tests and Lint show up on the PR.

JamMaster1999:
Live results on 9d0d6a5, all through any-llm on this head. Files checked with `file` where a format is named.

| Provider / model | Mode | Result |
|---|---|---|
| OpenAI `gpt-audio-mini`, `pcm16` | stream | 18 chunks, 122,400 audio bytes, transcript intact, one id, `expires_at` seen. Failed on cf55b68, passes here. |
| OpenAI `gpt-audio-mini`, `wav` | plain | Valid RIFF on `message.audio`, 16-bit mono 24000 Hz. |
| Gemini `gemini-2.5-flash-preview-tts` | stream | One chunk, one raw piece of 90,766 bytes on `delta.audio`, no header, finish stop. |
| Gemini `gemini-2.5-flash-preview-tts` | plain | `message.audio` is a WAV, 16-bit mono 24000 Hz, data length matches payload. |
| Gemini `gemini-2.5-flash-image` | plain | 1024x1024 PNG on `message.images`, text content alongside, finish stop. |
| Gemini `gemini-2.5-flash-image` | stream | `delta.images` carries the PNG data URL, finish stop. |
| Gemini `gemini-2.5-flash-image` | history replay | The image message sent back as history, follow-up answered. |
| OpenRouter `openai/gpt-5.4-image-2` | plain | `message.images[0]` parses as `ImageContent`, 1024x1024 PNG, finish stop. |
| OpenRouter `openai/gpt-5.4-image-2` | stream | `delta.images` parses as `ImageContent`, 1024x1024 PNG, finish stop. |
| OpenRouter `openai/gpt-5.4-image-2` | history replay | Follow-up answered. |
| OpenRouter `google/gemini-2.5-flash-image` | plain, stream, replay | Same as above, typed images and valid PNGs. |

The OpenRouter rows go through the OpenAI-compatible passthrough, so the new `images` type is confirmed on a second converter, not only the Gemini one. Two notes for the record: OpenAI sends no `finish_reason` on audio streams (raw SDK, this head, and main all agree), and OpenAI's own Chat Completions endpoint does not serve `gpt-image-2` at all, so the openai provider has no image path to regress.

Thank you @mikemikimike for the changes. 
@javiermtorres would appreciate another look when you get the chance.

JamMaster1999:
@HareeshBahuleyan this needs a maintainer approval to unblock. Javier's Copilot follow-ups were addressed in 1f519fa, and I ran it live on the current head 9d0d6a5 across Gemini, OpenAI and OpenRouter (table above). It's the fix for #1295 and we're carrying it on a fork downstream in the meantime.

javiermtorres (approved):

PR #1390, opened 2026-09-11
feat(gemini): accept OpenAI input_audio content parts

JamMaster1999:
## Description

OpenAI's audio input part is `{"type": "input_audio", "input_audio": {"data": "<base64>", "format": "wav"}}` ([API reference](https://developers.openai.com/api/docs/api-reference/chat/create)). The OpenAI provider passes it through untouched, so audio-capable models already take it. The Gemini converter skipped it with a debug log, so the one part shape that works on both providers went nowhere on Gemini.

This adds an `input_audio` branch to `_convert_messages` next to `image_url` and `file`. It base64-decodes the data, runs the existing 20 MB inline guard, and builds `Part.from_bytes` with `mime_type="audio/<format>"`. Gemini spells every audio format it accepts as `audio/<name>` (wav, mp3, aiff, aac, ogg, flac, mpeg, m4a, l16, opus, alaw, mulaw, webm, per the [audio guide](https://ai.google.dev/gemini-api/docs/audio)), so the format name maps straight to the MIME type with no table. The base64 decode is factored out of `_parse_data_uri` and shared.

Verified live on this branch. A WAV of "The quick brown fox jumps over the lazy dog" from `gpt-4o-mini-tts`, sent as `input_audio` with "Transcribe this audio exactly":

| provider / model | reply | usage |
|---|---|---|
| gemini `gemini-2.5-flash` | `The quick brown fox jumps over the lazy dog.` | in=122 out=32 |
| openai `gpt-audio-mini` | `The quick brown fox jumps over the lazy dog.` | in=55 out=10 |

Tests: `test_convert_messages_with_input_audio` (MIME and bytes on the part) and `test_convert_messages_input_audio_without_format_raises_invalid_request`. The oversized case shares the existing guard test. `tests/unit/providers/test_gemini_provider.py`, 266 passed. Pre-commit clean.

## PR Type

- 🆕 New Feature

## Relevant issues

None open. Audio output landed in #1344; this is the input side.

## Checklist

- [x] **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

- AI Model used: Claude (Fable 5.1)

https://claude.ai/code/session_01546kUvB5GcyCkhQVSjpjbk

## Summary by CodeRabbit

- **New Features**
  - Added support for converting OpenAI audio input into Gemini-compatible audio content.
  - Supports audio formats such as WAV and assigns the appropriate MIME type.

- **Bug Fixes**
  - Enforced the 20 MB limit for encoded image and file data before decoding.
  - Improved validation and error handling for malformed or non-ASCII base64 audio data.
  - Invalid audio input now produces clear request validation errors.
  - Images exactly at the 20 MB inline upload limit are now accepted.


JamMaster1999:
Both CodeRabbit findings addressed in 69fbb92: a non-dict `input_audio` value and non-ASCII base64 now raise `InvalidRequestError` instead of `AttributeError` / `ValueError`. The base64 change also covers the existing data-URI image and file paths, which had the same gap. Tests added for each, 269 passed.

JamMaster1999:
Round two addressed in 7f3dd44:

- **Size check before decode**: the 20 MB guard now runs on the encoded length inside `_decode_base64`, so nothing oversized is decoded first. Every inline path (image, file, audio) shares it and `_validate_inline_size` is gone.
- **Format must be a non-empty string**: a truthy non-string such as `1` is an `InvalidRequestError` now, with a regression case.
- **Restricting format to wav/mp3**: not taken. Gemini accepts thirteen audio formats (wav, mp3, aiff, aac, ogg, flac, mpeg, m4a, l16, opus, alaw, mulaw, webm), all spelled `audio/<name>`. Rejecting flac or ogg here would block formats the provider takes for no gain, and the messages list is untyped so there is no OpenAI enum to honour. An unknown format gets Gemini's own 400, typed by the handler.

JamMaster1999:
Good catch, fixed in 3c7a89a: the encoded limit is now `(_INLINE_SIZE_LIMIT + 2) // 3 * 4`, which is exactly the base64 length of a 20 MB payload. Added `test_convert_messages_image_at_inline_limit_is_accepted` as the boundary pair to the existing one-byte-over test.

JamMaster1999:
Correction in 3178ac0: an encoded-length cap can't separate 20 MB from 20 MB + 1 when the limit isn't a multiple of 3, so 3c7a89a let the one-byte-over case through. The guard now derives the decoded size from the base64 length and padding and compares that to the byte limit. Exact on both sides of the boundary, still no decode before the check. Both boundary tests pass.

JamMaster1999:
That review ran against 3c7a89a, which was already superseded by 3178ac0 when it posted. On the current head the guard compares the decoded size, derived from the base64 length and padding, against `_INLINE_SIZE_LIMIT`, so 20 MB + 1 is rejected on every inline path. `test_convert_messages_oversized_image_raises_invalid_request` is the one-byte-over regression and `test_convert_messages_image_at_inline_limit_is_accepted` pins the other side. Unpadded input is not a bypass either: `b64decode(validate=True)` rejects it.

HareeshBahuleyan (approved):

HareeshBahuleyan:
Thanks @JamMaster1999 for the contribution!

PR #1249, opened 2026-08-07
fix(minimax): preserve usage-only streaming chunks

zhuziqi97:
## Description

Fixes MiniMax streaming usage loss reported in #1185.

MiniMax returns token usage in a trailing streaming chunk with `choices=[]`.
`MinimaxProvider._convert_completion_response_async()` previously forwarded
only chunks containing a choice delta, so it discarded this usage-only chunk
before the Messages bridge or downstream stream handler could observe it.

This change keeps the existing MiniMax filtering behavior while also forwarding
chunks where `chunk.usage is not None`.

The regression test verifies that:

- normal content chunks remain available;
- unrelated empty chunks without usage remain filtered;
- a trailing `choices=[]` chunk with usage survives conversion;
- input, output, and total token values remain unchanged after the XML reasoning
  wrapper.

No usage is synthesized, and absent usage is not converted to zero. Non-streaming
behavior and other providers are unchanged.

### Validation

Full local unit suite:

```text
1920 passed, 68 skipped, 3 warnings
```

The warnings are existing provider exception deprecation warnings unrelated to
this change.

Full pre-commit suite passed, including Ruff, formatting, Mypy, codespell, and
repository file checks.

A live MiniMax-M3 streaming request was also verified with the patched provider.
The raw provider usage and normalized result matched:

```text
input_tokens: 191
output_tokens: 27
total_tokens: 218
cached_tokens: 128
```

No credentials, prompt text, response text, or request headers were retained.

## PR Type

- [x] Bug Fix

## Relevant issues

Fixes #1185

## Checklist

- [x] I understand the code I am submitting.
- [x] I have added unit tests that prove my fix/feature works
- [x] I have run this code locally and verified it fixes the issue.
- [x] New and existing tests pass locally
- [x] Documentation was updated where necessary
- [x] I have read and followed the [contribution guidelines](https://github.com/mozilla-ai/any-llm/blob/main/CONTRIBUTING.md)
- [x] **AI Usage:**
    - [ ] No AI was used.
    - [ ] AI was used for drafting/refactoring.
    - [x] This is fully AI-generated.

## AI Usage Information

- AI Model used: OpenAI GPT-5.6 Sol
- AI Developer Tool used: Codex
- Any other info you'd like to share: The scope and acceptance criteria were
  directed by the contributor. Codex implemented the change, added the
  regression test, ran the local validation, and drafted this PR description.
  The contributor reviewed the patch before submission.

[...]

Issue #1184, opened 2026-07-16
[BUG] `output_format` not working with Bedrock > Claude


AbdealiLoKo:
**`any-llm` version:** 1.19.0
**Python Version**: 3.11

I have been using any-llm-sdk with openAI and Gemini for a while. Recently trying to use it with bedrock, and facing some issues.

### Describe the bug
When I try to use `ouput_format` with the messages API, I get an error in bedrock. This error is confusing and the link is also invalid as I am using claude on bedrock ... not amazon's nova model.
```py
  File "/opt/corridor/venv/lib/python3.11/site-packages/any_llm/providers/bedrock/utils.py", line 47, in _convert_params
    raise UnsupportedParameterError(
any_llm.exceptions.UnsupportedParameterError: [bedrock] 'response_format' is not supported for bedrock.
Check the following links:
- https://docs.aws.amazon.com/nova/latest/userguide/prompting-structured-output.html
```

### Example w/ Code

When I run in Gemini:
```py
from any_llm import AnyLLM, LLMProvider

AnyLLM.create(provider=LLMProvider.GEMINI, api_key='...')
out = client.messages(
    model='gemini-2.5-flash',
    messages=[{"role": "user", "content": 'tell me a joke, return valid json'}],
    max_tokens=10240,
    temperature=0,
    top_p=0.4,
    output_format={"format": {"type": "json_object"}},
)
print(out.content[0].text)

### OUTPUT: 
'{\n  "joke": "Why don\'t scientists trust atoms? Because they make up everything!"\n}'
```

The bedrock example that throws an error:
```py
from any_llm import AnyLLM, LLMProvider

aws_access_key_id='....'
aws_secret_access_key='....'
aws_session_token='.....'
aws_region='us-west-2'

client = AnyLLM.create(
  LLMProvider.BEDROCK,
  aws_access_key_id=aws_access_key_id,
  aws_secret_access_key=aws_secret_access_key,
  aws_session_token=aws_session_token,
  region_name=aws_region,
)

out = client.messages(
    model='us.anthropic.claude-sonnet-4-20250514-v1:0',
    messages=[{"role": "user", "content": 'tell me a joke, return valid json'}],
    max_tokens=10240,
    temperature=0,
    top_p=0.4,
    output_format={"format": {"type": "json_object"}},
)
print(out)

## OUTPUT:
Traceback (most recent call last):
[...]

PR #1392, opened 2026-09-11
fix(anthropic): leave choices empty on the usage-only message_stop chunk

ranzhh:
## Description

Native Anthropic streaming reports token usage on the `message_stop` event, which arrives after the `message_delta` event that carries `stop_reason`. The chunk converter attached a synthetic choice (`index 0`, empty delta, `finish_reason: None`) to every chunk, including that usage-only one. OpenAI-compatible providers put final usage on a trailing chunk with `choices: []` (OpenAI documents this for `stream_options.include_usage`), and the MiniMax fix in #1249 aligned MiniMax to the same shape. Code that picks up usage with `if not chunk.choices` therefore worked for OpenAI and silently saw no usage for Anthropic.

Observed on `main` (last two chunks of a streamed `claude-sonnet-4-6` reply, null fields dropped):

```
8 {'choices': [{'delta': {}, 'finish_reason': 'stop', 'index': 0}]}
9 {'choices': [{'delta': {}, 'index': 0}], 'usage': {'completion_tokens': 5, 'prompt_tokens': 13, 'total_tokens': 18}}
```

With this change chunk 9 becomes `{'choices': [], 'usage': {...}}`, identical in shape to OpenAI's trailing usage chunk. Chunks 1 to 8 are unchanged.

Change: `_create_openai_chunk_from_anthropic_chunk` returns from the `MessageStopEvent` branch before the choice is appended. The stop event carries no delta and no stop reason, so nothing is lost.

Tests: the two existing `message_stop` usage tests now assert `choices == []`; two new converter tests pin that the usage chunk arrives after the `finish_reason` chunk with empty choices, and that a raw `message_stop` without an accumulated message yields neither choices nor usage. A third test drives `acompletion(stream=True)` through a fake SDK message stream and reads the chunks the way an OpenAI-style consumer does (content and `finish_reason` from chunks with choices, usage from the chunk without). On `main` that test fails with `usage is None`; text and `finish_reason` already arrive correctly.

Validation:

- `uv run pytest tests/unit`: 2435 passed, 69 skipped
- `uv run pre-commit run --all-files`: clean (ruff, ruff format, mypy strict, codespell)
- Live streams against Anthropic (`claude-sonnet-4-6`) and OpenAI (`gpt-5-nano`, `include_usage`) printed chunk by chunk; usage chunk shapes match after the fix
- `uv run pytest tests/integration -k anthropic` with real keys: 23 passed, 7 skipped. The skips are the existing capability skips (Anthropic has no batch, responses, moderation, or embeddings endpoints; one test targets Claude on Bedrock only, see #1184)

Not changed, same pattern: Bedrock also attaches a synthetic choice to its `metadata` usage chunk (`src/any_llm/providers/bedrock/utils.py`). Gemini attaches usage to every content chunk and never emits a usage-only chunk.

## PR Type

- 🐛 Bug Fix

## Relevant issues

None open. Same shape as the MiniMax fix in #1249.

## Checklist
- [x] **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

- AI Model used: Claude Fable 5.1 (`claude-fable-5-1`)
- Any other info you'd like to share: The bug was reported by an automated agent; the fix, tests, and this description were produced by Claude Code and reviewed by the submitter, who will answer reviewer questions personally.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected streaming responses so the final usage-only chunk no longer includes an empty choice.
  * Improved compatibility with OpenAI-style streaming consumers by delivering usage after the finish reason.
  * Ensured message-stop events produce the expected empty choices list, including when no message content has been accumulated.
  * Standardised streaming completion behaviour so consumers receive text, a single stop reason, and final usage in the correct order.


ranzhh:
Human here, ping me if you have any questions. This is to fix an issue I encountered mid langchain-anyllm usage, and I believe this half of the fix belongs upstream to align with the "let's standardise messages to OpenAI format" like in #1249 .

HareeshBahuleyan:
Thank you, @ranzhh! I’ve fixed the minor import cleanup in the tests. Otherwise, LGTM, pending CI.

HareeshBahuleyan (approved):

ranzhh:
> Thank you, @ranzhh! I’ve fixed the minor import cleanup in the tests. Otherwise, LGTM, pending CI.

Cheers, i completely missed that comment or I'd have fixed it myself. Thanks for taking a look @HareeshBahuleyan.

PR #1393, opened 2026-09-11
fix(bedrock): keep the assistant text when the turn also calls a tool

shoemoney:
## Description

Sibling gap from #1325, which fixed this exact shape in Gemini and Anthropic and left Bedrock with it.

`_convert_response` collects every `text` content block into `content_parts`, then in the `stopReason == "tool_use"` branch builds the message with `content=None` and discards all of it. Converse returns the model's own text as a separate text block in the same message as the `toolUse` block, so a turn like "I will look up the weather in Paris." followed by a `get_weather` call reached the caller as a bare tool call with no text.

Three things keep it quiet rather than loud:

1. The streaming converter already forwards those deltas (`_create_openai_chunk_from_aws_chunk` sets `content = delta["text"]`), so streaming and non-streaming disagreed about the same response.
2. `_convert_message_to_bedrock_format` already replays an assistant turn as a text block followed by `toolUse` blocks, so the request direction was fine. There was simply never any text to replay, because the response direction had dropped it.
3. There is no API error. The model just cannot see that it already announced its plan, so on the next turn it announces it again, which is the failure #1325 describes.

The non-tool branch of the same function joins `content_parts` correctly. Only the tool branch threw them away.

`"".join(content_parts) or None` preserves the existing contract for a bare tool call: joining an empty list yields `""`, and reporting an empty string where the model genuinely said nothing would be a behaviour change of its own. This matches Gemini, whose `text_content` stays `None` until a text part appears.

I swept the rest of the provider family for the same pattern while I was in here. `grep -rn 'content=None' src/any_llm/providers/*/utils.py src/any_llm/providers/*/base.py` returns this one site and nothing else, so Bedrock was the last one.

### Tests

3 added. The first two fail against unpatched sources with:

```
assert None == 'I will look up the weather in Paris.'
```

The third covers a tool call carrying no text and passes both before and after, deliberately: it is the guard that the `or None` did not turn every bare tool call into an empty string.

## PR Type

- 🐛 Bug Fix

## Relevant issues

No open issue; found by sweeping the provider family for the pattern #1325 fixed.

## Checklist

- [x] **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

- AI Model used: Opus 5
- Any other info you'd like to share: `tests/unit` 2542 passed, 66 skipped; pre-commit clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Non-streaming Bedrock responses now preserve model-generated text when it appears alongside a tool call.
  * Multiple text sections are combined correctly, while tool calls without accompanying text continue to show no message content.

* **Tests**
  * Added coverage for text before tool calls, multiple text sections, and tool calls without text.


HareeshBahuleyan (approved):
Thanks for the fix  @shoemoney

PR #1402, opened 2026-09-15
fix(messages): carry tool result is_error in content, not as a message key

### LINKED ISSUE #1401: Messages bridge forwards is_error on role: tool messages; strict OpenAI-compatible providers reject the request
`_convert_user_blocks_to_openai` in `utils/messages_compat.py` copies a tool result's `is_error: true` onto the emitted `role: tool` message (added in #1312). OpenAI's schema has no such field, and strict OpenAI-compatible backends reject the whole request: Fireworks returns `400 Extra inputs are not permitted, field: 'messages[N].is_error'`. Any conversation containing a failed tool result then fails on every later turn.

Proposed fix: stop emitting the key and carry the signal in the content instead (an `Error: ` prefix), which every backend reads. Present since 1.27.0.

No integration test sends a failed tool result through `messages()`, so the suite cannot catch this; the fix should add one.

Repro: `amessages(model="fireworks:<model>", messages=[..., {"role": "user", "content": [{"type": "tool_result", "tool_use_id": "...", "content": "failed", "is_error": True}]}])`.

njbrake:
## Description

The Messages bridge sent two keys the OpenAI schema has no field for: a tool result's `is_error`, and `extra_content`, any_llm's signature side-channel, on messages and tool calls. Strict backends reject both, so a later turn 400s: Fireworks on `is_error` (#1401), Groq and Cerebras on the bridge's `reasoning_content`.

`is_error` now becomes an `Error: ` prefix on the tool text. `strip_extra_content` drops the any_llm-internal namespaces, keeping the `google` one that Gemini's OpenAI-compatible API requires back. Groq and Cerebras rename `reasoning_content` to `reasoning`, which both SDKs document. Tool results also keep `search_result`, `tool_reference` and `browser_state` blocks, rendered as text.

Adds an integration test replaying a failed tool result through `messages()`.

## PR Type

- 🐛 Bug Fix

## Relevant issues

Fixes #1401. Follow-up to #1312.

## Checklist
- [x] **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility when sending failed tool results to OpenAI-compatible providers.
  * Failed tool results now include an `Error:` prefix, with a consistent fallback when no details are provided.
  * Unsupported tool metadata is no longer sent, while supported images and documents remain available.
  * Improved handling of search results, tool references and browser state content.
  * Preserved supported reasoning content for Cerebras, Groq and OpenAI-compatible requests while omitting unsupported metadata.

* **Tests**
  * Added coverage for tool-result conversion and reasoning handling across providers.


src/any_llm/utils/reasoning.py
```diff
@@ -182,6 +182,28 @@ async def process_streaming_reasoning_chunks(
         yield held_chunk
 
 
+def replay_reasoning_content_as_reasoning(messages: list[dict[str, Any]]) -> list[dict[str, Any]]:
+    """Rename a replayed ``reasoning_content`` to ``reasoning`` and drop ``extra_content``.
+
+    For providers whose SDK names the assistant reasoning field ``reasoning`` and whose API rejects
+    any other message key: Groq and Cerebras return 400 ``property 'reasoning_content' is
+    unsupported`` for the field the Messages bridge emits from a replayed ``thinking`` block.
+    ``extra_content`` is an any_llm side-channel that never belongs on their wire. An explicit
```
HareeshBahuleyan: `extra_content` has the same problem as `is_error` on every strict OpenAI-compatible backend, not just Groq and Cerebras. The bridge emits `extra_content={"anthropic": {"signature": ...}}` whenever a replayed thinking block carries a signature (`messages_compat.py:235`). Only Anthropic-native responses carry one, so the new integration test never puts `extra_content` on a Fireworks wire. A conversation started on `anthropic:` with thinking and continued on `fireworks:` should 400 on `messages[N].extra_content`, same class as #1401.

Either strip it generically (bridge or `BaseOpenAIProvider`) and re-add only for providers that read it (anthropic, gemini, deepseek), or file a follow-up issue and link it here.

src/any_llm/providers/groq/groq.py
```diff
@@ -147,6 +148,7 @@ async def _stream() -> AsyncIterator[ChatCompletionChunk]:
     async def _acompletion(
         self, params: CompletionParams, **kwargs: Any
     ) -> ChatCompletion | AsyncIterator[ChatCompletionChunk]:
+        params.messages = replay_reasoning_content_as_reasoning(params.messages)
```
HareeshBahuleyan, line 151: This runs on every `completion()` to Groq, not only the Messages bridge. A caller passing DeepSeek-style `reasoning_content` dicts directly gets a silent rename. Probably what we want since Groq would 400 otherwise, but the PR description frames it as bridge-only. Worth a line in the helper docstring saying plain completions are affected too.

src/any_llm/utils/reasoning.py
```diff
@@ -182,6 +182,28 @@ async def process_streaming_reasoning_chunks(
         yield held_chunk
 
 
+def replay_reasoning_content_as_reasoning(messages: list[dict[str, Any]]) -> list[dict[str, Any]]:
```
HareeshBahuleyan, line 235: Optional: `deepseek/utils.py:53` `_reinject_reasoning_content` already strips `extra_content` per message with the same loop shape. Could fold both into one helper here parameterised by target field name. Fine as a follow-up.

src/any_llm/utils/messages_compat.py
```diff
@@ -371,14 +371,15 @@ def _convert_user_blocks_to_openai(blocks: list[dict[str, Any]]) -> list[dict[st
...
             tool_text, extra_parts = _convert_tool_result_content(block.get("content", ""))
-            tool_message: dict[str, Any] = {
-                "role": "tool",
-                "tool_call_id": block.get("tool_use_id", ""),
-                "content": tool_text,
-            }
             if block.get("is_error") is True:
-                tool_message["is_error"] = True
-            results.append(tool_message)
+                tool_text = f"Error: {tool_text}" if tool_text else "Error"
```
HareeshBahuleyan: Nit: tool text already starting with `Error:` becomes `Error: Error: ...`. Cheap guard:

```suggestion
                if not tool_text:
                    tool_text = "Error"
                elif not tool_text.lstrip().lower().startswith("error"):
                    tool_text = f"Error: {tool_text}"
```

src/any_llm/utils/messages_compat.py
HareeshBahuleyan, line 1: Docstring of `_convert_assistant_blocks_to_openai` (line 180, outside this diff) is stale now: Groq and Cerebras rename `reasoning_content` to `reasoning` before it reaches the wire. One sentence pointing at `replay_reasoning_content_as_reasoning` keeps this accurate.

HareeshBahuleyan:
@njbrake Would you like to finish up and merge this PR as well to include it in the next release? No pressure :)

HareeshBahuleyan (approved):

PR #1420, opened 2026-09-22
fix(gemini): send tool results as a user turn, not role 'function'

MRenAIAgent:
## Description

`_convert_messages` emits every OpenAI `role: "tool"` message as `types.Content(role="function", parts=[functionResponse])`. `function` is not a documented Gemini `Content.role` (only `user` and `model` are); the API's own convention — and what google-genai / google-adk emit — is a **`user` turn carrying `functionResponse` parts**.

Older models tolerate the legacy role, but **Gemini 3 flash models reject it**, which breaks any multi-turn function-calling loop (the first tool call works; sending the result back fails):

```
400 INVALID_ARGUMENT. Role 'function' is not supported. Please use a valid role: SYSTEM, SYSTEM_1, USER, ASSISTANT, DEVELOPER, CONTEXT, USER_CONTEXT, MODEL, USER.
```

Reproduced on a clean `any-llm-sdk[gemini]==1.28.0` install with a minimal 2-turn conversation (assistant `tool_calls` + one `tool` result), Gemini API provider:

| model | `role="function"` (current) | `role="user"` (this PR) |
|---|---|---|
| gemini-2.5-flash | OK | OK |
| gemini-3.7-flash | **400 Role 'function' is not supported** | OK |
| gemini-3.8-flash | **400 Role 'function' is not supported** | OK |

Also verified with this patch that two parallel tool results, emitted as consecutive `user` turns, are accepted by gemini-3.7-flash — so no merging of consecutive tool messages is needed.

Change: one line in `src/any_llm/providers/gemini/utils.py`; the three unit assertions that pinned the old role are updated to `user`.

Related reports of the same behaviour in other clients: mozilla-ai/any-llm has none yet; see e.g. kurojs/Gemini-Kchat#5, agno-agi/agno#2298, BerriAI/litellm#41409, and the Google AI forum thread "Gemini API returns an error when trying to pass tool call results with role: tool".

## PR Type

- 🐛 Bug Fix

## Relevant issues

None filed; discovered in production (Google Docs/Slides editing agents on gemini-3.7-flash failing on the tool-result turn).

## Checklist
- [x] I have added unit tests that prove my fix/feature works (existing role assertions updated to the correct role; `tests/unit/providers/test_gemini_provider.py`: 286 passed)
- [x] I have run this code locally and verified it fixes the issue (table above, live Gemini API)
- [x] Documentation was updated where necessary (n/a — no documented behaviour changes)
- [x] **AI Usage:**
    - [ ] No AI was used.

## AI Usage Information
Root-cause investigation, the reproduction against the live Gemini API, and this PR text were done with Claude Code (Anthropic) under human direction and review; the change itself is a one-line role fix with the corresponding test updates.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Gemini tool-result message handling to use the supported `user` role, preventing 400 errors with Gemini 3 Flash models.
  * Improved compatibility when processing tool responses supplied as JSON or parsed content.


deepak-zyphra (approved):

njbrake (changes requested):
Three changes before this merges. The role fix itself is right: only `user` and `model` are documented Content roles, and google-genai emits `user` for function responses.

**1. Group parallel tool results into one turn.** With the role changed, N parallel results now emit N consecutive `user` turns, and a user message following a tool result produces back-to-back `user` turns. google-genai's own function-calling loop puts all results in one turn (`chats.py`: `types.Content(role="user", parts=func_response_parts)`), and both sibling converters here already do the same: `anthropic/utils.py` ("Merge consecutive tool results into a single user message") and `bedrock/utils.py` (`flush_tool_results`).

```python
def _pending_function_response_parts(contents: list[types.Content]) -> list[types.Part] | None:
    if not contents:
        return None
    last = contents[-1]
    if last.role != "user" or not last.parts:
        return None
    if any(part.function_response is None for part in last.parts):
        return None
    return last.parts
```

```python
if (pending := _pending_function_response_parts(formatted_messages)) is not None:
    pending.append(part)
else:
    formatted_messages.append(types.Content(role="user", parts=[part]))
```

**2. Narrow the comment.** `gemini-3-flash-preview` is a Gemini 3 Flash model, is what `tests/conftest.py` pins for the integration suite, and passed `test_agent_loop_multiple_tool_calls[gemini]` on main with `role="function"`. Name the models the 400 was seen on:

```python
# The API documents only "user" and "model" as Content roles. "function" is tolerated by
# older models and rejected with 400 INVALID_ARGUMENT by gemini-3.7-flash and 3.8-flash.
```

**3. Fix the test helper.** `_function_response_names` indexes `content.parts[0]` behind a `parts is None` guard, so a user turn with an empty parts list raises IndexError, and it will miss the second response once results are grouped.

```python
for content in contents:
    if content.role != "user":
        continue
    for part in content.parts or []:
        if part.function_response is not None:
            names.append(part.function_response.name)
```

Cover the grouping, a tool result after a text turn and after a partless turn, a leading tool result, and two rounds separated by a model turn.

Rebase on main while you are in there. The fork is org-owned, so the maintainer-edit grant does not apply and I cannot push these for you.

Separate from this PR: `FunctionResponse.id` is never set, and Google's 3.8 migration checklist requires `call_id` and `name` on every `FunctionResponse` under `generateContent`. `Part.from_function_response()` takes no id argument. Worth its own issue if you are running 3.8.

MRenAIAgent:
Thanks @njbrake — all three addressed in 407443a → HEAD, rebased on `main` (57186d5).

1. **Parallel results grouped into one `user` turn.** Added `_pending_function_response_parts()` as proposed; the tool branch now appends to a trailing all-`functionResponse` user turn and only opens a new turn otherwise (so a text user turn or a partless turn is never merged into). Live-checked the grouped shape (one `user` turn, two `functionResponse` parts) against gemini-3.7-flash and 2.5-flash — accepted.
2. **Comment narrowed** to the models the 400 was actually observed on: gemini-3.5-flash-lite, 3.7-flash and 3.8-flash (I had also reproduced it on 3.5-flash-lite). Dropped the "Gemini 3 flash" generalisation — you're right that `gemini-3-flash-preview` accepts `function`.
3. **Test helper** iterates every part of every user turn (no `parts[0]`). New tests cover: grouping of two parallel results; a result after a text user turn stays a separate turn; partless / empty / model-turn cases for the helper (and that it returns the turn's own list, so appending extends the turn in place); a leading tool result; two rounds separated by a model turn stay separate turns.

`tests/unit/providers/test_gemini_provider.py`: 291 passed; ruff, ruff format, mypy and pre-commit clean.

On `FunctionResponse.id`: agreed it's separate from this PR — we run 3.7-flash with 3.8 in the fallback chain, so we'll hit the 3.8 checklist too; happy to open an issue for it.

njbrake:
Thanks @MRenAIAgent! @daavoo or @HareeshBahuleyan should be able to get this over the finish line. Thank you for the contribution 🙏

HareeshBahuleyan:
@MRenAIAgent could you please fix the linting errors and push the changes. Apart from that we should be good to merge. The integration test failures are unrelated to this change.

tests/unit/providers/test_gemini_provider.py
```diff
@@ -2409,13 +2410,99 @@ def test_convert_messages_with_thought_signature_in_extra_content() -> None:
...
+        kinds = [
+            "function_response" if p.function_response is not None else "function_call" if p.function_call else "text"
+            for p in content.parts or []
+        ]
+        shapes.append((content.role or "", kinds))
+    return shapes
+
+
+def test_convert_messages_groups_parallel_tool_results_into_one_user_turn() -> None:
+    messages = [
```
daavoo, line 2442:
```suggestion
    messages: list[dict[str, Any]] = [
```

tests/unit/providers/test_gemini_provider.py
```diff
@@ -2409,13 +2410,99 @@ def test_convert_messages_with_thought_signature_in_extra_content() -> None:
...
+    assert _turn_shapes(contents) == [
+        ("user", ["text"]),
+        ("model", ["function_call", "function_call"]),
+        ("user", ["function_response", "function_response"]),
+    ]
+    assert _function_response_names(contents) == ["get_weather", "get_time"]
+
+
+def test_convert_messages_does_not_merge_a_tool_result_into_a_text_user_turn() -> None:
+    messages = [
```
daavoo, line 2458:
```suggestion
    messages: list[dict[str, Any]] = [
```

tests/unit/providers/test_gemini_provider.py
```diff
@@ -2409,13 +2410,99 @@ def test_convert_messages_with_thought_signature_in_extra_content() -> None:
...
+
+
+def test_convert_messages_accepts_a_leading_tool_result() -> None:
+    contents, _ = _convert_messages([_tool_result("c1", "get_weather")], provider_name="gemini")
+    assert _turn_shapes(contents) == [("user", ["function_response"])]
+    assert _function_response_names(contents) == ["get_weather"]
+
+
+def test_convert_messages_keeps_tool_results_of_separate_rounds_in_separate_turns() -> None:
+    messages = [
```
daavoo, line 2488:
```suggestion
    messages: list[dict[str, Any]] = [
```

daavoo (commented):
Thanks, the fix looks good and the Gemini integration tests pass. The only blocker is `run-linter`: mypy infers these mixed dict literals as `list[object]` and rejects them as `_convert_messages` arguments (`[arg-type]`). Annotating the three lists fixes it; I have verified locally that `pre-commit run --all-files` and `pytest tests/unit` then pass. Could you apply the suggestions ("Add suggestions to batch", then commit)? Pushing to the Zyphra fork is not possible for maintainers. The `run-integration-tests` failures are all `inception` returning HTTP 500, which is unrelated to this change.

daavoo:
Opened #1428 with your commits plus the suggested mypy fix, since maintainers can't push to the Zyphra fork. Your authorship is preserved; thanks for the fix!

PR #1421, opened 2026-09-22
feat(registry): migrate cascadia, vllm and llamacpp to registry rows (#1417)

### LINKED ISSUE #1417: Migrate cascadia, vllm and llamacpp to registry rows
Migrate three local-runner providers to `PROVIDER_REGISTRY` rows now that `api_key_optional` exists (#1416). Do them in this order; each is a larger behavior change than the last.

1. **cascadia.** `CascadiaProvider._verify_and_set_api_key` resolves identically to `_OptionalApiKeyProvider`, so the row is a no-op. Land it with #1416 or straight after, so the flag has a consumer in tree.
2. **vllm.** A behavior change, and a fix. `VllmProvider` declares `ENV_API_KEY_NAME = "VLLM_API_KEY"` and never reads it, while `scripts/generate_provider_table.py` publishes it as the provider's key env var. Anyone running vLLM with `--api-key` and exporting `VLLM_API_KEY` sends `no-key-required` today and gets a 401. Call the change out in the PR; someone may depend on the current behavior.
3. **llamacpp.** The largest change. `LlamacppProvider._verify_and_set_api_key` discards a caller-supplied key, and its `ENV_API_KEY_NAME` is the literal string `"None"`. A row needs a real name, so this means adopting `LLAMACPP_API_KEY` and honoring both it and explicit keys. llama.cpp's server supports `--api-key`, so that is the correct end state.

**Not migratable, contrary to #1412's list.** `llamafile` also overrides `_convert_completion_response` and `_acompletion`. `lmstudio` subclasses `AnyLLM` and drives the LM Studio SDK rather than the OpenAI wire protocol. Both keep their folders.

**Separately:** `OpenAICompatibleProvider` in `openai/custom.py` carries the same three-tier expression and can subclass `_OptionalApiKeyProvider` to drop its copy.

The vllm and llamacpp key-handling defects stand on their own, whether or not the migrations land.

00200200:
## Description

Migrates three local-runner providers (`cascadia`, `vllm`, and `llamacpp`) to `PROVIDER_REGISTRY` rows using `api_key_optional=True` added in #1416, following the plan outlined in #1417.

### Changes
1. **cascadia**: Migrated to a `PROVIDER_REGISTRY` row with `api_key_optional=True`. Replaced `src/any_llm/providers/cascadia/cascadia.py` with an import shim calling `get_registry_provider_class("cascadia")`.
2. **vllm**: Migrated to a `PROVIDER_REGISTRY` row with `api_key_optional=True`. Fixed defect where `VllmProvider` previously declared `ENV_API_KEY_NAME = "VLLM_API_KEY"` but never read it in `_verify_and_set_api_key`. It now properly honors `VLLM_API_KEY` from the environment while still defaulting to `"no-key-required"` when unset. Replaced `src/any_llm/providers/vllm/vllm.py` with an import shim.
3. **llamacpp**: Migrated to a `PROVIDER_REGISTRY` row with `api_key_optional=True`. Fixed `ENV_API_KEY_NAME` which was previously the string `"None"`, adopting `"LLAMACPP_API_KEY"` and honoring both explicit keys and environment variables (useful when llama.cpp server runs with `--api-key`). Replaced `src/any_llm/providers/llamacpp/llamacpp.py` with an import shim.
4. **OpenAICompatibleProvider**: In `src/any_llm/providers/openai/custom.py`, subclassed `_OptionalApiKeyProvider` to drop its duplicate `_verify_and_set_api_key` implementation.
5. **Tests**: Added comprehensive test coverage in `test_vllm_provider.py` and `test_llamacpp.py` verifying metadata, capability flags, loader/enum resolution, and key resolution precedence (explicit key > env key > `"no-key-required"`). Added shim resolution tests for `cascadia`, `llamacpp`, and `vllm` in `test_registry.py`.

## PR Type
- 🆕 New Feature

## Relevant issues
Resolves #1417

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added registry-based configuration for Cascadia, Llama.cpp and vLLM providers.
  - Added provider metadata, documentation links and capability information, including image completion, embeddings and moderation support for Llama.cpp.
  - Preserved existing provider import paths for compatibility.

- **Bug Fixes**
  - Improved API key handling for custom OpenAI-compatible endpoints and supported providers, including optional keys and explicit key precedence.


njbrake (approved):
Both key-handling defects from #1417 are fixed. Against a llama.cpp server started with `--api-key`: llamacpp with an explicit key or `LLAMACPP_API_KEY`, and vllm with `VLLM_API_KEY`, all returned 401 before this change and complete now. Cascadia is the no-op the issue predicted.

Provider metadata across every provider is unchanged except llamacpp's `env_key`, `"None"` to `LLAMACPP_API_KEY`. Capability flags, tiers and class names hold, so the migration is behavior-preserving outside the two intended fixes.

llamacpp integration: 14 passed, 17 skipped, all skips capability-gated. The vllm and cascadia rows serve completion, streaming and list_models on their row defaults; both sit in `CI_EXCLUDED_PROVIDERS`, so CI does not cover them.

Non-blocking follow-ups: llamafile still discards caller-supplied keys and publishes `None` as its key env var; llamacpp's `supports_moderation=True` is wrong, the endpoint 404s.

PR #1422, opened 2026-09-22
feat(messages): expose whether a provider serves the Anthropic Messages API natively (#1418)

### LINKED ISSUE #1418: Expose whether a provider serves the Anthropic Messages API natively
## Problem

`SUPPORTS_MESSAGES` is true for every provider the Messages-to-Completions bridge covers, so a caller cannot tell a provider with a native Messages API (Anthropic, and the few that override `_amessages`) from one that is bridged. The difference matters before the call is made: the bridged base implementation raises `NotImplementedError` when `betas` or `context_management` is set (`any_llm.py`, "context_management and betas require a provider with a native Anthropic Messages API").

A gateway that accepts Anthropic-shaped requests and routes them to other providers has to decide whether to forward `betas`. Forwarding them to a bridged provider fails the whole request over a header the provider was never going to honor; dropping them for a native provider loses a feature the caller asked for. Today the only way to answer is to compare the provider class's `_amessages` against `AnyLLM._amessages`, which is a private attribute and breaks silently on a refactor. Otari does exactly that in `src/gateway/api/routes/messages.py` (`_serves_messages_natively`, mozilla-ai/otari#1366) and would like to stop.

## Proposal

A public class attribute alongside `SUPPORTS_MESSAGES`, for example `MESSAGES_NATIVE: bool = False`, set to `True` by the providers that implement `_amessages` themselves, and surfaced in `get_provider_metadata()` next to `messages`. The bridge's `NotImplementedError` check could then read the same flag.

Any other public answer to "will this provider accept `betas` on a Messages call" works too.

00200200:
## Description

Resolves #1418 by exposing whether a provider serves the Anthropic Messages API natively rather than through any-llm's Messages-to-Completions bridge.

### Motivation
`SUPPORTS_MESSAGES` is `True` for every provider because of the bridge. However, callers and gateways (such as Otari routing Anthropic requests to other providers) need to know if a provider accepts `betas` or `context_management` natively before making a call, because the bridged base implementation raises `NotImplementedError` on them. Previously, gateways had to compare `provider_class._amessages != AnyLLM._amessages`, which relied on private internals.

### Changes
- Added `SUPPORTS_MESSAGES_NATIVE: bool = False` (and alias `MESSAGES_NATIVE: bool = False`) to `AnyLLM`.
- Set `SUPPORTS_MESSAGES_NATIVE = True` and `MESSAGES_NATIVE = True` on `BaseAnthropicProvider` (inherited by Anthropic, AzureAnthropic, VertexAIAnthropic), `MetaProvider`, and `OtariProvider`.
- Surfaced `messages_native: bool = False` in `ProviderMetadata` and wired it in `get_provider_metadata()`.
- Guarded `_amessages` checks for non-native params (`container`, `betas`, `context_management`) with `getattr(self, "SUPPORTS_MESSAGES_NATIVE", False) is not True`.
- Added unit tests in `tests/unit/test_messages.py` verifying class attributes and `ProviderMetadata` reporting across native and bridged providers.

## PR Type
- 🆕 New Feature

## Relevant issues
Resolves #1418

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added provider metadata indicating whether Anthropic Messages API support is native.
  * Native Messages API support is now recognised for Anthropic, Meta, and Otari providers.
  * Providers with native support can use container, context-management, and beta parameters without fallback rejection.

* **Bug Fixes**
  * Improved handling of Messages API requests by distinguishing native provider support from converted requests.

* **Tests**
  * Added coverage for provider capability flags and Messages API metadata.


njbrake (approved):
Approving. The flag answers #1418 and is accurate: `SUPPORTS_MESSAGES_NATIVE` is true for the five providers that override `_amessages` (anthropic, azureanthropic, vertexaianthropic, meta, otari) and false everywhere else.

Two changes pushed to the branch as 547c8ab1:

- Dropped the `MESSAGES_NATIVE` alias. Two independent attributes for one capability, and `get_provider_metadata()` read only `SUPPORTS_MESSAGES_NATIVE`, so a provider setting just the alias reported `messages_native=False`.
- Reverted the conditional in the base `_amessages`. That body is the bridge, so a native provider overrides it and never reaches the check. Reading the flag there would silently drop `container`, `betas` and `context_management` for a future provider that set the flag without overriding the method.

Added a test pinning the flag to whether the provider overrides `_amessages`, so the two cannot drift.

Otari cleanup tracked in mozilla-ai/otari#1494.

